PR #216, opened 2016-04-04
signal from activity to decider via worker made by John R with implem…

Jenniferstrej:
…entation removed


gnott:
I'm testing the new return values. 

The ACTIVITY_PERMANENT_FAILURE activity return value is not doing what we had before in testing and as was discussed at the time.

We ended up not failing the activity right away for a permanent failure, making this line undesirable
https://github.com/Jenniferstrej/elife-bot/blob/feature-signal-workflow-termination/worker.py#L95

Otherwise SWF triggers a new decision right away. It leads to "FailWorkflowExecutionFailed" in the workflow event history and can potentially run an activity twice.

I suggest we remove L95 for now and go forward with the behaviour we (John and myself) discussed.

jhroot:
@gnott  huh, this is the code from my branch, maybe that last change something you (or I perhaps!) only had locally. But yes, looking at the flow it does make sense to remove L95.

PR #224, opened 2016-04-27
Cr 20 - Redis Session implementation and a couple of ConvertJATS activity unit tests

Jenniferstrej:
@gnott :
- I have changed the File Session get_value implementation so when the key doesn't exist, it returns None instead of raising an exception. So I adapted the code where it was expecting an exception (activity_PostEIF.py, activity_ExpandArticle.py). That way it stays consistent with the Redis session implementation. Redis returns None when the key requested doesn't exist.
- In activity_ScheduleCrossref.py I added the bucket prefix concatenation so then I can run the activity locally and it stays consistent with the other bucket references.
- (provider/execution_context.py) Added RedisSession as the default implementation for Session. It can be changed on the settings file via the attribute session_class.
- Added a couple of unit tests for ConvertJATS activity and it's fixtures. On readme I explain how to execute and the tests are run on Travis CI https://travis-ci.org/Jenniferstrej/elife-bot.


gnott:
Are there any new values for in "settings-example.py" regarding Redis? 

And do you know if Redis is already running on the production instance? Just wondering how long it will be before we can merge this into master later.

Jenniferstrej:
@gnott - Sorry, I didn't realise there was settings-example.py in this project. I have added session_class to it now. Also, "RedisSession" is applied as default if session_class isn't present in the settings.

Redis is not yet running in the builder, but according to Luke, we can update it by running a simple command "./bldr aws_update_stack", so it shouldn't be a problem.

jhroot:
@gnott is there anything else we should do prior to merging this do you think?

gnott:
Merging to develop should be ok. I wasn't sure if Redis was ready for deployment yet, though we can probably test that out before we merge develop to master.

Jenniferstrej:
I am going to merge this PR. I will then merge it into continuum-test branch and test it there. Then, before merging into master, I need to add some settings to the live() config on the builder, Like this https://github.com/elifesciences/elife-builder/commit/2fc7fe35137b485a42646cb3c2e42a1d32fadb41

PR #228, opened 2016-05-06
Fixes ApplyVersionNumber error monitor event

Jenniferstrej:
Replaces wrong activity reference with right one. And returns False when there is an exception. 

@gnott , @jhroot


gnott:
Good catch. I think the "False" could be "ACTIVITY_PERMANENT_FAILURE", now or in the near future.

jhroot:
Yes, that would be a good return value. I believe we're planning to update the failure codes for each activity at the same time as adding the additional unit tests

PR #229, opened 2016-05-06
ELPP-851

Jenniferstrej:
request will not verify certificate when we are testing since we can't get a certificate for the test sites.

http://jira.elifesciences.org:8080/browse/ELPP-851?filter=-1

Since there aren't enough certificates available for all sites, Luke requested that we have a flag to not require ssl certificate when in dev.

@gnott , @jhroot


gnott:
I'm looking forward to this. It can also solve http://jira.elifesciences.org:8080/browse/ELPP-836

PR #231, opened 2016-05-09
elementtree is in the standard library

giorgiosironi:
The package is called `xml.etree` and indeed Python 2.7 provides it by default. The code of elife-bot correctly refers to it as `xml.etree`.
@gnott

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -24,10 +24,6 @@ Python dependencies should be installed with `pip` using:
     
     $ pip install -r requirements.txt
 
-And for the library `elementtree` which is not part of Pypi
-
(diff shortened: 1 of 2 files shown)


gnott:
That fits with what I remember.

I tracked down where I removed this requirement from the PoA generation (https://github.com/elifesciences/elife-poa-xml-generation/commit/6728c3a953e2c2ffd1e5ae4b170d396a955a68c6) and I think that was the last place it may have been needed. 

The elife-bot requirements didn't get updated at the same time, and the instances and environments we are running may have been built prior to Oct 2015 when the requirement was no longer required.

PR #232, opened 2016-05-10
Committing empty root for temporary folders

giorgiosironi:
The tests create many temporary folders such as
`tests/tmp/016-05-10.07.52.15.filesystem_unzip`. However, their
parent `tests/tmp/` is not created by default, and the mkdir fails.

The behavior of Filesystem in case of creation error didn't
make sense:
- assignment of a variable to itself
- ignoring the error and nullifying the field in case
it already exists, although its name should be randomly
generated

Finally, .gitignore also precisely ignores the
content of that folder and Vim swap files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,9 +2,10 @@
 ftp-credentials.txt
 *.log
 *.pyc
+*.swp
 settings.py
 .boto
-tmp
+tests/tmp/*
 venv
 .idea
 workflow-context/*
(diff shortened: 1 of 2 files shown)


gnott:
Does it only affect the file system provider tests Giorgio? Plans are to remove it after some refactoring (http://jira.elifesciences.org:8080/browse/CR-23) so it and its tests may get removed soon, just a note.

giorgiosironi:
It's possible that only those tests require this, but I seem to remember that as long as the file system provider is used inside the application there may be other larger tests that depend on it.

gnott:
Runs ok too on a workflow so it has no side effects.

giorgiosironi:
Thanks for the in-depth check.

PR #239, opened 2016-06-01
Changes for Post EIF (shim)

Jenniferstrej:
CR-43, CR-44, CR-45, CR-46.
I also have made a pull request on the builder, to get the new settings for the new queue that the shim will read from. https://github.com/elifesciences/elife-builder/pull/158
@gnott - Could you take a look run the aws_update_stack command on the builder, or if you prefer I can do that once merging.
@jhroot - that looks like everything, right?


gnott:
Started with a quick read through. Will the final PR point to the develop branch or directly to master? (or is this just for viewing first?)

Does shimmy.py need to get started as a process by scripts/run_env.sh?

Jenniferstrej:
@gnott - Yes, the shim will start with the others. I've just added the commit with it. I will first merge it into develop and leave it on continuum today.

PR #244, opened 2016-06-03
Revert "Revert "Changes for Post EIF (shim)""

gnott:
Reverts elifesciences/elife-bot#241


gnott:
Attempting to merge develop into master has conflicts. When I try other methods it still ends up deleting the files that were deleted as part of the recent reverted commit. It seems a way to fix it is to revert the revert, then we should hopefully be able to proceed normally.

PR #246, opened 2016-06-06
Uniforming cli interface to accept -f parameter

giorgiosironi:
This is an intermediate step as -f will become always 1 in the future (and hence removed) while upstart manages the number of processes.

### DIFF
--- a/queue_workflow_starter.py
+++ b/queue_workflow_starter.py
@@ -34,6 +34,7 @@ def main():
     parser = OptionParser()
     parser.add_option("-e", "--env", default="dev", action="store", type="string", dest="env",
                       help="set the environment to run, either dev or live")
+    parser.add_option("-f", "--forks", default=5, action="store", type="int", dest="forks", help="specify the number of forks to start")
     (options, args) = parser.parse_args()
     if options.env:
         env = options.env
@@ -48,7 +49,7 @@ def main():
     # Simple connect
     queue = get_queue()
 
-    pool = Pool(settings.workflow_starter_queue_pool_size, initialise_pool, [env])
+    pool = Pool(options['forks'], initialise_pool, [env])
 
     while True:
         messages = queue.get_messages(num_messages=settings.workflow_starter_queue_message_count, visibility_timeout=60,
--- a/scripts/run_env.sh
+++ b/scripts/run_env.sh
@@ -11,5 +11,5 @@ alias ..='cd ..'
 source /opt/elife-bot/venv/bin/activate && python decider.py -f 3 -e $1 &
 source /opt/elife-bot/venv/bin/activate && python worker.py -f 5 -e $1 &
 source /opt/elife-bot/venv/bin/activate && python queue_worker.py -f 3 -e $1 &
-source /opt/elife-bot/venv/bin/activate && python queue_workflow_starter.py -e $1 &
+source /opt/elife-bot/venv/bin/activate && python queue_workflow_starter.py -f 5 -e $1 &
 source /opt/elife-bot/venv/bin/activate && python shimmy.py -e $1 &
--- a/shimmy.py
+++ b/shimmy.py
@@ -117,6 +117,7 @@ if __name__ == "__main__":
     parser = OptionParser()
     parser.add_option("-e", "--env", default="dev", action="store", type="string", dest="env",
                       help="set the environment to run, either dev or live")
+    parser.add_option("-f", "--forks", default=1, action="store", type="int", dest="forks", help="specify the number of forks to start")
 
     (options, args) = parser.parse_args()
     if options.env:


gnott:
@jhroot maybe wants to look to confirm?

giorgiosironi:
Talking with @jhroot we want to keep one process that does the polling to avoid creating contention by the children. So I will go back to that solution and put just 1 process in upstart for each of the process types.

Moreover, we want configuration of the number of processes to be in a single place, and to be configurable per-environment. Therefore that should be moved to `settings.py`.

giorgiosironi:
What I discovered so far:
- each of the forks of `decider`, `worker`, `queue_worker` does its own polling
- `shimmy` also does its own polling, but it does not support
- `queue_workflow_starter` (for reasons I don't know, maybe it's newer code with respect to the others) does the polling in the main loop and maps received messages over its children)
- given that `queue_workflow_starter` consolidates the polling of about 130,000 requests per month ( given a long polling interval of 20 seconds) which otherwise would be multiplied by 5. This feature saves us about $0.50 per environment per month?
- `queue_workflow_starter` is intelligent enough to respawn dead children
- if `queue_workflow_starter` is killed, the children do not with it but remain there orphaned, with the old code, for an unspecified amount of time.

The options to let the python code deal with the forking:
- `daemon` option is not really viable: as it prevents children from creating other processes, a `os.system` call in a library will make the bot explode 
- handling signals and accepting SIGTERM in the parent, propagate to the children before exiting. SIGKILL seems automatically propagated. Here we open the Pandora's box of signals...
- do some blunt commands with automated `killall -u elife python` at each deploy
To let upstart deal with the forking:
- we let each process do its own polling, possibly spend some dollars more, all processes are children of upstart and are managed and respawned with it; no wheel reinvention.

@lsh-0 what do you think?

lsh-0:
From what I can tell we have a case of premature optimization that has led to a script spawning/managing children that aren't killed when it's parent is killed. 

Yes, it's possible to brutally kill anything that looks like a child process of the parent process when we need to update code and I'm confident that with periodic `kill python` cronjob the bot has probably grown skin thick enough to deal with processes being killed mid-task. But encouraging what feels like anti-pattern in process management and making the bot take on responsibilities it need not for a very (very) minor efficiency gain should be something that is nipped in the bud. Keep code, especially multiprocess code, dumb. Keep it easy to reason about.

I would suggest we refactor `queue_workflow_starter` to behave similarly to the others and leave process management to the init script (upstart/systemd/supervisord/whatever).

Giorgio informs me the refactor should be straightforward to do.

giorgiosironi:
To be precise I can take on the task to change `queue_workflow_starter`, if we have you all on board. But since I may be missing additional reasons for the consolidation of polling, this is the place to discuss.

lsh-0:
Indeed.

On Thu, Jun 9, 2016 at 3:28 PM, Giorgio Sironi <notifications@github.com>
wrote:

> To be precise I can take on the task to change queue_workflow_starter, if
> we have you all on board. But since I may be missing additional reasons for
> the consolidation of polling, this is the place to discuss.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/elifesciences/elife-bot/pull/246#issuecomment-224911973>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ACCyHHO2YgG7TPqqcpYIKe2AwqM-9iFcks5qKCMbgaJpZM4IuxzV>
> .
>



-- 

Luke Skibinski

Formally defined,
Extensively tested,
Cryptic, abstract, thing.

--

eLife Sciences Publications, Ltd is a limited liability non-profit
non-stock corporation incorporated in the State of Delaware, USA, with
company number 5030732, and is registered in the UK with company number
FC030576 and branch number BR015634 at the address First Floor, 24 Hills
Road, Cambridge CB2 1JP.

gnott:
Probably already implied in the above though I'll mention it anyway: decider and worker poll SWF, and there's no additional costs associated with polling. We could have thousands of each on multiple instances. (on multiple instances provided the Redis server is federated across all instances that use it). Those can also be run as single processes with no multiprocessing / forks, they don't care.

cron.py (not mentioned here) is currently only happen when it is only running one process on one instance. I know we're not running multiple instances, but if we did, we'd only want one of those to run the cron.py

The processes that poll an SQS queue may also be happier when there's only one script polling each SQS queue.

giorgiosironi:
On decider and worker: great point, didn't know that.

On cron.py: definitely need to run on a "leader" instance, if we have multiple servers. It's not a problem for process management because (from the name) it seemed to me a process that is launched periodically, but is not a long-running process to monitor and respawn continuously.

On the SQS polling: from the documentation it always seemed that multiple writers and readers are supported by SQS. They specify how multiple readers work with a visibility timeout in the [FAQ](https://aws.amazon.com/sqs/faqs/) at `How does Amazon SQS allow multiple readers to access the same message queue, without losing messages or processing them many times?`, for example. 
So Competing Consumers which is our use case does not seem to have an impact (in fact if we had multiple servers that would be no way around it) apart from the possible costs of making millions of calls. In what sense they are happier? I realize more polling can create more contention in general, but I would worry about it if we had a Postgres table with the queue messages rather than a specialized queue service.

jhroot:
"From what I can tell we have a case of premature optimization that has led to a script spawning/managing children that aren't killed when it's parent is killed."

I just want to chip in that I don't consider it premature optimisation, its about much more efficient use of the AWS API and reducing the cost involved. @lsh-0  can you please explain why this is "very (very)" minor?  and why it is an anti-pattern. 

The way the scripts in the bot were being managed meant that there wasn't a need to separately clean up these processes. Now that is being changed, there is. There are several strategies for this, which I sent links for to Giorgio. I'm not clear why any of these are not acceptable?

TLDR: If you change queue workflow starter to use the same pattern as worker, decider, etc. You'll increase the SQS cost but n times - where n is the number of processes.

jhroot:
The queue_workflow_starter uses a single process to poll for incoming messages and receives up to a configurable number (settings.workflow_starter_queue_message_count) of messages. It is configured to use the maximum wait time (20 seconds).  see http://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-long-polling.html

rather than spawn N processes, each of which poll the same SQS queue for messages to process, the queue_workflow_starter (and some other parts of continuum) have a main process that polls the queue, receives any messages that need processing and then uses the multiprocessing library to parcel this work out to a Pool of worker processes. upon completion of that set of messages the process starts polling the queue again for more. This doesn't feel like an anti-pattern to me but I'm happy to listen to reasons why it is.

giorgiosironi:
Perfectly clear on how `queue_workflow_starter` works.

About cost, I've taken a look at the May 2016 bill:
```
$0.50 per 1,000,000 Amazon SQS Requests per month thereafter 	1,367,460 Requests 	$0.68
First 1,000,000 Amazon SQS Requests per month are free 	1,000,000 Requests 	$0.00
```

About the pattern, having consumers that directly work with the queue is also pretty common (http://www.enterpriseintegrationpatterns.com/patterns/messaging/CompetingConsumers.html) Of course given the fact that SQS is implemented with polling over HTTP which is inefficient, it multiplies the number of requests.
But: doesn't processing multiple messages in batches defeat the purpose of having multiple processes for them? `map` is synchronous, so e.g. if the parent process receives 5 messages that would take [10, 0.1, 0.1, 0.1, 0.1] seconds to process, it has to wait 10 seconds before processing any new one. I don't know if starting a workflow is an expensive operation, but if we have multiple processes it's because we want to have an higher throughput, not to block all processing waiting for the slowest of them to finish?

jhroot:
Yes, I've implemented it both ways before but here I've been following the AWS recommendations regarding long polling and consumers. If you want truly competing consumers then you should probably only obtain one message at a time and that will also include the numbers. If thats fine then we can modify this to spawn multiple processes it should work fine. (Personally I'd just fix the orphaned child issue, but either way will work).

PR #247, opened 2016-06-07
reduce logging in shimmy

jhroot:


giorgiosironi:
:+1: on precise logging

Still I don't understand the name, not being a native speaker? https://en.wiktionary.org/wiki/shimmy

jhroot:
its from 'shim' :) https://en.wikipedia.org/wiki/Shim_(computing)

PR #251, opened 2016-06-13
Squashed commit of the following:

jhroot:
commit 5d21273e34e201fc1a2f8545d4c99b2f15d60818
Author: John Root <john.root@digirati.co.uk>
Date:   Fri Jun 3 12:30:12 2016 +0100

    use new return codes for activity_ExpandArticle

### DIFF
--- a/activity/activity_ExpandArticle.py
+++ b/activity/activity_ExpandArticle.py
@@ -70,13 +70,13 @@ class activity_ExpandArticle(activity.activity):
         if version == '-1':
             self.logger.error("Name '%s' did not match expected pattern for version" %
                               filename_last_element)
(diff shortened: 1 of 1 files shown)


gnott:
Would it work as self.ACTIVITY_PERMANENT_FAILURE too, instead of activity.activity.ACTIVITY_PERMANENT_FAILURE?  Just curious

Jenniferstrej:
@jhroot and @gnott - I am going to close this PR since I already did the same work as I was writing tests for ExpandArticle.
Graham, yes, it looks like self would work the same way. Should we change?

gnott:
On using ``self`` or not I wasn't decided one way or the other, since they're both valid, although I personally would probably go with ``self`` for shortness and clarity.

PR #252, opened 2016-06-17
Basic tests for the PackagePOA activity.

gnott:


Jenniferstrej:
@gnott - could you please change your file name from test_activity_package_poa.py to activity_package_poa_tests.py? having *_tests.py in the end will make it easier for the discover command to find the tests among other files. They were like this before and I changed them. I wanted the file names to be lower case but didn't realise that removing "_tests.py" from the end would make the file name difficult to differ from the other helper files.

Jenniferstrej:
I deleted my previous message as I see the changes I was thinking of were made by Giorgio for a reason. I will take a look.

giorgiosironi:
Which changes?

gnott:
@giorgiosironi I think it was about the naming of test file names but everything is named as it should be.

giorgiosironi:
:+1: on that

Jenniferstrej:
Sorry @giorgiosironi, I meant the changes in the test file names. But it's all back to normal now. :)

Jenniferstrej:
I am merging this PR now and will merge my tests too. We can take today and tomorrow for exploratory tests on the continuum-test instance. If anyone does not agree please let me know and I will revert those changes.

PR #253, opened 2016-06-17
ExpandArticle and ResizeImages activity unit tests

Jenniferstrej:


Jenniferstrej:
@gnott - there is also some refactoring on the ExpandArticle activity, if you'd like to take a look.

gnott:
I like it. Lots of work here!

For testing, the fake StorageContext is hardcoded to read from "data.ExpandArticle_files_source_folder", so as long as we can improve this later so I can use it too in other tests.

I'll want to continue S3StorageContext for it to list keys from a bucket later. We can do that incrementally as we go.

Minor detail, you could remove the file "tests/files_source/.DS_Store"

PR #257, opened 2016-06-23
Minimal process management

giorgiosironi:
Summarizing what we discovered/discussed so far.

- The test pipeline requires deployment to be fully automated, so that at each new push we can run test without someone logging in into a machine and restarting processes.
- Upstart is the standard daemon at this time that Luke chose to automate the long-running processes. The dashboard already uses it.
- Upstart does not play well with processes that create multiple children: to fully manage their start/stop/restart it wants a single layer of processes under it.
- We want to have configuration of the number of processes in one place.
- There aren't costs associated with multiple SWF workers, and there are some in the order of units of dollars/month for multiple SQS workers.

Therefore:
- this minimal change make it so that any of the 5 processes (worker, queue_worker, queue_workflow_starter, decider, shimmy), when called without an `-f` argument, does not create children and stays single-threaded.
- since `queue_workflow_starter` did not have such an argument, I tried to make a temporary change for making it single-process.
- https://github.com/elifesciences/elife-builder/pull/175 can be merged right after this.
- The Salt configuration will provide the number of processes, which will then vanish from the configuration when specified, or become actually configurable if it was hardcoded in the script.
- This is intended as a *minimal* change; if it gets into `develop` I will then go on in cleaning up the related code, greatly simplifying all 5 processes, removing the `-f` argument, modifying/deleting `scripts/run_env.sh` to avoid redundancy.

Take a look around, to bring stability to our deployments I intend to go ahead with the start of next week if there are no objections, in the spirit of Ian's last email on unblocking decisions.

### DIFF
--- a/decider.py
+++ b/decider.py
@@ -243,7 +243,7 @@ if __name__ == "__main__":
     parser = OptionParser()
     parser.add_option("-e", "--env", default="dev", action="store", type="string",
                       dest="env", help="set the environment to run, either dev or live")
-    parser.add_option("-f", "--forks", default=10, action="store", type="int",
+    parser.add_option("-f", "--forks", default=1, action="store", type="int",
                       dest="forks", help="specify the number of forks to start")
     (options, args) = parser.parse_args()
     if options.env:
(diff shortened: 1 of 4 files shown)


gnott:
The plan is reasonable to me, thanks Giorgio

Jenniferstrej:
@jhroot - any input?

giorgiosironi:
Talking with John he does not block this. Let's see how it works.

PR #259, opened 2016-06-27
Processes streamlining

giorgiosironi:
We can now remove a lot of code related to multiprocessing, since that is managed by the Upstart configuration.

### DIFF
--- a/decider.py
+++ b/decider.py
@@ -5,8 +5,8 @@ import json
 import random
 import importlib
 import time
-from multiprocessing import Process
 from optparse import OptionParser
+from provider import process
 
 import workflow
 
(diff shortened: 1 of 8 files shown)


giorgiosironi:
Rebased to resolve conflicts.

Jenniferstrej:
👍

giorgiosironi:
Long story short: due to timeouts of precisely 5 minutes in the end2end tests, I discovered that we need to shutdown the workers in a cleaner way if we want to avoid activities being assigned to dead workers:
http://docs.aws.amazon.com/java-sdk/latest/developer-guide/swf-graceful-shutdown.html
If there are no blockers to this PR, tomorrow I'll merge and test it in end2end, to then build upon `monitor_interrupt` and intercept the SIGTERM.

giorgiosironi:
As discussed on Slack/email, going for this and building graceful shutdown on top to avoid 15-minute timeouts of activities after deploys.

PR #261, opened 2016-06-28
Retry on 429 response code (throttling)

giorgiosironi:
The end2end tests evidenced that this can happen, despite the shimmy
being a single process. In this case, we stop processing the message and
let it go back to the queue for a retry.

Testing the logic with unit tests, more difficult to actually test the
integration with SQS. There may be a better way of triggering a retry,
and this logic targets 10 seconds from the 429 error.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,2 @@
+[pytest]
+testpaths=tests
--- a/shimmy.py
+++ b/shimmy.py
@@ -15,104 +15,104 @@ logging.basicConfig(filename='shimmy.log', level=logging.INFO)
 class ShortRetryException(RuntimeError):
     pass
 
-def listen():
+class Shimmy:
+    def __init__(self, settings):
+        self._settings = settings
 
-    global settings
-    settings = settings_lib.get_settings(ENV)
+    def listen(self):
+        conn = boto.sqs.connect_to_region(self._settings.sqs_region,
+                                          aws_access_key_id=self._settings.aws_access_key_id,
+                                          aws_secret_access_key=self._settings.aws_secret_access_key)
+        input_queue = conn.get_queue(self._settings.website_ingest_queue)
+        output_queue = conn.get_queue(self._settings.workflow_starter_queue)
+        if input_queue is not None:
+            while True:
 
-    conn = boto.sqs.connect_to_region(settings.sqs_region,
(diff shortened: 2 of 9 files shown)


gnott:
On the use of settings.py, is it possible to use the mock settings file in some way, like https://github.com/elifesciences/elife-bot/blob/develop/tests/activity/settings_mock.py

It would be good to know the automated tests do not load a real settings.py file, even if it does not use the settings inside it.

giorgiosironi:
It's actually the `shimmy` loading `settings`:
https://github.com/elifesciences/elife-bot/blob/develop/shimmy.py#L9
I guess I will refactor that code to avoid depending on the global `settings`, transforming it into a class.

gnott:
I see, it is different and probably more challenging than the activity tests. It'd be great if you can ignore the main settings.py though

giorgiosironi:
Refactored to a class. Same situation on testing: unit passes, manual startup works, wondering about working in the real world. The `settings.py` file is not read anymore by tests.

Also, removed a warning on a temporary directory not cleaned up, and the deletion of `.keepme` which is not performed anymore.

Jenniferstrej:
Right, so I don't need to do this bit anymore (the .keepme bit). Thanks
Giorgio.

On 29 June 2016 at 09:30, Giorgio Sironi <notifications@github.com> wrote:

> Refactored to a class. Same situation on testing: unit passes, manual
> startup works, wondering about working in the real world. The settings.py
> file is not read anymore by tests.
>
> Also, removed a warning on a temporary directory not cleaned up, and the
> deletion of .keepme which is not performed anymore.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/elifesciences/elife-bot/pull/261#issuecomment-229291126>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AC4pPiWEquESVKtJOK0jGqbhgZnMMJMFks5qQi0PgaJpZM4JAEsa>
> .
>



-- 
Jennifer

gnott:
Looks like it solves my settings concern, it looks good there @giorgiosironi

giorgiosironi:
Finishing this run of the end2end tests, then merging and keeping an eye on them.

PR #265, opened 2016-06-30
Test, file open use 'rb' mode instead of 'r'.

gnott:


giorgiosironi:
From what I know this can only improve performance :+1:

Jenniferstrej:
From https://docs.python.org/2/tutorial/inputoutput.html#reading-and-writing-files

"On Windows, 'b' appended to the mode opens the file in binary mode, so there are also modes like 'rb', 'wb', and 'r+b'. Python on Windows makes a distinction between text and binary files; the end-of-line characters in text files are automatically altered slightly when data is read or written. This behind-the-scenes modification to file data is fine for ASCII text files, but it’ll corrupt binary data like that in JPEG or EXE files. Be very careful to use binary mode when reading and writing such files. On Unix, it doesn’t hurt to append a 'b' to the mode, so you can use it platform-independently for all binary files."

gnott:
The advantage is it will run for me (on Windows) otherwise it fails without this fix.

Jenniferstrej:
👍

PR #266, opened 2016-07-01
Elpp 754 - Extracted lax call for article_version and publication date and applied where it's used

Jenniferstrej:
@gnott - I have extracted the lax_article_versions references into a new provider. I realise that there may be more areas where other lax references could be extracted there. I will continue checking, but if you could give me your thumbs up on the progress it would be helpful, so then I stop and make any changes early on. 

Another thing is that when I run test_activity_package_poa locally, the test_do_activity test doesn't pass, but on travis it does. I get "ValueError: need more than 2 values to unpack". Can you run this locally alright?


gnott:
Quick answer for running tests, please try to pull the master branch of elife-poa-xml-generation. The latest version will be required to operate the PackagePOA test.

More comments soon!

gnott:
@Jenniferstrej that's what I was imagining in a provider, the basics look good.

Some things I noticed on reading for confirmation,

- A major tweak: In activity_ExpandArticle.py the original function actually returns version + 1 instead of the highest version only
- Minor tweak: In activity_ExpandArticle.py you could also pass the self.logger as a parameter
- A new or more detailed specification: The article_publication_date() actually needs to only check version 1 of the article for the correct date. I was making an incorrect assumption for what data was in lax. If you are able to get it to only read the "datetime_published" from version 1, that'd be great (if not, I will patch it later).

I assume you're planning to add some tests with some mock data?

Jenniferstrej:
Hi @gnott - Right, I made the corrections and wrote some tests. It would be worth it double checking the acceptance criteria of the tests too. You will notice that I duplicated settings_mock. There is a new one in the root of tests folder and the old one in the tests/activity folder. The only reason I haven't moved all the imports there is that I haven't managed to make my PyCharm work properly with importing "import tests.settings_mock as settings_mock". When I debug or run "py.test tests/provider/test_article.py" for example, I get an import error. However when doing "python -m pytest" (which is what travis runs) it's all fine. So it's doing something clever there that I haven't worked out. So, that's why it's still there. I'll keep trying to make it debug and run via the py.test command but for now I will leave the settings_mock.py file inside activity folder as well.

giorgiosironi:
@Jenniferstrej The `py.test` command does not add the current directory to the `PYTHONPATH` environment variable, so you get import errors as the local package cannot be imported but it can only find the ones installed in the system (`/var/local/python2.7/site-packages` and similar).
When you run `python -m ...`, the Python interpreter adds `.` to the PYTHONPATH for you, and the test suite works.

Jenniferstrej:
@giorgiosironi - I added the bot root to the PYTHONPATH before running py.test. It did successfully the provider.article export https://github.com/Jenniferstrej/elife-bot/blob/730e035e569511d669cc9b9ad1967dc1910cb60e/tests/provider/test_article.py#L2 but still couldn't find settings_mock https://github.com/Jenniferstrej/elife-bot/blob/730e035e569511d669cc9b9ad1967dc1910cb60e/tests/provider/test_article.py#L3

gnott:
I tried some experiments with other ways to run tests (I use a different one to compile coverage for my own information at times).

Just an idea that maybe the tests/ and tests/activity folders need a __init__.py file? That might allow them to be imported correctly. No guarantees whether that will help though, it might even make it worse, I'm not sure.

giorgiosironi:
That may help; pytest suggests not to do it to keep tests running against an installed version of the package: http://pytest.org/latest/goodpractices.html ..but we are not interested in that at all. So I second the attempt to use `__init__.py` files there.

Jenniferstrej:
I have already tried the adding the __init__py files. They are part of the changes in this PR. Since the tests are working with pytest I suggest we merge it, then when you have it in your machine you could try to run it in your environments and if it works for you we could compare our settings. What do you think?

Any objections on the changes themselves?

giorgiosironi:
:+1: from me

gnott:
I like it for develop branch testing.

PR #270, opened 2016-07-05
Tweaks for article_structure lettuce tests to pass.

gnott:
The main edit is in the regexp in article_structure.py: it makes the -v1 / -r1 file name part optional again. This allows the existing test to pass (even though we do not current read zip files with no r or v value, it was an existing test condition).

The other tweaks were to improve the tests with expected data and checking all attributes.

PR #271, opened 2016-07-06
Latest develop changes into master

Jenniferstrej:
I am opening this one just so we can discuss if there is anything stopping us from deploying the bot as there are quite a few changes awaiting and increasing this queue can become 'dangerous'.


giorgiosironi:
What if I create the `prod-elife-bot` pipeline and we deploy from there? It is comparatively less risky to deploy changes that pass the existing tests.

Jenniferstrej:
Yes. I am totally for it. I just find creating a PR more encouraging to make us look at the changes and take decision to deploy :). 👍

giorgiosironi:
Created http://ci.alfred.elifesciences.org/job/prod-elife-bot/
but the code in the `approved` branch is already old because of the broken tests. So I'm merging https://github.com/elifesciences/elife-bot/pull/270 to see if it solves.

gnott:
I see some smaller improvements and backwards-compatible changes. I think if we tested a PPP article and a PoA generate on a non-production environment then we can say this is deploy ready.

giorgiosironi:
I am trying to configure `master` to be a protected branch so that only automated processes like the build can merge into it, you may see the `Merge pull request` button here disappear or something like that.

giorgiosironi:
@gnott I want to understand how difficult is to automate these two checks.
Does the PPP article consist of uploading one article to the S3 `*-elife-production-final` bucket?
What does the PoA one require to do instead?

Jenniferstrej:
"Does the PPP article consist of uploading one article to the S3 *-elife-production-final bucket?": Yes AND also when you feed through econ-tools

gnott:
For PoA I was meaning PackagePOA workflow specifically. It requires a zip file and a set of CSV files, and the result should two or three files as an output.

Subsequently, if that all works, you can run a second workflow PublishPOA and those files should be packaged into a PPP style single zip file, copied into the *-elife-production-final bucket, and then it will kick off another PPP workflow.

This is not too hard to trigger manually with the right input data, although automating it would be very cool if it is possible.

giorgiosironi:
If you can send me a sample zip file plus its CSV, I can work on them to make a single end2end that starts from PackagePOA. But how are the workflows (PackagePOA and PublishPOA) triggered by users? Is there some user interface or do I need to put them into some bucket?

gnott:
I think this PR is good to merge, anyone disagree?

Jenniferstrej:
👍

giorgiosironi:
Time to run http://ci.alfred.elifesciences.org/job/prod-elife-bot/ then!
It won't merge anything into `master` because it is already merged, but it should deploy the changes to `elife-bot-master`.
If you have already deployed, I will run it afterwards to check that it deploys no changes (the same commit) but all the process of setup and restart of processes goes smooth.

giorgiosironi:
I've read in the mailing list that it was already deployed.
I hence tested automatic deploy, which also works as can be seen at run 2 at http://ci.alfred.elifesciences.org/job/prod-elife-bot/ (run 1 had an outdated `approved` branch, now aligned).
However there is no way to stop a pull request from being merge-able here from repository administrators (`master` is a protected branch), and in that case `master` gets 1 commit more than `approved` stopping the pipeline from working (.

PR #273, opened 2016-07-13
Remove duplicate FakeKey class.

gnott:
Something I noticed when working on the "update" JSON feature.


Jenniferstrej:
If the tests still run I think it's fine. I probably forgot it there, sorry. 👍

PR #283, opened 2016-07-22
Processes graceful shutdown

giorgiosironi:
`SIGTERM` is intercepted to make the `Flag` object red. All polling cycles exit at the next iteration, and the processes terminate.

### DIFF
--- a/provider/process.py
+++ b/provider/process.py
@@ -1,14 +1,38 @@
+import signal
+
 """
 Provides process-management utilities such as catching signals and interrupts
 """
+
+class Flag:
+    def __init__(self):
+        self._red = False
+
+    def green(self):
+        return not self._red
+
+    def red(self):
+        return self._red
+
+    def stop_process(self):
+        self._red = True
+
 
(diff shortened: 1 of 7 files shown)


provider/process.py
```diff
@@ -1,14 +1,38 @@
...
     if they happen.
+
+    Exits immediately for Ctrl+C interrupts. Waits for graceful shutdown
+    for SIGTERM instead.
     """
     try:
-        work()
+        flag = Flag()
+        def signal_handler(signum, _frame):
+            print "received signal %s" % signum
```
giorgiosironi: Here we should use the `logger` from the process, but it's not possible to pass it in with the current architecture; will do that in another PR once the tests are stable.

giorgiosironi:
http://ci.alfred.elifesciences.org/job/test-elife-bot/89/ continues the series of failures related to the dead processes, so if nothing comes up I will merge this today to try and improve the stability.

giorgiosironi:
End2end build somewhat green, let's go for this.

PR #284, opened 2016-07-22
Feature image resize to disk

giorgiosironi:
May improve the reliability of a downloaded image, since in end2end tests we sometimes attempt to resize a corrupted one.

### DIFF
--- a/activity/activity_ResizeImages.py
+++ b/activity/activity_ResizeImages.py
@@ -5,6 +5,7 @@ from mimetypes import guess_type
 import activity
 import boto.swf
 import log
+import os
 import provider.imageresize as resizer
 import yaml
 from boto.s3.connection import S3Connection
@@ -108,8 +109,12 @@ class activity_ResizeImages(activity.activity):
             self.generate_images(formats, fp, info, cdn_path)
 
     def get_file_pointer(self, key):
-        fp = StringIO.StringIO()
+        file_name = key.name.split('/')[-1]
+        file_path = self.get_tmp_dir() + os.sep + file_name
+        fp = open(file_path, mode='wb')
         key.get_file(fp)
+        fp.close()
+        fp = open(file_path, mode='r')
         return fp
 
 
--- a/activity/activity_ResizeImages.py
+++ b/activity/activity_ResizeImages.py
@@ -109,7 +109,7 @@ class activity_ResizeImages(activity.activity):
             self.generate_images(formats, fp, info, cdn_path)
 
     def get_file_pointer(self, key):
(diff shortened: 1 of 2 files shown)


Jenniferstrej:
When resizing via command line, although there is an error, a jpeg is created in the file system and 'looks' ok. However @jhroot and I thought it may be unreliable to have the image saved to cdn even though an exception is thrown on the image resizing (Wand) plugin. Could you clarify to me how you think it will improve reliability of the downloaded image? Thanks.

giorgiosironi:
I don't think we are talking about the same error: this one is non-deterministic so you should not be able to reproduce an error with the command line; rerunning the same article often fixes this problem (but we don't want to have to work twice).
To clarify, this is the error under scrutiny here:
```
2016-07-20T13:07:41Z ERROR worker_234 error resizing image elife-1919895345-fig1-figsupp3-v2
  File "/opt/elife-bot/provider/imageresize.py", line 11, in resize
    with Image(file=filep, resolution=96) as tiff:
  File "/opt/elife-bot/venv/local/lib/python2.7/site-packages/wand/image.py", line 2054, in __init__
    self.raise_exception()
  File "/opt/elife-bot/venv/local/lib/python2.7/site-packages/wand/resource.py", line 222, in raise_exception
    raise e
CoderError: Read error on strip 84; got 24231 bytes, expected 42418. `TIFFFillStrip' @ error/tiff.c/TIFFErrors/508
```
but it vanishes when performing the test again.

giorgiosironi:
Ah, this can't run in Travis because it does not have the `settings.py` files... Looking into it.

Jenniferstrej:
Funny because the error happens in the same line. We get mostly the same ones
convert: bad.tiff: Null count for "Tag 34022" (type 1, writecount -3, passcount 1). `_TIFFVSetField' @ error/tiff.c/TIFFErrors/553
or
: Null count for "Tag 34031" (type 1, writecount -3, passcount 1). `_TIFFVSetField' @ error/tiff.c/TIFFErrors/557

Jenniferstrej:
Regarding the settings file, don't you have an end2end class on the bot settings file that rests in the builder? (things may have changed and I may be getting confused here)

giorgiosironi:
The errors: the same Python line because they come from the resize call, but the error messages from the C code of the library as you can see are different.
The settings: yes, the `settings.py` file is deployed on every builder machine, but not on Travis; so I'll have it skip this test.

Jenniferstrej:
Ok. The changes seem fine to me 👍

giorgiosironi:
Let's go, it's Friday, what could possible go wrong :)

PR #285, opened 2016-07-22
Uniforming shimmy logs to all the others processes

giorgiosironi:
While I was debugging 500 errors from the website, I found that this log
lacked the time label over each line, which the other processes have.
So I set out to make it use the same `log` module as them.

### DIFF
--- a/log.py
+++ b/log.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import random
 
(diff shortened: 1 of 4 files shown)


giorgiosironi:
Rebased to solve conflicts, waiting for the build to finish and then merging if ok.

PR #286, opened 2016-07-22
Remove settings for the old lens Converter, not used in years.

gnott:


gnott:
Thanks Giorgio. I will start updating the new repos when I start using them

PR #287, opened 2016-07-25
Feature PostEIF cleanup

gnott:
I'm looking at where update / update_date is specified, and these look old or obsolete now.

PostEIF activity - we've replaced this
ProcessXMLArticle workflow - is the only one that uses PostEIF, and I think it is very old

@jhroot might you confirm about ProcessXMLArticle workflow?
@Jenniferstrej confirm PostEIF we do not need anymore?


giorgiosironi:
As a passing comment, what do you think of running the pull requests into Alfred (only project tests as now, not end2end tests)? That way they will run in the same environment where the `develop` branch will, and we can then divest the project from Travis CI.

The feedback is very similar: `V` and `X` over the various commits linked to the passing/failing build, see https://github.com/elifesciences/journal/pull/8 for an example.

Jenniferstrej:
"@Jenniferstrej confirm PostEIF we do not need anymore?": Yes.

Jenniferstrej:
@giorgiosironi : In what level do you mean? Would the "Merge pull request" button be located in Alfred? Would also the conversation? Would we not use github interface anymore?
In either case, I don't see any harm trying. 👍

giorgiosironi:
@Jenniferstrej The "Merge pull request" button will be here, the conversation too, continuing to use the Github interface.
You'll see in the `All checks have passed` section the face of Michael Caine, and a link to the Alfred build, instead of the current `continuous-integration/travis-ci/pr — The Travis CI build passed` text.

Jenniferstrej:
Fine. Let's try 👍

PR #289, opened 2016-07-27
Fixed problems in existing tests in PostEIFBridge and increased test …

Jenniferstrej:
…coverage

### DIFF
--- a/tests/activity/test_activity_post_eif_bridge.py
+++ b/tests/activity/test_activity_post_eif_bridge.py
@@ -9,7 +9,6 @@ from classes_mock import FakeSQSQueue
 from classes_mock import FakeMonitorProperty
 from classes_mock import FakeLogger
 from classes_mock import FakeEmitMonitorEvent
-import classes_mock
 from testfixtures import TempDirectory
 import json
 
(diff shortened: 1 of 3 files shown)


tests/activity/test_activity_post_eif_bridge.py
```diff
@@ -19,31 +21,98 @@ def setUp(self):
...
         self.assertEqual(True, success)
+
         self.assertEqual(json.dumps(data.PostEIFBridge_message), data_written_in_test_queue)
+        self.assertEqual(fake_monitor_property.item_identifier, '00353')
+        self.assertEqual(fake_monitor_property.name, 'publication-status')
+        self.assertEqual(fake_monitor_property.value, 'published')
+        self.assertEqual(fake_monitor_property.property_type, 'text')
+        self.assertEqual(fake_monitor_property.version, '1')
+
+        self.assertEqual(fake_monitor_event.item_identifier, '00353')
```
giorgiosironi: This seems to specify the properties of the emitted properties and events. Is it possible to do a single `self.assertEqual` comparing the properties of `fake_monitor_event` with a single dictionary `{'item_identifier':'00353', ...}` or does it have some obscure properties (like timestamps) that we don't want to pin down?

Jenniferstrej:
@giorgiosironi - is that what you meant? It does look better and avoids repetition. The only property that I am not asserting that is sent to event and property monitors is "settings" because that is the whole settings file...

giorgiosironi:
Yeah, looks good. Of course checking all the settings would not add value. :+1:

PR #290, opened 2016-07-28
Feature tests tweaks

gnott:
Minor changes otherwise the tests fail in Windows.

### DIFF
--- a/tests/activity/test_activity_apply_version_number.py
+++ b/tests/activity/test_activity_apply_version_number.py
@@ -74,12 +74,10 @@ class MyTestCase(unittest.TestCase):
         self.applyversionnumber.rewrite_xml_file(u'elife-15224-v1.xml', example_file_name_map)
 
         #then
-        result_file = open(u'tests/files_dest_ApplyVersionNumber/elife-15224-v1.xml', 'r')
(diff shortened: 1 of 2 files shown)


tests/activity/test_activity_apply_version_number.py
```diff
@@ -76,8 +76,10 @@ def test_rewrite_xml_file(self, mock_path_join):
         #then
         result_file = open(u'tests/files_dest_ApplyVersionNumber/elife-15224-v1.xml', 'r')
```
giorgiosironi: what about the idiom `with open(u'tests/files_dest_ApplyVersionNumber/elife-15224-v1.xml', 'r') as result_file:`?

Jenniferstrej:
@gnott , I forgot about your situation with Windows and file open mode, sorry. 
Got point I also forgot to close the file, my bad. Thanks for spotting 👍

gnott:
The "with open(..." works ok too without the separate close() line, to note. Either or - I can change it if there's a preference.

giorgiosironi:
Yes, `with open()` makes it impossible to forgot the `close()` (and in case of exceptions coming up catches them, closes the file, and reraises)

PR #292, opened 2016-07-30
Feature update date redux

gnott:
For a first preview, this is a second try at the EIF update date feature (after the scope was altered).

Basically,

* Use the update_date value from the zip filename, if present. Do not set it prior to posting the EIF file otherwise
* The "update" value will be returned from the website endpoint when the article is published. This happens either in a) shimmy.py or b) ApprovePublication activity
* The "update" value is parsed, converted, and saved into the session
* Later in PostPerfectPublication workflow, the new RewriteEIF activity adds the "update" value to the EIF json file on S3. This is prior to when it is sent to lax by UpdateLAX

Also included
* new "eif" provider for some shared functions
* FakeKey object can support multiple data inputs now so it can be used by more than one test scenario
* tests for ApprovePublication activity
* expanded PreparePostEIF tests
* some tests changed with more "update" date scenarios
* resolved a merge conflict in "tests/activity/test_activity_post_eif_bridge.py" (which I hope I did correctly)


shimmy.py
```diff
@@ -90,7 +92,7 @@ def post_eif(self, eif, bucket, filename, passthrough, output_queue):
                 'article_path': passthrough.get("article_path"),
                 'expanded_folder': passthrough.get("expanded_folder"),
                 'status': passthrough.get("status"),
-                'update_date': passthrough.get("update_date"),
+                'update_date': update_date,
```
giorgiosironi: speaking about bot/website conversations, `update_date` and `published` seem to be what will have to be passed back to the bot after publication (as it's genuinely new information), while all the rest of the fields here seem local to the bot.

gnott:
I tested this branch on continuum-test today - feeding an archive zip or a new zip with publish: false or publish: true. Everything worked as far as setting and rewriting the "update" value in the JSON.

Any reason not to merge it into develop?

giorgiosironi:
(if anyone is wondering about the Jenkins build, don't worry, I just interrupted it as soon as it started because the configuration was not ready)

giorgiosironi:
The more the PR stays open, the difficult it is to integrate, so for me :+1:

PR #296, opened 2016-08-03
PR to master, hotfix. Overrides timeout on activity scheduling.

Jenniferstrej:
We needed the timeout to increase since ResizeImages for article 16039 was taking a bit longer than the limit.

@giorgiosironi - I cherry picked this commit from develop as it needs to be deployed now.

PR #297, opened 2016-08-03
PR to master, hotfix. Overrides timeout on activity scheduling.

Jenniferstrej:
We needed the timeout to increase since ResizeImages for article 16039 was taking a bit longer than the limit.

@giorgiosironi


giorgiosironi:
Closed in favor of #296

PR #304, opened 2016-09-02
Elpp 469 - article xml updated automatically to github 

Jenniferstrej:
@giorgiosironi - once this change is approved, I will need a Github token with "public_repo" access in order to be able to update the repository elife-articles-xml. I will then add the required settings to the builder. Can you provide me with that please?

Fyi The additional settings will be:
github_token
git_repo_name
git_repo_path

the function that does the update of the repo is in the file activity_UpdateRepository.py

We could then have a test with a test cdn s3 bucket (already existent), aws_incoming_queue (permission between them is also necessary). For end2end test I guess we will need a test repository and use that token to try updating it and verifying.

Does it sound right? If there is anything else I can do on my side please let me know. Thanks.

### DIFF
--- a/activity/activity_UpdateRepository.py
+++ b/activity/activity_UpdateRepository.py
@@ -61,7 +61,8 @@ class activity_UpdateRepository(activity.activity):
 
         try:
             xml_file = article_xml_repo.get_contents(filename)
-        except GithubException:
+        except GithubException as e:
+            self.logger.info("GithubException - description: " + e.message)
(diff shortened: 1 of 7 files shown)


giorgiosironi:
Let's coordinate on this, we port the production instance first to `elife-bot--prod` in the new builder, so that we only need to change the configuration there rather than in two places (old builder and new builder). Need to coordinate with Graham first.

I will send you the github token on a private channel.

giorgiosironi:
I have no blocker for merging the code however, it's just for configuring the settings that make it work in production that we should wait for the migration.

PR #308, opened 2016-09-14
I believe the session changes fall into a bug when using econ-tools a…

Jenniferstrej:
I believe the session changes fall into a bug when using econ-tools article feeder, this change should fix this. Does it look right @jhroot ?


john-root:
👍

PR #310, opened 2016-09-16
Use pid in logs

giorgiosironi:
Useful to correlate Upstart and other system-based logs with the *.log files written by the Python code


Jenniferstrej:
👍

PR #318, opened 2016-10-14
ApplyVersionNumber needs to happen in IngestArticleZip since it will …

Jenniferstrej:
…rename the files that will be validated at lax ingest stage

This should fix the problem with the file pattern being wrong. The file is changed during ApplyVersionNumber activity, then it's verified. The error was that the verification with the new bot code was happening before ApplyVersionNumber happened. I just set ApplyVersionNumber to happen right after ExpandArticle, not after ingestion.

@giorgiosironi - I may have a few questions with regards running end2end tests before merging into develop (using continuum env). I am not sure I should start the end2end machines or the continuum ones(?)

### DIFF
--- a/workflow/workflow_IngestArticleZip.py
+++ b/workflow/workflow_IngestArticleZip.py
@@ -57,6 +57,17 @@ class workflow_IngestArticleZip(workflow.workflow):
                         "start_to_close_timeout": 60 * 15
                     },
                     {
+                        "activity_type": "ApplyVersionNumber",
+                        "activity_id": "ApplyVersionNumber",
+                        "version": "1",
+                        "input": data,
+                        "control": None,
+                        "heartbeat_timeout": 60 * 10,
+                        "schedule_to_close_timeout": 60 * 10,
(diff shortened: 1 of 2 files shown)


giorgiosironi:
Depends on which environment you want to run the tests. In general `end2end` is used by the automated builds of the `develop` branches so to avoid clashes it's easier to run your PR in `continuumtest` where no build will overwrite your code.

Jenniferstrej:
If I merge this PR, will the end2end tests run with the new builder code or "new lax" won't be considered?

giorgiosironi:
At the moment I reverted the change to the settings so we are in "old lax" mode. I can change the settings, should I leave that on "new lax" in the end2end environment?

Jenniferstrej:
Lets run once with the old, as it is, if it passes, we then run with the new one. Is it possible?

giorgiosironi:
Yes, I'll make the (manual) change after this merge has passed the tests and run a new build.

Jenniferstrej:
Ok. Merging then.

PR #319, opened 2016-10-14
using boto sqs RawMessage instead of Message

Jenniferstrej:
For Lax (boto3) compatibility.

@giorgiosironi - These are the only points where the bot sends a message to Lax. I believe I don't need to do anything with relation to reading from Lax. 
Anything else I may have misunderstood?

### DIFF
--- a/activity/activity_IngestToLax.py
+++ b/activity/activity_IngestToLax.py
@@ -7,7 +7,7 @@ from boto.s3.connection import S3Connection
 from provider.execution_context import Session
 import datetime
 import boto.sqs
-from boto.sqs.message import Message
+from boto.sqs.message import RawMessage
 import provider.lax_provider as lax_provider
 
 """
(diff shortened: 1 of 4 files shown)


giorgiosironi:
Everything seems consistent: we send only JSON strings here and we should only get JSON strings on the other side. :+1:

giorgiosironi:
(it may be worth trying to lock down this choice with a unit test on the activities, if we can extract some small and easy to test method from them)

Jenniferstrej:
I added two tests in Lax provider, where the message dictionary is defined. I wanted to avoid to depend on boto Message/RawMessage unless I misunderstood your suggestion and you think we should add that dependency on the tests.

giorgiosironi:
My question is: is there a test that can tell us if we are accidentally encoding the body of the message in base64 before we go into the end2end environment?
By the way, this is not blocking for this PR as we need it to explore the integration with Lax, so :+1:

PR #321, opened 2016-10-17
Added logging in lax response adapter that gives me a string version …

Jenniferstrej:
…of a Lax response message
- For more clarity in knowing the actual message that came from lax


giorgiosironi:
:+1: I log, you log, we all log for nice logs

PR #322, opened 2016-10-19
Token seems to be built with concatenation

giorgiosironi:
Seems that there is an integer hiding in these fields, it may be the `version`:
```
2016-10-19T15:39:54Z ERROR worker_304 Exception when Preparing Publish action for Lax
  File "/opt/elife-bot/activity/activity_PublishToLax.py", line 57, in do_activity
    article_id, run, expanded_folder, version, status, eif_location, 'publish')
  File "/opt/elife-bot/provider/lax_provider.py", line 71, in prepare_action_message
    'token': lax_token(run, version, expanded_folder, status, eif_location)
  File "/opt/elife-bot/provider/lax_provider.py", line 86, in lax_token
    '"status": "' + status + '"}'
TypeError: coercing to Unicode: need string or buffer, int found
```

### DIFF
--- a/provider/lax_provider.py
+++ b/provider/lax_provider.py
@@ -2,6 +2,7 @@ import requests
 import time
 from article import *
 import base64
+import json
 
 def article_versions(article_id, settings):
     url = settings.lax_article_versions.replace('{article_id}', article_id)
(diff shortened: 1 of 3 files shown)


giorgiosironi:
It was the version:
```
{"article_id": "628573234", "version": 1, "run": "46dfea4c-411b-48b9-b3ac-b75585f2f1e5", "publication_data": "eyJ3b3JrZmxvd19uYW1lIjo...
```
(from https://console.aws.amazon.com/swf/home?region=us-east-1#execution_activities:domain=Publish.end2end;workflowId=ApproveArticlePublication_628573234.1.288;runId=228knPwpB+280Bj9aeoe4FgQTZlHREwpvVdkOe8thCow4!=)

Jenniferstrej:
@giorgiosironi - yes, concatenation was not a good idea, but I intend to have the version as a string in the token since this is the way it's used further down the bot pipeline. It's only integer for lax, and the token is an internal thing for the bot, so it should remain as string. I was misled by the mutability of the python language.

giorgiosironi:
I will just refactor how the token is built here, to make the test failures more clear; somewhere an integer is passed in the activity, which would be elsewhere.

Jenniferstrej:
I am trying to understand where it became an int as I am reading the version data from the token, which gives me a string. I only convert it to int to send to Lax...

giorgiosironi:
Will take a look tomorrow, but be aware that the point where it breaks is upon starting this workflow after the Publish button on the dashboard is pressed, so I would check the data coming in from the dashboard into one of the queues. I will add the very valuable logs we need for that communication...

Jenniferstrej:
@giorgiosironi - I added a Pull request for it now https://github.com/elifesciences/elife-bot/pull/325

PR #323, opened 2016-10-19
In PreparePostEIF, get the update_date value from the session.

gnott:
Doing a silent correction today, the workflow threw an exception. The update_date was not found in data{}. I tracked it down to this hotfix to get it from the session.

To @Jenniferstrej could you please confirm this looks ok?

To @giorgiosironi is this ok still to PR a hotfix directly to the master branch? Is there anything else required to get it applied to master directly? (if we cannot and it must go through develop branch and wait for the next deployment, that's probably ok in this case).

### DIFF
--- a/activity/activity_PreparePostEIF.py
+++ b/activity/activity_PreparePostEIF.py
@@ -56,7 +56,7 @@ class activity_PreparePostEIF(activity.activity):
             expanded_folder = session.get_value(run, 'expanded_folder')
             status = session.get_value(run, 'status')
 
-            update_date = data['update_date']
+            update_date = session.get_value(run, 'update_date')
 
             carry_over_data = {
                 'eif_filename': eif_filename,


giorgiosironi:
If you are in need of making an hotfix, you have to:
- merge the PR into `develop`
- merge `develop` into `approved` (which is what it makes it skips the test)
- run the `prod-elife-bot` pipeline in Alfred, which is fed from `approved` and which will merge to `master`.
Without running that pipeline (or the underlying builder commands if you know what they are), the code on the `elife-bot--prod` EC2 instance won't change, and if changes it will roll back to the previous version (vicious).

Tests are actually green at the moment when we disable the new ingest (production in fact is not using it). If this is merged today we will stop the new ingest tests, and it should take 15-20 minutes to bring this to production.

Jenniferstrej:
👍

gnott:
So to hotfix without also deploying the new code from the develop branch, then the approved branch would first need to be rebased as master branch, then directly PR to the approved branch?

Not a problem now, I was just curious. I will close this PR and point a new PR to develop branch, since this change is not all that critical.

PR #325, opened 2016-10-20
version should always be string on the bot

Jenniferstrej:
keeping history:
token version should always be string

The token is a way for bot to keep state between lax calls/workflows. Since version should be always string and message to lax may be triggered by the external apps (the dashboard in this case), I need to make sure version is string in the token.

@giorgiosironi - just a small change explained above.


Jenniferstrej:
Actually, I think I should have it transformed into a string earlier on, at the starter o ApproveArticlePublication, for consistency. Let me change.

giorgiosironi:
This goes into the direction of always having a string for `version` inside the bot, so :+1:

PR #326, opened 2016-10-20
When in doubt, log the input of a long running process

giorgiosironi:
Every time something is pulled off a queue or a workflow, the input to the process is logged with INFO level for troubleshooting. Valuable when integrating, see the recent "is version a string or an integer here?"

Some of these data can be pulled off the SWF console, but textual logs are easier to parse.

I will set up logrotate on the /opt/elife-bot/*.log files, to compress older logs and save space - in this case we are not adding much as the really heavy stuff are the empty polling loops which are at DEBUG level

### DIFF
--- a/decider.py
+++ b/decider.py
@@ -48,8 +48,7 @@ def decide(ENV, flag):
 
             token = get_taskToken(decision)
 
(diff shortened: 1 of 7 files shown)


Jenniferstrej:
👍

PR #333, opened 2016-10-31
Elpp 1683 support for new Lax endpoint

Jenniferstrej:
@gnott 

Lax provider endpoint changes;
More unit tests to cover more 'failing' scenarios

### DIFF
--- a/provider/article.py
+++ b/provider/article.py
@@ -659,26 +659,10 @@ class article(object):
         doi_id = int(self.get_doi_id(doi))
         article_id = str(doi_id).zfill(5)
 
-        lax_url = self.settings.lax_article_versions.replace('{article_id}', article_id)
-        response = requests.get(lax_url)
+        status_code, data = lax_provider.article_versions(article_id, self.settings)
 
-        # We will check for any published version with these article status below
-        poa_status = None
-        vor_status = None
-
-        if response.status_code == 200:
-            data = response.json()
-
-            for version in data:
-                try:
-                    status = data[version]['status']
-                except:
-                    status = None
-
-                if status and status == 'poa':
-                    poa_status = True
-                if status and status == 'vor':
-                    vor_status = True
+        if status_code == 200:
+            poa_status, vor_status = lax_provider.poa_vor_status(data)
 
         # Now a decision can be made
         if (is_poa is True or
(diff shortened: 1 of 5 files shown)


Jenniferstrej:
@gnott - Please check commit https://github.com/elifesciences/elife-bot/pull/333/commits/d395e4ec88738c6b3b4105b7f68fdd94f2d76d46 for changes in provider/article.py/check_is_article_published_by_lax

gnott:
Looking good reading the code (although it's not very simple is it!).

1. I wonder when None is returned instead of data, and then functions that try to get len(data) will throw an exception. Maybe it should return an empty list instead?
https://github.com/Jenniferstrej/elife-bot/blob/d395e4ec88738c6b3b4105b7f68fdd94f2d76d46/provider/lax_provider.py#L15

2. Do we have a test where the lax response has no "published" value for the first version? We'd need an additional "lax_article_versions_response_data" in the test where there is only one version returned and it has no "published" value (it is still unpublished). I'm assuming that is what lax would respond back with if the version 1 is still unpublished?

The changes satisfy the original return values of these functions (getting the pub date, checking for if a VoR is published, etc.) so the activities that depend on these *should* work just the same - great work there!

Jenniferstrej:
Regarding question number 2. I assumed we would not see versions that did not have a pub date. From Luke's comment http://jira.elifesciences.org:8080/browse/ELPP-1555?focusedCommentId=14190&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14190 "Lax and some other services will be aware of unpublished content and will be requiring authentication to receive unpublished content. If we don't give the bot an api key, you will only get the version history of published articles - articles with no pubdate will be excluded from the results."

Jenniferstrej:
Question number one. I believe the case of having a response 200 and None as data isn't an expected option. View https://lax.elifesciences.org/api/docs/#!/v2/Article_Version_List_GET, open GET /api/v2/articles/{id}/versions and use an example of article 0001. It will return 200 but an empty list anyway. I am going to change the examples whose mocks are returning 404 or 500 to have None as data, but things won't change. Please let me know if I missed some case.

gnott:
Q1 I see, I wasn't aware of how the unpublished v1 would be back from Lax, good detail!

Q2 I didn't notice all the checks on len(data) only happen when the status code is 200, makes my point irrelevant.

Thanks Jen!

Jenniferstrej:
Conflicts solved. @giorgiosironi - I changed the title as this is no longer blocked. So we can merge it when you are ready.

PR #336, opened 2016-11-04
FIX - ELPP-1613 adds execution lax/website on workflow id on PostPerfectPublication + tests

Jenniferstrej:
@giorgiosironi - Added where the execution comes from "lax/website" to PostPerfectPublication starter and a unit test. I did PostPerfectPublication_%s.%s.%s % (article_id, process_id, publication_from). Is that what you had in mind?

### DIFF
--- a/starter/starter_PostPerfectPublication.py
+++ b/starter/starter_PostPerfectPublication.py
@@ -29,20 +29,18 @@ class starter_PostPerfectPublication():
         # logFile = None
         logger = log.logger(log_file, settings.setLevel, identity)
 
-
         if info['article_id'] is None:
             raise NullArticleException("article id is Null. Possible error: Lax did not send back valid data from ingest.")
 
+        workflow_id, \
+        workflow_name, \
+        workflow_version, \
+        child_policy, \
+        execution_start_to_close_timeout, \
+        workflow_input = self.set_workflow_information("PostPerfectPublication", "1", None, info, os.getpid())
+
         # Simple connect
         conn = boto.swf.layer1.Layer1(settings.aws_access_key_id, settings.aws_secret_access_key)
-
-        # Start a workflow execution
-        workflow_id = "PostPerfectPublication_%s.%s" % (info['article_id'], os.getpid())
-        workflow_name = "PostPerfectPublication"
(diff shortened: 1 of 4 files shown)


Jenniferstrej:
Oops! I need to fix the settings import thing. I'll have to see if I can have it injected on the class construction...

starter/starter_PostPerfectPublication.py
```diff
@@ -56,9 +54,26 @@ def start(self, info, ENV="dev"):
...
         except boto.swf.exceptions.SWFWorkflowExecutionAlreadyStartedError:
             # There is already a running workflow with that ID, cannot start another
-            message = 'SWFWorkflowExecutionAlreadyStartedError: There is already a running workflow with ID %s' % workflow_id
+            message = 'SWFWorkflowExecutionAlreadyStartedError: ' \
+                      'There is already a running workflow with ID %s' % workflow_id
             logger.info(message)
 
+    def set_workflow_information(self, name, workflow_version, child_policy, data, process_id):
+        publication_from = "lax" if 'requested_action' in data else 'website'
+        workflow_id = "%s_%s.%s.%s" % (name, data['article_id'], process_id, publication_from)
```
giorgiosironi: `publication_from` is useful to differentiate the two workflows who have the same name and article id. I would remove `process_id` from `workflow_id` instead: `process_id` is only useful when written in a log file.

Jenniferstrej: If I remove process_id the chances of getting workflow_id's clashing are much higher. It's not unlikely that someone may publish another version of an article right after a previous one has been publish and also there is the risk from a silent correction. I know that it wouldn't happen everyday, but a combination of just article id + (lax or website) seems a bit too general to me (since the workflow_id must be unique SWF may complain even if a workflow with the same ID has already been terminated)

Jenniferstrej:
@giorgiosironi - I looks like I was wrong about thinking the workflow ids would clash after the workflow has been terminated (I tested locally). I guess it's worth seeing how it would work without the process id and if there are any problems, we can judge wether it's better to change the ids or the process...

giorgiosironi:
:+1:

PR #337, opened 2016-11-07
Minimal change for version in silent corrections

giorgiosironi:
It appears that `article_highest_version` is returning sometimes a `str`, sometimes an `int`. 

However, I'm not touching that as I see there are conditionals like:
```
if isinstance(version, (int,long)) and version >= 1:
    version = str(version + 1)
```
which rely on the type being an `int` (sometimes).

Therefore, I cast to `str` the version the activity obtains after this call.


Jenniferstrej:
👍

PR #338, opened 2016-11-07
New Relic APM on SWF activities

giorgiosironi:
We should be able to track how long each activity takes and get an aggregation of these results per activity type (changes to the formula are also needed).

Work in progress, testing it out in the right environments.

### DIFF
--- a/worker.py
+++ b/worker.py
@@ -6,6 +6,7 @@ import random
 import os
 import importlib
 import time
+import newrelic.agent
 from provider import process
 from optparse import OptionParser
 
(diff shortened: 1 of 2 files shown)


worker.py
```diff
@@ -54,60 +55,62 @@ def work(ENV, flag):
                     # Build a string for the object name
                     activity_name = get_activity_name(activityType)
```
giorgiosironi: this seems much more complex than it is, it is simple indenting all the code for activity execution as it gets wrapped into the `newrelic.agent.BackgroundTask` context manager

giorgiosironi:
@Jenniferstrej check out this for measurement of the bot's activities performances

PR #339, opened 2016-11-08
Elpp 469

Jenniferstrej:
@giorgiosironi - I believe you already have the settings in place for this, but a small thing has changed:
now git_repo_path is "/articles/"
and to recap, the other settings are:
git_repo_name and github_token

@gnott - Can you review, please? Thanks

Now the git repository is updated at the end of PostPerfectPublication. So we know for sure that the article has been published.

### DIFF
--- a/activity/activity_UpdateRepository.py
+++ b/activity/activity_UpdateRepository.py
@@ -61,7 +61,8 @@ class activity_UpdateRepository(activity.activity):
 
         try:
             xml_file = article_xml_repo.get_contents(filename)
-        except GithubException:
+        except GithubException as e:
+            self.logger.info("GithubException - description: " + e.message)
             self.logger.info("GithubException: file " + filename + " may not exist in github yet. We will try to add it in the repo.")
             response = article_xml_repo.create_file("/" + filename, "Adds XML first time", content)
             self.logger.info("File " + filename + " successfully added. Commit: " + response)
@@ -78,7 +79,7 @@ class activity_UpdateRepository(activity.activity):
                 return
 
             #there are changes
-            response = article_xml_repo.update_file("/" + filename , "Updates xml 4", content, xml_file.sha)
+            response = article_xml_repo.update_file("/" + filename , "Updates xml", content, xml_file.sha)
             self.logger.info("File " + filename + " successfully updated. Commit: " + response)
             return
 
--- a/activity/activity_UpdateRepository.py
+++ b/activity/activity_UpdateRepository.py
@@ -14,6 +14,7 @@ class activity_UpdateRepository(activity.activity):
         activity.activity.__init__(self, settings, logger, conn, token, activity_task)
 
         self.name = "UpdateRepository"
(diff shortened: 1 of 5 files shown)


giorgiosironi:
I updated `git_repo_path` in the bot settings from `articles/` to `/articles/` for all environments (it will get deployed with the first build.) You should update `settings-example.py`, too.

Jenniferstrej:
Thanks @giorgiosironi for the reminder. Updated settings_example.py and pushed.

gnott:
Q. Would we maintain multiple repositories for testing on end2end?

Maybe in the settings, for a particular environment, if "git_repo_name" is set to None, then UpdateRepository just returns true but does not attempt to commit any files?

giorgiosironi:
The `None` value will be good for `dev` and `ci` as it will make it easier to run automated tests there. For `end2end` we have instead a test repository at https://github.com/elifesciences/elife-article-xml-end2end

Jenniferstrej:
Added the following block https://github.com/Jenniferstrej/elife-bot/blob/e29736f77ffe23c8dd9ac541f59441329a3d33a1/activity/activity_UpdateRepository.py#L31-L45 
Checks if any of the Github settings are None;
If True and env = live it will report an error; 
If True and env is not live, it will simply skip the execution and terminate the activity successfully (return True)
@giorgiosironi and @gnott - Does it help?

giorgiosironi:
This seems the intended behavior. I'm not a fan of checking environment names in code as that makes it untestable (how do you run the code inside the inner `if` in a testing environment?) but if you want to do it the environments that should produce an error are `end2end`, `prod` and `live` (the last two are aliases for the same thing.)

Jenniferstrej:
I see that @giorgiosironi - I could simply remove the error bit, it would make it simpler indeed, but if there was an actual missing config in the settings, we would not find them out by an error. We would only know by noticing by chance the message "skipping update repository" after lots of articles have been fed successfully then we would need to add the XML manually to Github as a result. I guess it's a choice we have to make... 
Maybe adding a setting that will be only in dev environments and only check that? What do you think?

giorgiosironi:
I understand it's useful to have the error, not suggesting to remove it but to keep it, but they should follow the principle that `end2end` and `prod`|`live` behave in the exact same way so that we test what we give to people. Solution: add `end2end` to the `prod`|`live` if.

Jenniferstrej:
Added end2end and prod in the check along with live. We could later add a setting that will be only necessary on local dev environments that could be something like "validate_github", if False or not present, we would expect the settings to be valid, if True we would just skip the activity returning True. Up to us to decide... Anyway, the change has been made and we can try merging it if everyone is ok 👍 with it.

gnott:
The "validate_github" value could be a good option. I will defer to @giorgiosironi might have better feedback on validating things.

giorgiosironi:
True that an option would scale better to many environments, the current version is good enough for merging for me :+1:

PR #340, opened 2016-11-08
The silent corrections actually wasn’t publishing on the old elife-we…

Jenniferstrej:
…bsite. That should fix it

After looking well, I had missed the actual publishing of EIF. That should trigger the second PostPerfectPublication workflow and work with either setting (lax/elife-website).

I still need to look on the failure though. I am checking the bot logs now.


Jenniferstrej:
@giorgiosironi , as I said on the jira ticket, the logs seem fine. Shall I merge this?

giorgiosironi:
I think we shall merge, retest on `end2end` and see what is the situation now.

PR #341, opened 2016-11-10
Logging handler compatible with logrotate

giorgiosironi:
logrotate is a Unix daemon that every morning takes the log file, compresses it as worker.log.1.gz, and then creates a new empty one to be written by the application.

The Python code needs to be aware of this to reopen the file after it has been changed, otherwise logs written after the rotation will be lost.

### DIFF
--- a/log.py
+++ b/log.py
@@ -1,4 +1,5 @@
 import logging
+from logging import handlers
 import os
 import random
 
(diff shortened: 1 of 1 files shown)


Jenniferstrej:
Ok 👍

PR #342, opened 2016-11-10
Instrumenting more long-running processes

giorgiosironi:
This is the form that empirically we have seen as not impacting performance. It only measures the total time to process each message/job and external interactions like S3 calls.


Jenniferstrej:
Looks ok 👍

PR #343, opened 2016-11-11
Make sure eif_location is passed on to PostPerfectPublication when do…

Jenniferstrej:
…ing a silent correction

This problem hasn’t happened yet. The issue that this code fixes is when a silent correction is ran and the “publication_authority” is set to lax.

@gnott & @giorgiosironi :
I'm not sure I was able to be clear enough in the code comments. 
I noticed that if we have a silent correction and the publication authority is set to lax, "PublishToLax" activity wouldn't have access to the attribute eif_location to send to PostPerfectPublication. In the normal workflows it is possible because PreparePostEIF sets the data dictionary with the eif location attribute in the last activity of the Ingest workflow and PublishToLax has access because it is part of the "Process..." workflow then and the data had been mutated before starting the ProcessArticleZip workflow.
It's a problem that we need to rely on the session when in the first workflow as we cannot mutate data across activities. I wasn't able to find enough documentation when I first searched AWS SWF but I believe it should be possible. I just don't know yet how I can get a changed data from the end of an activity and send to another. That way we wouldn't need a session at all. But for now, I keep the first activity with a session and the others I try to only use the data... In future hopefully we can just unplug the session from the "...Ingest..." workflows.

@gnott - could you review please?

Thanks.


giorgiosironi:
Agreed, hasn't happened yet but will in the future if not fixed.

gnott:
I'm trying to follow where ``eif_filename`` gets set originally, maybe in ``ConvertJATS``. This activity here though also looks for ``eif_location`` (https://github.com/Jenniferstrej/elife-bot/blob/9ae64002ffb70538c74d4f34d883cdc3b5c2e318/activity/activity_PublishToLax.py#L49) and think it will find that to be blank too?

I guess there was a global rename of ``eif_location`` to ``eif_filename``, yes @Jenniferstrej ? I didn't notice that before. If it was renamed, I think it needs to be traced back and confirmed changed in all places.

Jenniferstrej:
Hi @gnott, yes, the variables are used in different places. eif_filename was used up to PreparePostEIF and when it leaves the session it is treated as eif_location. However since it cause you this doubt, I decided the change it almost everywhere (except when it is just passed as a parameter to a function (in this case the name doesn't really matter. I saw variations such as being called"filename" for example) See commit https://github.com/elifesciences/elife-bot/pull/343/commits/1d496a50cea1bfe677c666b2dafc9d27a720d07f

PR #345, opened 2016-11-14
Let’s try to have Lax reporting a publishing error event to the dashb…

Jenniferstrej:
…oard even if the publication authority is the website.

We could try monitoring this case and see if we would be able to republish in both sites (fixing lax and double publication on elife-website) without problems and a clear message on the dashboard.

Related to the silent corrections run we made on Friday @gnott . Luke made a fix on his side and I am adding an error report on the dashboard for next time we can be aware of errors without only relying on finding the logs.

@giorgiosironi - for consideration in end2end..

### DIFF
--- a/activity/activity_VerifyPublishResponse.py
+++ b/activity/activity_VerifyPublishResponse.py
@@ -45,15 +45,26 @@ class activity_VerifyPublishResponse(activity.activity):
 
                 if 'requested_action' in data:
                     # Publication authority is the old site but this call is from new lax.
-                    # Terminate Workflow gracefully, log
-                    start_event = [self.settings, article_id, version, run, self.pretty_name + ": Lax", "start",
+                    start_event = [self.settings, article_id, version, run, self.pretty_name + ": Journal", "start",
                                    "Starting verification of Publish response " + article_id]
 
-                    return [start_event
-                            , [self.settings, article_id, version, run, self.pretty_name + ": Lax", "end",
-                               "Finish verification of Publish response. Authority: elife-website. Exiting "
-                                "this workflow " + article_id]
-                            , activity.activity.ACTIVITY_EXIT_WORKFLOW]
+                    # Before terminating Workflow gracefully, emit the result of publication to lax on the dashboard
+                    # Terminate Workflow gracefully, log
+                    if data['result'] == "published":
+                        return [start_event,
+                                [self.settings, article_id, version, run, self.pretty_name + ": Journal", "end",
+                                 " Finished Verification. Lax has responded with result: published."
+                                 " Authority: elife-website. Exiting."
+                                 " Article: " + article_id],
+                                activity.activity.ACTIVITY_SUCCESS]
+                    else:
+                        return [start_event,
+                                [self.settings, article_id, version, run, self.pretty_name + ": Journal", "error",
+                                 " Lax has not published article " + article_id +
+                                 " We will exit this workflow as the publication authority is elife-website."
+                                 " result from lax:" + str(data['result']) + '; message from lax: ' +
+                                 data['message'] if ("message" in data) and (data['message'] is not None) else "(empty message)"],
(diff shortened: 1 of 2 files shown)


giorgiosironi:
To end2end-test this, I would need an article sample that is ingested into Lax, but fails on publishing. How do I obtain such an article? Simple errors like invalid XML don't seem to cut it.

gnott:
@giorgiosironi that's a challenge to get an ingested article to fail publication. On my search, I think this is controlled in Lax, so you might need to edit Lax data internally in order to simulate that situation, if it can ever happen. I don't see a way it could happen normally. You might be able to simulate it by setting the versionDate to null after ingest for that version?

@Jenniferstrej I think I see where this will send the error message to the dashboard when Journal is not the publishing authority. I wonder though where it potentially returns ``ACTIVITY_PERMANENT_FAILURE`` then it really wouldn't be respecting the pub_authority: if pub_authority is "elife-website" and Journal publishing fails, then the entire workflow would be cancelled after the ``ACTIVITY_PERMANENT_FAILURE`` is issued? Would it be better to report a different activity return value to allow the workflow to continue?

giorgiosironi:
Since we are on the subject of `ACTIVITY_PERMANENT_FAILURE`: currently in the happy path half of the `PostPerfectPublication` workflows are declared failed:
```
PostPerfectPublication_20713.website	
22EAOpEDtnmUW+7MXavQsJTo/sCDSu5C7YPo6HAF+UaGA=
PostPerfectPublication (1)
Completed
Tue Nov 15 08:00:14 GMT+000 2016
Tue Nov 15 08:00:28 GMT+000 2016
	
PostPerfectPublication_20713.lax	
22778JIQM+yJ8wRp9LYBN/88LSckeEo0vyDbSW9tDho2c=
PostPerfectPublication (1)
Failed
Tue Nov 15 08:00:12 GMT+000 2016
Tue Nov 15 08:00:12 GMT+000 2016
```
The last activity that we see in the lax failed workflow is `VerifyPublishResponse`:
```
VerifyPublishResponse
1
Started
Tuesday, November 15, 2016 8:00:12 AM UTC
Tuesday, November 15, 2016 8:00:12 AM UTC
none
```
and this activity is not reported explicitly as failure, but just as `Started`. This is confusing as when debugging you never know if it's normal that the workflow failed or not.

Jenniferstrej:
@giorgiosironi - where are you getting this information from? AWS console, logs? What I do is to terminate the workflow at the end of VerifyPublishResponse when the activity is not triggered by the publication authority. The dashboard gets an "end" event message, which marks it visually as successful. If the information you get is from AWS console, I would need to find a way to implement a workflow that does not end when it's supposed to and that doesn't mean that it's a failure (not sure if it's possible or not, it's also not currently implemented in the bot worker);

@gnott - I wasn't sure if I should not display an error in case it is not from the publication authority. I could also just display the error message from Lax without displaying an "error" event in the dashboard but @lsh-0 and @giorgiosironi suggested that there should be an error displayed in the dashboard (since we still run both publishing). I think that it would not happen the way you think, I believe the other events would appear correctly since they would have come from another workflow (PublishPerfectPublication triggered by "elife-website" - this one should be successful). I'm not sure I am very clear and we could, in our call tomorrow talk about how the dual publication works and that may answer some questions you may have.

What I am still not fully clear is that if we re-published both elife-website and Lax, after an error on the new Jounal side, would we break anything on the elife-website, since it would be publishing a second time? @gnott , would you be able to answer the question? @nlisgo , Luke asked on this comment on Jira http://jira.elifesciences.org:8080/browse/ELPP-1613?focusedCommentId=15468&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15468, maybe you can help us?

gnott:
I'm ok with the logic as long as you thought it through. I forgot about the split workflows. :+1:

PR #346, opened 2016-11-14
I previously thought when Lax responded with status 200, the response…

Jenniferstrej:
… would never be None, but I assumed wrong. This Fix is for that.

@giorgiosironi - It seems that the error reported here http://jira.elifesciences.org:8080/browse/ELPP-1613?focusedCommentId=15474&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15474 was a wrong assumption that I explain on the title of this PR.

### DIFF
--- a/provider/lax_provider.py
+++ b/provider/lax_provider.py
@@ -12,6 +12,8 @@ def article_versions(article_id, settings):
         data = response.json()
         if "versions" in data:
             return status_code, data["versions"]
+        if data is None:
+            return status_code, []
     return status_code, None
 
 


giorgiosironi:
Are we hiding a problem or is it a perfectly fine response? @lsh-0 is that a 200 with an empty body?

Jenniferstrej:
Apparently this is a bug in Lax, @lsh-0 - Jamie is also having some problem related to that.

giorgiosironi:
Not sure about this, but now that the configuration has `/api/v2/articles/:articleId/versions` all tests pass again. So this PR may (not necessarily) be an improvement, but it's not necessary for being green.

Jenniferstrej:
I guess I should close it then. However I am still pretty sure this may be an error in future as Lax 200 responses may give empty bodies. 
In either case, like you said, this PR would just be hiding the problem.

PR #348, opened 2016-11-16
Finally figured out how to monitor short scripts

giorgiosironi:
The application has to be forced to be started, in order for it to be ready
to collect data when the short task `run_cron` is executed.


giorgiosironi:
From https://discuss.newrelic.com/t/monitor-one-off-scripts/35407/2

PR #349, opened 2016-11-16
Fix for wrong behaviour that I introduced with previous PR

Jenniferstrej:
pr: https://github.com/elifesciences/elife-bot/pull/345

I left a whole and one test was missing. Added.


giorgiosironi:
:+1: it would be helpful to extract some private methods since this is a very long one

PR #352, opened 2016-11-18
Logging SWF exceptions contents in worker.py

giorgiosironi:
This error happened in the `end2end` environment:
```
2016-11-18T13:18:42Z INFO worker_3059 SWFResponseError:
SWFResponseError: 400 Bad Request on respond_completed
```
but the log message was hardcoded instead of showing what the exception was about.


Jenniferstrej:
👍

giorgiosironi:
I'll merge this and vanish until monday.

PR #353, opened 2016-11-18
Version date lookup for silent corrections

Jenniferstrej:
@gnott - could you please review? thanks

### DIFF
--- a/activity/activity_ExpandArticle.py
+++ b/activity/activity_ExpandArticle.py
@@ -67,11 +67,6 @@ class activity_ExpandArticle(activity.activity):
                               filename_last_element)
             return activity.activity.ACTIVITY_PERMANENT_FAILURE  # status could not be determined, exit workflow.
 
-        # Extract and store updated date if supplied
-        update_date = article_structure.get_update_date_from_zip_filename()
-        if update_date:
-            session.store_value(run, 'update_date', update_date)
-
         article_version_id = article_id + '.' + version
         session.store_value(run, 'article_version_id', article_version_id)
         session.store_value(run, 'run', run)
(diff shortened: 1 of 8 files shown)


gnott:
Is there more settings for an additional Lax endpoint to be used? It seems as though it uses the same versions endpoint as the others, but the test data is looking like that from a different endpoint? (I could be wrong, you could explain)

Is anything but the tests using lax_provider.article_version_date() ?

gnott:
I forgot to add, I like where it will take the date from the zip file name if available first, and if that is not availbale then it looks up the value in Lax :+1:

giorgiosironi:
The bot is configured to use `lax_article_versions = 'https://{environment}--lax.elifesciences.org/api/v2/articles/{article_id}/versions'` in all environments.

Jenniferstrej:
@gnott - it may be confusing because I also added the implementation for the other endpoint, that will be used when Luke has fixed it. 
But for now, the only used endpoint is the article versions one: https://github.com/Jenniferstrej/elife-bot/blob/1f1165b99bcfecc205cea316f709cd0fd9be3534/provider/lax_provider.py#L51

I forgot to add the new activity in the silent corrections ingest workflow definition and remove the implementation of date lookup in "ExpandArticle". I believe we don't need this update_date lookup in the normal workflow, do we? I made a commit for these changes.

Jenniferstrej:
@gnott - I guess you are right, the different endpoint implementation is making things confusing. I removed the implementation and commented the test.

gnott:
Thanks for these. 

I agree the normal ingest workflow will need to lookup a version date, only in the case of a silent correction would we take it from the file name or look it up.

Removing and commenting out the functions that are not yet supported on the second Lax endpoint is much clearer to merge now.

:+1:

PR #354, opened 2016-11-22
changes message that represents creation of XML .

Jenniferstrej:
@giorgiosironi - that's the commit that changes the message from "Adds XML first time" to "Creates XML".


giorgiosironi:
:+1:

PR #356, opened 2016-11-22
Fix in logger

Jenniferstrej:


giorgiosironi:
`logger.exception` (confusingly) takes a message, not an exception as it should get it from the global state: https://docs.python.org/2/library/logging.html#logging.Logger.exception

Jenniferstrej:
Oops, ok, I guess this one will fail too. I ran it quickly in my machine and it seemed to have worked... 
I'll write a fix, sorry.

PR #358, opened 2016-11-23
Related the problem on PI-206

Jenniferstrej:
There is no point in having an error event on Version Lookup. If there is an error here, we will have no version to log it against, so it’s better to throw an exception.
plus fixed tests

@gnott @jhroot - This is the bot part of dealing with errors in the call for version from Lax. There are some fixes in the tests too.

### DIFF
--- a/activity/activity_ExpandArticle.py
+++ b/activity/activity_ExpandArticle.py
@@ -73,10 +73,7 @@ class activity_ExpandArticle(activity.activity):
         session.store_value(run, 'status', status)
         self.emit_monitor_event(self.settings, article_id, version, run, "Expand Article", "start",
                                 "Starting expansion of article " + article_id)
-        self.set_monitor_property(self.settings, article_id, "article-id", article_id, "text")
-        self.set_monitor_property(self.settings, article_id, "publication-status",
-                                  "publication in progress", "text",
-                                  version=version)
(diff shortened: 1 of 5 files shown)


gnott:
I suppose that makes sense if we cannot add a message to the dashboard without a version. I guess that means it will just not appear at all on the dashboard if there is no version, so we would need to check the bot logs to diagnose in that case?

giorgiosironi:
`Logger.exception` logs with `ERROR` level, so that's the best that we can do in giving feedback to developers if we can't tell the user. We will get to have these errors monitored upon, is there a way with the `logging` module to provide an Handler that receives all exceptions? From there we have Loggly and New Relic waiting to collect those.

Jenniferstrej:
@gnott  - Unfortunately the dashboard events currently only work with run, article id AND version provided. So, like Giorgio said, we would need to rely on logs on this first activity at least.

@giorgiosironi - I believe logging.handlers HTTPHandler class is what you are looking for? 

Shall I merge?

giorgiosironi:
For me ok to merge.

The problem was not having an handler that can talk to an external service, but getting a hold of the exception rather than the record in the handler. Is it documented somewhere?

gnott:
:+1:

PR #359, opened 2016-11-28
Elpp 1656 - starter refactoring and tests

Jenniferstrej:
Tests were added in starters I am familiar with and some degree of refactoring has been added to all of them (settings injection).


gnott:
I had a quick read through. I wonder, if all the starters are already objects (if that is true across the board), would it make sense to create a Starter object and then inherit from it? Then the starter_helper details would go into that base starter object.

giorgiosironi:
I agree @gnott, there is a lot of duplication in the starters like all the `identity = "starter_%s" % int(random.random() * 1000)` lines. A single superclass will extract this code so that it is consistent across all of them and we only need to tweak it in one place.

Jenniferstrej:
Instead of creating a superclass I created `starter_helper` which is implemented by all starters that I could write tests for at the moment. The others can be refactored and use `starter_helper`. I only haven't done it yet for the other starters because I don't know much about them and we created another task to gather data so then I can refactor and add tests for each. Any reason to use a superclass over `starter_helper`?

giorgiosironi:
Same reasoning (limiting duplication), and composition is usually even more flexible then inheritance as you can compose many different helpers if needed but have only one superclass. :+1: for the helper since it solves the same problem.

PR #360, opened 2016-12-01
copies contents added to elife-publishing-cdn to elife-published/arti…

Jenniferstrej:
…cles

plus
Fixes content-type download problem: When loading a key from S3, only a flat copy is made and we don’t have access to the current metadata in this copy, so content-type was being lost. Now I add the content-type when adding content-disposition.

The copy is made in the following pattern

elife-published/articles/[article id]

### DIFF
--- a/activity/activity_DepositAssets.py
+++ b/activity/activity_DepositAssets.py
@@ -72,13 +72,15 @@ class activity_DepositAssets(activity.activity):
                     dict_metadata = {'Content-Disposition':
                                      str("Content-Disposition: attachment; filename=" + file_name + ";"),
                                      'Content-Type': content_type}
+                    file_download = file_name_no_extension + "-download." + extension
 
-                    dest_resource = storage_provider + cdn_bucket_name + "/" + article_id + "/" + \
-                                    file_name_no_extension + "-download." + extension
-                    additional_dest_resource = storage_provider + published_bucket_path + "/" + article_id + "/" + \
-                                               file_name_no_extension + "-download." + extension
-                    storage_context.copy_resource(orig_resource, dest_resource, additional_dict_metadata=dict_metadata)
-                    storage_context.copy_resource(orig_resource, additional_dest_resource)
+                    orig_resource_download = dest_resource
+                    dest_resource_download = storage_provider + cdn_bucket_name + "/" + article_id + "/" + file_download
+                    additional_dest_resource_download = storage_provider + published_bucket_path + "/" \
+                                                        + article_id + "/" + file_download
+                    storage_context.copy_resource(orig_resource_download, dest_resource_download,
+                                                  additional_dict_metadata=dict_metadata)
+                    storage_context.copy_resource(dest_resource_download, additional_dest_resource_download)
 
 
 
--- a/activity/activity_DepositAssets.py
+++ b/activity/activity_DepositAssets.py
@@ -78,8 +78,11 @@ class activity_DepositAssets(activity.activity):
                     dest_resource_download = storage_provider + cdn_bucket_name + "/" + article_id + "/" + file_download
                     additional_dest_resource_download = storage_provider + published_bucket_path + "/" \
                                                         + article_id + "/" + file_download
+
+                    # file is copied with additional metadata
                     storage_context.copy_resource(orig_resource_download, dest_resource_download,
                                                   additional_dict_metadata=dict_metadata)
+                    # additional metadata is already set in origin resource so it will be copied accross by default
                     storage_context.copy_resource(dest_resource_download, additional_dest_resource_download)
 
 
--- a/activity/activity_DepositAssets.py
+++ b/activity/activity_DepositAssets.py
@@ -50,7 +50,7 @@ class activity_DepositAssets(activity.activity):
 
             storage_context = StorageContext(self.settings)
             storage_provider = self.settings.storage_provider + "://"
-            published_bucket_path = self.settings.publishing_buckets_prefix + self.settings.published_bucket_path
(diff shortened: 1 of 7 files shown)


activity/activity_DepositAssets.py
```diff
@@ -46,20 +48,41 @@ def do_activity(self, data=None):
...
+                    content_type, encoding = guess_type(file_name)
+                    dict_metadata = {'Content-Disposition':
+                                     str("Content-Disposition: attachment; filename=" + file_name + ";"),
+                                     'Content-Type': content_type}
+
+                    dest_resource = storage_provider + cdn_bucket_name + "/" + article_id + "/" + \
+                                    file_name_no_extension + "-download." + extension
+                    additional_dest_resource = storage_provider + published_bucket_path + "/" + article_id + "/" + \
+                                               file_name_no_extension + "-download." + extension
+                    storage_context.copy_resource(orig_resource, dest_resource, additional_dict_metadata=dict_metadata)
```
giorgiosironi: not sure while this `addictional_dict_metadata` is only necessary for one of the two `copy_resource()` calls

provider/storage_provider.py
```diff
@@ -44,18 +44,48 @@ def get_resource_as_string(self, resource):
...
+                metadata[mdk] = additional_dict_metadata[mdk]
+
+        dest_bucket, dest_s3_key = self.s3_storage_objects(dest_resource)
+
+        dest_key = dest_bucket.get_key(dest_s3_key, validate=True)
+        if dest_key is None:
+            dest_key = dest_bucket.new_key(dest_s3_key)
+            dest_key.set_contents_from_string('')
+
+        dest_bucket.copy_key(dest_key.name[1:], orig_bucket.name, orig_key.name[1:], metadata=metadata)
```
Jenniferstrej: You mean, not using "orig_bucket.get_key"? That makes sense now. I tried this way because I wasn't getting the key name right so I wanted to get it from the object (it didn't help much since the object key name is not the one accepted on copy_key (I need to remove the '/').
I will remove "orig_bucket.get_key" then.

activity/activity_DepositAssets.py
Jenniferstrej: I am going to merge some small fixes for this one. Thanks!

Jenniferstrej:
@giorgiosironi - made a few changes based on your observations. Thanks!

activity/activity_DepositAssets.py
```diff
@@ -46,20 +48,43 @@ def do_activity(self, data=None):
...
+                    dict_metadata = {'Content-Disposition':
+                                     str("Content-Disposition: attachment; filename=" + file_name + ";"),
+                                     'Content-Type': content_type}
+                    file_download = file_name_no_extension + "-download." + extension
+
+                    orig_resource_download = dest_resource
+                    dest_resource_download = storage_provider + cdn_bucket_name + "/" + article_id + "/" + file_download
+                    additional_dest_resource_download = storage_provider + published_bucket_path + "/" \
+                                                        + article_id + "/" + file_download
+                    storage_context.copy_resource(orig_resource_download, dest_resource_download,
```
giorgiosironi: I get from your commit messages that for some reason on one of the two copies the `dict_metadata` are already set. This isn't clear from the code, can you add some comments here and in the other activity where only one operation is passed `dict_metadata`?

giorgiosironi:
Observations made on the simmetry of the two copies, but I defer to @gnott for the final review

gnott:
Looks clear - is the "self.settings.published_bucket_path" in the settings-example.py file? Other than that :+1:

gnott:
:+1:

settings-example.py
```diff
@@ -44,6 +44,8 @@ class exp():
     archive_bucket = 'elife-publishing-archive'
     xml_bucket = 'elife-publishing-xml'
 
+    published_bucket_path = 'elife-published/articles'
```
giorgiosironi: I'm not sure where elife-published comes from, but there is no CDN (CloudFront configuration to serve files) attached to it, and it is an empty bucket having been created on 10/11/2016 with no correspondence on other environments.
The CDN I think Luke is talking about is `cdn.elifesciences.org` which points to `elife-cdn` (`end2end-elife-cdn` in testing) and is used without attaching a prefix. It is already configured at `settings.cdn_bucket` so that should be the configuration to refer to.
@lsh-0 do you confirm?

giorgiosironi: (the `/articles` string to use as a path can probably be hardcoded and afterwards being extracted in a `settings` key)

giorgiosironi:
Just creating the buckets through builder, then merging.

giorgiosironi:
Produces this across all articles:
```
2016-12-08T16:34:23Z ERROR worker_4571 Exception when resizing images
  File "/opt/elife-bot/activity/activity_ResizeImages.py", line 72, in do_activity
    self.process_key(key, cdn_path)
  File "/opt/elife-bot/activity/activity_ResizeImages.py", line 110, in process_key
    self.generate_images(formats, fp, info, cdn_path)
  File "/opt/elife-bot/activity/activity_ResizeImages.py", line 142, in generate_images
    self.store_in_cdn(filename, image, cdn_path, download)
  File "/opt/elife-bot/activity/activity_ResizeImages.py", line 178, in store_in_cdn
    additional_dict_metadata=dict_metadata)
  File "/opt/elife-bot/provider/storage_provider.py", line 76, in copy_resource
    metadata[mdk] = additional_dict_metadata[mdk]
```

giorgiosironi:
https://github.com/elifesciences/elife-bot/pull/368

PR #361, opened 2016-12-01
only logs at info level when something more relevant has happened

Jenniferstrej:
@giorgiosironi - now the logs should appear like this

2016-12-01T12:41:50Z INFO decider_5266 polling for decision...
2016-12-01T12:42:28Z INFO decider_17416 polling for decision...
2016-12-01T12:43:28Z INFO decider_17416 polling for decision...

until an event happens.

### DIFF
--- a/decider.py
+++ b/decider.py
@@ -52,7 +52,10 @@ def decide(ENV, flag):
 
             token = get_taskToken(decision)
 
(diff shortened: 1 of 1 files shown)


giorgiosironi:
:+1:

PR #362, opened 2016-12-01
Missing registration of activity

giorgiosironi:
We got a huge log containing:
```
$ grep DOES_NOT_EXIST decider.log -A 4 -B 4 | grep name | uniq -c
 933661                     "name": "VersionDateLookup",
```
but there are also other error logs elsewhere when we try to test the
silent corrections. The test still do not find out everything has
failed.

Problems should be limited to the \`end2end\` environment.


Jenniferstrej:
I'd forgotten to commit that for sure :-/. 👍

PR #363, opened 2016-12-01
Conflicting logger name

giorgiosironi:
This logger name conflicts with the `myapp` used in `parseJATS.py` from `elifetools`. Therefor, with this conflict in place all the logs from the bot are also duplicated to the `test.log` file, leading to confusing grep commands.


Jenniferstrej:
👍

PR #364, opened 2016-12-01
Getting exceptions onto New Relic

giorgiosironi:
Highly experimental, but this should get the exception being handled from Python's global stack, and upload it on New Relic for monitoring and alerting


giorgiosironi:
Trying it out on `end2end`.

PR #366, opened 2016-12-08
Fixing missing force flag set to True

Jenniferstrej:
It was missed when refactoring the class.

@giorgiosironi - this should fix the problem on end2end, I confirm I missed that setting on the starter when I was refactoring it.


giorgiosironi:
Merging upon green. Any unit test that can be added here?

Jenniferstrej:
I'm trying to think here, I have a test on the helper, where the workflow_input is put together, but a general test wouldn't know what is specifically needed for SilentCorrectionsIngest. I'm thinking I could try to mock the helper and check the information sent to it. That could do it. Let me try..

PR #367, opened 2016-12-08
Logging SQS messages send to Lax

giorgiosironi:
Lightweight, but very useful in debugging.

This screams for an abstraction to be extracted for composing the message, but I'm not ambitious enough.


Jenniferstrej:
Yes, I agree the write_message method could be extracted and reused. It can be added to my support Todo list. 👍

PR #368, opened 2016-12-08
Initializing metadata to empty dict

giorgiosironi:
Pylint finds the problem, so adding the file to the list of linted ones:
```
************* Module provider.storage_provider
E: 76,16: Value 'metadata' is unsubscriptable (unsubscriptable-object)
```

### DIFF
--- a/lint.sh
+++ b/lint.sh
@@ -3,4 +3,4 @@ set -e
 source venv/bin/activate
 
 # intentionally only the script files in the root folder
-pylint -E *.py
+pylint -E *.py provider/storage_provider.py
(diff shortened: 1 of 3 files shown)


Jenniferstrej:
I think I tried starting the metadata with an empty dict but I believe the behaviour is different since when I pass metadata=None to the copy_key function it simply doesn't change the metadata, which is the behaviour I want. I believe if metadata is passed as `{}` it may erase the existing metadata for that key.

giorgiosironi:
Writing a couple of tests to capture this...

giorgiosironi:
Tomorrow I'll extend this to cover the other case, cleaning up the test.

giorgiosironi:
Ok, ready for review. This may fix the end2end environment, which is currently red because of the exception.

giorgiosironi:
@lsh-0 has some pressing concerns about going in production, so merging to check the end2end tests become green, but feel free to comment to request changes later.

PR #381, opened 2017-01-12
Improving error message on bad file structure

giorgiosironi:
When files inside the zip are put in a subfolder, a cryptic error message results from the ApplyVersionNumber about concatenating a string with a None. This should provide a nicer error message, and earlier in the flow.


Jenniferstrej:
The only point to be careful is to not forget to provide article, version and run so then the event appears on the right place on the dashboard (through emit_monitor_event).  The code added doesn't seem to change that so 👍

PR #383, opened 2017-01-16
Fixes XML that wasn’t being found by Article class

Jenniferstrej:
It was necessary to put VerifyGlencoe happening after ApplyVersionNumber since the checks in Article class are made on an assumption that -v[number] has been added to the file, which only occurs on ApplyVersionNumber activity.

### DIFF
--- a/activity/activity_VerifyGlencoe.py
+++ b/activity/activity_VerifyGlencoe.py
@@ -50,7 +50,7 @@ class activity_VerifyGlencoe(activity.activity):
             expanded_bucket = self.settings.publishing_buckets_prefix + self.settings.expanded_bucket
             self.logger.info("expanded_bucket: " + expanded_bucket)
 
-            xml_filename = lax_provider.get_xml_file_name(self.settings, expanded_folder, expanded_bucket)
+            xml_filename = lax_provider.get_xml_file_name(self.settings, expanded_folder, expanded_bucket, version)
             if xml_filename is None:
                 raise RuntimeError("No xml_filename found.")
 
--- a/workflow/workflow_IngestArticleZip.py
+++ b/workflow/workflow_IngestArticleZip.py
@@ -68,17 +68,6 @@ class workflow_IngestArticleZip(workflow.workflow):
                         "start_to_close_timeout": 60 * 15
                     },
                     {
-                        "activity_type": "VerifyGlencoe",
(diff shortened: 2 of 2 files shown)


giorgiosironi:
We had only 1 failure in https://ci--alfred.elifesciences.org/job/test-elife-bot/126/console which is promising (may not be related).

giorgiosironi:
No, it is related:
https://console.aws.amazon.com/swf/home?region=us-east-1#execution_activities:domain=Publish.end2end;workflowId=IngestArticleZip_elife-7534800704167100230-vor-r1.zip.3014;runId=22C+C9zRbXS31IX46TtAMCJNol1KnOav7IMXM4W8v2X+M!=
The log says:
```
2017-01-16T16:24:02Z INFO worker_2984 activityType: VerifyGlencoe
2017-01-16T16:24:02Z INFO worker_2984 data: {
    "bucket_name": "end2end-elife-production-final", 
    "event_name": "ObjectCreated:CompleteMultipartUpload", 
    "event_time": "2017-01-16T16:05:50.342Z", 
    "file_etag": "f75b5fbbbdaa4147a55e6d0359ae5d3d-9", 
    "file_name": "elife-7534800704167100230-vor-r1.zip", 
    "file_size": 71045759, 
    "run": "e643f0f9-842e-4b62-91f3-ef984e8e2aed", 
    "version_lookup_function": "article_next_version"
}
2017-01-16T16:24:02Z INFO worker_2984 expanded_bucket: end2end-elife-publishing-expanded
2017-01-16T16:24:03Z INFO worker_2984 article has no videos
```

Jenniferstrej:
@giorgiosironi - this is expected, not an error, but we may need to put the timeout time on the settings file, since it's set for 5 hours now. I will add the setting and let you know the name of the variable. It also sleeps for 60 seconds after a temporary failure, should I parametrize that too so then you can have a different value on end2end?

giorgiosironi:
So this would be expected behavior, but this `7534800704167100230` testing article should have been mapped to `00230` by https://github.com/elifesciences/elife-bot/pull/380/files#diff-47fb2106ab7e41743d1958b09bbcbe56R89. `00230` is the original template obtained by using the last 5 ciphers), and that original template has videos, so the expected behavior for this article is that it sees the video of `00230` and the activity immediately succeeds.

PR #387, opened 2017-01-19
Starter should log an error if PostPerfectPublication workflow cannot start

giorgiosironi:
This has happened sometimes in tests that are too eager to publish a new version of an article while the old run has not finished yet.

If this workflow cannot start, it is simply dropped and no error is seen on
the dashboard or on the SWF console, but the end result is updates like the
Github repository are missing.


Jenniferstrej:
👍

PR #388, opened 2017-01-23
Added the whole code of VersionLookup within a try except

Jenniferstrej:
There is no error event on this one as we most likely won’t have the version information to set the event to so it makes sense to put everything under try/except as any error should simply cause a permanent failure and not issue a dashboard event.


giorgiosironi:
Improves the situation, as it stops the continuous retry

PR #389, opened 2017-01-24
WIP Elpp 2047

Jenniferstrej:
Placeholder for PDF covers e-mail solution.

### DIFF
--- a/activity/activity_GeneratePDFCovers.py
+++ b/activity/activity_GeneratePDFCovers.py
@@ -1,0 +1,60 @@
+import activity
+import provider.article as articlelib
+
+"""
+activity_GeneratePDFCovers.py activity
+"""
+
+class activity_GeneratePDFCovers(activity.activity):
+    def __init__(self, settings, logger, conn=None, token=None, activity_task=None):
+        activity.activity.__init__(self, settings, logger, conn, token, activity_task)
+
+        self.name = "GeneratePDFCovers"
+        self.pretty_name = "Generate PDF Covers"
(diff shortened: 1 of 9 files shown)


giorgiosironi:
Definitely blocked until the personalized-covers project has a formula and runs in the various environments (they didn't exactly give us a priority to do that however).

Jenniferstrej:
Settings to be added to the environments are mentioned on settings-example file
+    pdf_cover_generator = "http://localhost:8082/personalcover/"
+    pdf_cover_landing_page = "http://localhost:8082/personalcover/options"

Anything else you need before I merge this @giorgiosironi ?
Any observations @gnott ?

Jenniferstrej:
Actually I should change the examples to ones that are more similar to the URL pattern now. 1 sec

gnott:
If any exception is raised during the Generate PDF Covers, does that mean it will permanently fail the workflow and not continue with any of the following activities in the workflow? Is there a case where the activity should be retried before permanently failing?

Jenniferstrej:
@gnott - currently the endpoint that I call to generate the covers doesn't return anything but 500 when there is an error so it makes it a little difficult to identify when to retry or not. I have been talking to the guys that built the app to enhance it. I am planning to merge this now and we can see how many errors we get on end2end that we can expect and then take any action from there. 

A small change I will do to make it not harmful to anything else is putting this activity as the last one on the PostPerfectPublication.

gnott:
As the last activity that's a good plan for any minimal impact :+1:

PR #390, opened 2017-01-25
Provide an exception message showing the problem when Lax is not available

giorgiosironi:
Currently `lax_provider` relies on a logger being passed in, but almost no activity actually passes it to `article_highest_version`. Therefore, throw an exception instead as soon as the status code is recognized to be invalid.

The exception is then logged by the appropriate activity.


Jenniferstrej:
👍 👍 👍

PR #393, opened 2017-02-08
Logging requests and responses from Lax

giorgiosironi:
The only call that is made is to the /articles/:id/versions endpoint. Calls will have a separate log file


giorgiosironi:
Had to use a fixed logging level because the module doesn't know what is the environment at loading time (Python modules are the new singletons), but it works.

PR #394, opened 2017-02-16
Prevent multiple workflows running concurrently

giorgiosironi:
AFAIK, two workflows with the same id will be stopped from running (the
second one will be dropped). For these workflows it seems safe to say we
don't want more than one copy running at a given time.

The pid remains useful for the identity of processes (like worker.py),
which will continue to log it. It's also logged as the `Identity` field
of all the activities on the SWF console, as different activities are
performed by different workers (e.g. `worker_31510`)


Jenniferstrej:
I'm trying to think if there will be any gotchas in that but we should be able to see them on end2end and I don't see why it's not worth trying so 👍

gnott:
It is one of the good features of SWF is having no duplicate runs when using the same runId :+1:

PR #397, opened 2017-02-23
Article structure provider to recognise file types from the new kitch…

gnott:
…en sink, and should be compatible with all previous file types. Includes tests!


gnott:
If you get a chance @Jenniferstrej @giorgiosironi if this looks ok, I am hoping it will handle all the old and new file formats for the files we care about.

gnott:
I will hope for the best it does not cause any issues, merging now. If you want to review and send feedback we can change it.

PR #399, opened 2017-03-02
Retry on SSL timeout from Github

giorgiosironi:
Also:
- Using storage_provider for S3 access
- Stop reading the file to a tmp file that is subsequently ignored
- Linter rightly complains that instantiating a class should not return a different class

### DIFF
--- a/activity/activity_UpdateRepository.py
+++ b/activity/activity_UpdateRepository.py
@@ -4,7 +4,7 @@ from boto.s3.connection import S3Connection
 import tempfile
 from github import Github
 from github import GithubException
-from provider import lax_provider
+import provider
 from provider.storage_provider import StorageContext
 
 """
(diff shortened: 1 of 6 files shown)


Jenniferstrej:
👍

gnott:
Is there a test for UpdateRepository that passes? Or at least to know whether the storage provider is working as expected? I'm curious with the large try except block whether it all gets covered by the test case.

giorgiosironi:
The test only covers the new case (temporary failure) but I'll expand it with a couple of happy paths then. Expect a new `github` provider module then...

giorgiosironi:
We have been plagued by Github failures during end2end runs, so I'll update this to get it merged.

giorgiosironi:
Implementation changed because the storage provider has been introduced separately. Closing and opening a new PR.

PR #400, opened 2017-03-02
Verify Image Server activity and some test

Jenniferstrej:
+ registering activity

@giorgiosironi - the settings variables added can be viewed on settings-example.py
@gnott 

suggestions for important/useful tests that I may have missed are welcome.

### DIFF
--- a/activity/activity_VerifyImageServer.py
+++ b/activity/activity_VerifyImageServer.py
@@ -3,6 +3,7 @@ import json
 from provider.execution_context import Session
 from provider.storage_provider import StorageContext
 import provider.article_structure as article_structure
+import provider.iiif as iiif
 import requests
 
 """
@@ -54,8 +55,7 @@ class activity_VerifyImageServer(activity.activity):
 
             iiif_path_for_article = self.settings.iiif_resolver.replace('{article_id}', article_id)
 
-            results = list(map(lambda fig: self.test_iiif_endpoint(self.iiif_endpoint(self.settings, iiif_path_for_article, fig)),
-                                                              original_figures))
+            results = self.retrieve_endpoints_check(original_figures, iiif_path_for_article)
 
             bad_images = list(filter(lambda x: x[0] == False, results))
 
@@ -77,19 +77,6 @@ class activity_VerifyImageServer(activity.activity):
                                     article_id + '; message: ' + str(e))
             return activity.activity.ACTIVITY_PERMANENT_FAILURE
 
-
-    def iiif_endpoint(self, settings, iiif_path_for_article, figure):
-        iiif_path_for_figure = iiif_path_for_article.replace('{article_fig}', figure)
-        return settings.path_to_iiif_server + iiif_path_for_figure
-
-    def test_iiif_endpoint(self, endpoint):
-        try:
-            response = requests.head(endpoint)
-            if response.status_code != 200:
-                return False, endpoint
-            return True, endpoint
-        except Exception as e:
-            self.logger.exception(str(e))
-            return False, endpoint
-
-
+    def retrieve_endpoints_check(self, original_figures, iiif_path_for_article):
+        return list(map(lambda fig: iiif.try_endpoint(iiif.endpoint(self.settings, iiif_path_for_article, fig), self.logger),
+                        original_figures))
--- a/activity/activity_VerifyImageServer.py
+++ b/activity/activity_VerifyImageServer.py
@@ -1,0 +1,95 @@
+import activity
+import json
+from provider.execution_context import Session
+from provider.storage_provider import StorageContext
+import provider.article_structure as article_structure
+import requests
+
+"""
+activity_VerifyImageServer.py activity
+"""
+
+
+class ValidationException(RuntimeError):
+    pass
+
+
+class activity_VerifyImageServer(activity.activity):
+
+    def __init__(self, settings, logger, conn=None, token=None, activity_task=None):
+        activity.activity.__init__(self, settings, logger, conn, token, activity_task)
+
+        self.name = "VerifyImageServer"
+        self.pretty_name = "Verify Image Server"
+        self.version = "1"
+        self.default_task_heartbeat_timeout = 30
+        self.default_task_schedule_to_close_timeout = 60 * 5
+        self.default_task_schedule_to_start_timeout = 30
+        self.default_task_start_to_close_timeout = 60 * 5
+        self.description = "Checks if original figure is valid when called via the IIIF standard using an Image server"
+        self.logger = logger
+
+    def do_activity(self, data=None):
+
+        try:
+            if self.logger:
+                self.logger.info('data: %s' % json.dumps(data, sort_keys=True, indent=4))
+
+            run = data['run']
+            session = Session(self.settings)
+            article_id = session.get_value(run, 'article_id')
+            version = session.get_value(run, 'version')
+
+        except Exception as e:
+            self.logger.exception(str(e))
+            return activity.activity.ACTIVITY_PERMANENT_FAILURE
+
+        try:
+            storage_context = StorageContext(self.settings)
+            bucket = self.settings.publishing_buckets_prefix + self.settings.ppp_cdn_bucket
+            images_resource = "".join((self.settings.storage_provider, "://", bucket, "/", article_id))
+
+            files_in_bucket = storage_context.list_resources(images_resource)
+            original_figures = article_structure.get_original_figures(files_in_bucket)
+
+            iiif_path_for_article = self.settings.iiif_resolver.replace('{article_id}', article_id)
+
+            results = list(map(lambda fig: self.test_iiif_endpoint(self.iiif_endpoint(self.settings, iiif_path_for_article, fig)),
+                                                              original_figures))
+
+            bad_images = list(filter(lambda x: x[0] == False, results))
+
+            if len(bad_images) > 0:
+                # print endpoints that did not work
+                self.emit_monitor_event(self.settings, article_id, version, run, self.pretty_name, "error",
+                                        "Some images are not available through the IIIF endpoint: " + str(bad_images))
+
+                return activity.activity.ACTIVITY_PERMANENT_FAILURE
+
+            self.emit_monitor_event(self.settings, article_id, version, run, self.pretty_name, "end",
+                                    "Finished Verification. All endpoints work. Article: " + article_id)
+            return activity.activity.ACTIVITY_SUCCESS
+
+        except Exception as e:
+            self.logger.exception(str(e))
+            self.emit_monitor_event(self.settings, article_id, version, run, self.pretty_name, "error",
+                                    "An error occurred when checking IIIF endpoint. Article " +
+                                    article_id + '; message: ' + str(e))
+            return activity.activity.ACTIVITY_PERMANENT_FAILURE
+
+
+    def iiif_endpoint(self, settings, iiif_path_for_article, figure):
+        iiif_path_for_figure = iiif_path_for_article.replace('{article_fig}', figure)
+        return settings.path_to_iiif_server + iiif_path_for_figure
+
+    def test_iiif_endpoint(self, endpoint):
+        try:
+            response = requests.head(endpoint)
+            if response.status_code != 200:
+                return False, endpoint
+            return True, endpoint
+        except Exception as e:
+            self.logger.exception(str(e))
+            return False, endpoint
+
+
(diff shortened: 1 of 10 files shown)


gnott:
Looks clear to me. A couple other file types or ideas, I'm not sure if it is time to add now or later

- Inline graphics, e.g. elife-00666-inf001-v1.jpeg, I think we might want to check they are in the image server (yes, no?)
- For videos we might have their static thumbnail frames in the image server, although we're not copying those over yet

activity/activity_VerifyImageServer.py
```diff
@@ -0,0 +1,95 @@
...
+            session = Session(self.settings)
+            article_id = session.get_value(run, 'article_id')
+            version = session.get_value(run, 'version')
+
+        except Exception as e:
+            self.logger.exception(str(e))
+            return activity.activity.ACTIVITY_PERMANENT_FAILURE
+
+        try:
+            storage_context = StorageContext(self.settings)
```
giorgiosironi: I run into the problem of always having to pass in `settings` everywhere, even when we could conceptually depend upon the provider (in this case the storage one) without the activity code being aware of the settings:
```
activity --depends-on--> provider.storage --depends-on--> settings
```
rather than
```
activity --depends-on--> provider.storage 
        \--depends-on--> settings
```
There is an environment variable `ENVIRONMENT_NAME` that should be available everywhere and that tells us which environment we should use to choose the settings. We can create a small module `environment_settings` that loads the settings for the environment described by this variable, defaulting to `dev`. Then `provider.storage` can do:
```
from environment_settings import settings
# convention is uppercase for global "constant" variables, right? correct me on this
STORAGE_CONTEXT = StorageContext(settings)
```
and this activity can just say
```
from provider.storage import STORAGE_CONTEXT
```
In the tests, we mock
```
@patch('activity.activity_VerifyImageServer.STORAGE_CONTEXT')
```

provider/storage_provider.py
```diff
@@ -67,6 +67,18 @@ def set_resource_from_string(self, resource, data):
         key.key = s3_key
         key.set_contents_from_string(data)
 
+    def list_resources(self, folder):
```
giorgiosironi: :+1: I like the providers acquiring higher-level abstractions

settings-example.py
```diff
@@ -194,6 +194,10 @@ class exp():
     pdf_cover_generator = "http://localhost:8082/personalcover/generate/"
     pdf_cover_landing_page = "http://localhost:8082/personalcover/landing/"
 
+    # IIIF
+    path_to_iiif_server = "https://pathto--iiif.elifesciences.org/"
+    iiif_resolver = "{article_id}/{article_fig}/full/full/0/default.jpg"
```
giorgiosironi: settings added to builder-private :+1: we don't have a iiif server for `continuumtest` so I'll create one

tests/activity/test_activity_verify_image_server.py
```diff
@@ -0,0 +1,77 @@
...
+        fake_storage_context.return_value = FakeStorageContext()
+        self.verifyimageserver.emit_monitor_event = MagicMock()
+        self.verifyimageserver.logger = MagicMock()
+        # When
+        result = self.verifyimageserver.do_activity(data)
+        # Then
+        self.assertEqual(result, self.verifyimageserver.ACTIVITY_PERMANENT_FAILURE)
+
+
+if __name__ == '__main__':
```
giorgiosironi: this looks old school, is it still needed? I usually just run `python -m pytest tests/activity/test_activity_update_repository.py`

tests/provider/test_article_structure.py
```diff
@@ -58,6 +59,65 @@ def test_get_file_type_from_zip_filename(self, input, expected):
...
+        {'input': 'elife-00666-fig3-figsupp1-v1.tif', 'expected': True},
+        {'input': 'elife-00666-fig3-video1.mp4', 'expected': False},
+        {'input': 'elife-00666-fig4-v1.tif', 'expected': True},
+        {'input': 'elife-00666-fig4-code1-v1.xlsx', 'expected': False},
+        {'input': 'elife-00666-figures-v1.pdf', 'expected': False},
+        {'input': 'elife-00666-inf001-v1.jpeg', 'expected': False},
+        {'input': 'elife-00666-repstand1-v1.pdf', 'expected': False},
+        {'input': 'elife-00666-resp-fig1-v1.png', 'expected': True},
+        {'input': 'elife-00666-resp-video1.mp4', 'expected': False},
+        {'input': 'elife-00666-supp1-v1.csv', 'expected': False},
```
giorgiosironi: thorough tests, please also add `elife-00666-supp1-v1.tif` which should be `False` (it's supplementary material not a figure)

activity/activity_VerifyImageServer.py
Jenniferstrej: If I understand what you mean that would require a change in the activity class, which all other activities inherit from. Unfortunately the activity depends heavily on the settings, for Session, for itself to get the buckets... I see your solution working for not needing to pass the settings in some specific cases and we could simplify it but I don't see us removing the dependency on the settings on the activity itself still.

giorgiosironi: Yes, not the intention: activities use the settings and will continue to use it since they need them for many other configurations; but they would be able to require the storage provider without having to specify the settings again (which makes for a simpler interface with that module)

Jenniferstrej: The variables you're talking about are already available here https://github.com/elifesciences/elife-bot/blob/develop/queue_workflow_starter.py#L35-L36, they can be loaded for example https://github.com/elifesciences/elife-bot/blob/develop/starter/starter_PackagePOA.py#L128 and https://github.com/elifesciences/elife-bot/blob/develop/queue_workflow_starter.py#L44. 

I think the one of the reasons I chose to inject the settings was so I could have a settings_mock.py rather than another env in the same file and to have less external dependencies on each provider I change. 
Perhaps I need to change my approach and have the settings for the tests to be loaded from the same file. Trying to think of the pros and cons of each. I normally like to have the providers with least dependencies as possible and leave most of the dependencies on the activity, since it's needs to load a lot of settings vars anyway. But I am open for hearing your cons of passing settings everywhere and the benefit of having potentially all providers in future loading the settings themselves.

giorgiosironi: Yeah I don't think the activities will ever get rid of the settings in general, because they need them for other reasons.

The tests are not really affected by this decision since they usually do:
```
 @patch('activity.activity_ExpandArticle.StorageContext')
...
mock_storage_context.return_value = FakeStorageContext()
```
so the whole provider is mocked and doesn't care about settings being passed in or not.

Having the settings available in providers would be helpful in creating module-wide objects like in here:
https://github.com/elifesciences/elife-bot/pull/393/files#diff-a79d1b704065d473d42cf9accf6221c8R11
Now they cannot have a dependency on settings (in this case `settings.setLevel` for loggers), because when the module is loaded settings are not available yet.

That doesn't mean hardcoding references to settings however, one pattern I have seen working well elsewhere is:
```
# provider/myprovider.py
class MyProvider:
    def __init__(setting1, setting2):
        ...

MY_PROVIDER = MyProvider(settings.settings1, settings.setting2)
```
In that case:
- the class `MyProvider` can be tested without even a reference to settings
- the `import provider.myprovider` is what is mocked in activities tests to use a FakeMyProvider

Jenniferstrej: I will experiment that with storage_provider, but it will require me to change all the tests across all activities. It seems better to merge this one, then make a PR with that change only. Is it ok if merge this one?

provider/iiif.py
```diff
@@ -0,0 +1,15 @@
+import requests
```
giorgiosironi: nice to see this provider taking shape

giorgiosironi:
Feedback from end2end follows.

1) https://end2end--ppp-dash.elifesciences.org/article/6148691793723703318/1/a9f9e5ad-cf06-4040-ac2d-3ab210082f44 
is checking .gif files, it should skip them as they do not go through IIIF:
```
    Some images are not available through the IIIF endpoint: [(False, u'https://end2end--iiif.elifesciences.org/lax:6148691793723703318/elife-6148691793723703318-fig10-v1.gif/full/full/0/default.jpg'), (False, u'https://end2end--iiif.elifesciences.org/lax:6148691793723703318/elife-6148691793723703318-fig11-v1.gif/full/full/0/default.jpg'), (False, u'https://end2end--iiif.elifesciences.org/lax:6148691793723703318/elife-6148691793723703318-fig12-v1.gif/full/full/0/default.jpg'), (False, u'https://end2end--iiif.elifesciences.org/lax:6148691793723703318/elife-6148691793723703318-fig13-v1.gif/full/full/0/default.jpg'), (False, u'https://end2end--iiif.elifesciences.org/lax:6148691793723703318/elife-6148691793723703318-fig15-v1.gif/full/full/0/default.jpg'), (False, u'https://end2end--iiif.elifesciences.org/lax:6148691793723703318/elife-6148691793723703318-fig2-v1.gif/full/full/0/default.jpg'), (False, u'https://end2end--iiif.elifesciences.org/lax:6148691793723703318/elife-6148691793723703318-fig3-v1.gif/full/full/0/default.jpg'), (False, u'https://end2end--iiif.elifesciences.org/lax:6148691793723703318/elife-6148691793723703318-fig4-v1.gif/full/full/0/default.jpg'), (False, u'https://end2end--iiif.elifesciences.org/lax:6148691793723703318/elife-6148691793723703318-fig5-v1.gif/full/full/0/default.jpg'), (False, u'https://end2end--iiif.elifesciences.org/lax:6148691793723703318/elife-6148691793723703318-fig6-v1.gif/full/full/0/default.jpg'), (False, u'https://end2end--iiif.elifesciences.org/lax:6148691793723703318/elife-6148691793723703318-fig7-v1.gif/full/full/0/default.jpg'), (False, u'https://end2end--iiif.elifesciences.org/lax:6148691793723703318/elife-6148691793723703318-fig9-v1.gif/full/full/0/default.jpg')]
```

2) https://end2end--ppp-dash.elifesciences.org/article/9204580859652100230/1/49e4bcea-5c21-4873-a033-7363b95ac301
is checking .xls files, they should also be skipped:
```
Some images are not available through the IIIF endpoint: [(False, u'https://end2end--iiif.elifesciences.org/lax:9204580859652100230/elife-9204580859652100230-fig1-data1-v1.xls/full/full/0/default.jpg'), (False, u'https://end2end--iiif.elifesciences.org/lax:9204580859652100230/elife-9204580859652100230-fig2-data1-v1.xls/full/full/0/default.jpg'), (False, u'https://end2end--iiif.elifesciences.org/lax:9204580859652100230/elife-9204580859652100230-fig3-data1-v1.xls/full/full/0/default.jpg')]
```

3) https://end2end--ppp-dash.elifesciences.org/article/8736294942595715600/1/15fcbca8-37e5-4d40-b282-5a240802ad88
has some bad TIFFs, but it should fallback to the JPG (nothing to do on the bot side, it's a Nginx problem)
```
    Some images are not available through the IIIF endpoint: [(False, u'https://end2end--iiif.elifesciences.org/lax:8736294942595715600/elife-8736294942595715600-fig1-v1.tif/full/full/0/default.jpg'), (False, u'https://end2end--iiif.elifesciences.org/lax:8736294942595715600/elife-8736294942595715600-fig2-v1.tif/full/full/0/default.jpg'), (False, u'https://end2end--iiif.elifesciences.org/lax:8736294942595715600/elife-8736294942595715600-fig3-v1.tif/full/full/0/default.jpg'), (False, u'https://end2end--iiif.elifesciences.org/lax:8736294942595715600/elife-8736294942595715600-fig4-v1.tif/full/full/0/default.jpg'), (False, u'https://end2end--iiif.elifesciences.org/lax:8736294942595715600/elife-8736294942595715600-fig5-v1.tif/full/full/0/default.jpg'), (False, u'https://end2end--iiif.elifesciences.org/lax:8736294942595715600/elife-8736294942595715600-fig6-v1.tif/full/full/0/default.jpg'), (False, u'https://end2end--iiif.elifesciences.org/lax:8736294942595715600/elife-8736294942595715600-fig7-v1.tif/full/full/0/default.jpg')]
```

I'll fix 3, please fix 1 and 2.

giorgiosironi:
3 is now fixed, problem with the configuration of the fallback:

Jenniferstrej:
@giorgiosironi - So should the rule be only tiffs and jpegs?

giorgiosironi:
The best rule I can come up with is the intersection of:
1. whatever is classified as `Figure` by https://github.com/elifesciences/elife-bot/blob/develop/provider/article_structure.py#L5
2. *and* has a .tif or .jpg extension

Note that when we add Glencoe video preview images, those will have to be checked here too (even if they are not figures, they will go through IIIF IIRC)

PR #402, opened 2017-03-07
adds failure cases to article_structure tests and add condition for .…

Jenniferstrej:
…tif and .jpg extensions

### DIFF
--- a/tests/provider/test_article_structure.py
+++ b/tests/provider/test_article_structure.py
@@ -115,7 +115,9 @@ class TestArticleStructure(unittest.TestCase):
                  'elife-00666-inf001-v1.jpg',
                  'elife-00666-inf001-v1-80w.jpg',
                  'elife-00666-table3-data1-v1.xlsx',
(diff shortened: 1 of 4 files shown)


tests/provider/test_article_structure.py
```diff
@@ -113,7 +113,9 @@ def test_get_original_figures(self):
                  'elife-00666-inf001-v1.jpg',
                  'elife-00666-inf001-v1-80w.jpg',
                  'elife-00666-table3-data1-v1.xlsx',
-                 'elife-07702-vor-r4.zip']
+                 'elife-07702-vor-r4.zip',
+                 'elife-6148691793723703318-fig10-v1.gif',
```
giorgiosironi: `get_original_figures` has become an inappropriate method name maybe, since `...-fig10-v1.gif` still _is_ a figure, but it does not need to be checked here. `get_iiif_figures` would be crude but more truthful. `get_original_figures_for_iiif`?

PR #403, opened 2017-03-07
Fail if image resizing fails

giorgiosironi:
The activity catches the exception already, and exits by returning False
(not sure what happens after that, is it retried automatically? It should
not)

So today I fed in by error images which were just small text files (long
story). Everything was green on the dashboard, my goal is to fail at the
ImageResize activity; it already logs errors about the images which are not
working.

### DIFF
--- a/activity/activity_ResizeImages.py
+++ b/activity/activity_ResizeImages.py
@@ -82,8 +82,8 @@ class activity_ResizeImages(activity.activity):
                                     "Resize Images", "error",
                                     "Error resizing images for article" + article_id +
                                     " message:" + e.message)
-            return False
-        return True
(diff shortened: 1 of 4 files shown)


Jenniferstrej:
return False is the same as activity.activity.ACTIVITY_TEMPORARY_FAILURE, so yes, it retries. In the past we didn't have other ways of failing an activity, until @jhroot implemented activity failure codes that would exit the workflow on failure. Most activities have been updated with the correct return code, but not all of them, so when we find something that should not retry and there is "return False" that should be changed to "return activity.activity.ACTIVITY_PERMANENT_FAILURE"

giorgiosironi:
Ok, I'll update

PR #404, opened 2017-03-07
adding log for error response in iiif provider

Jenniferstrej:


provider/iiif.py
```diff
@@ -8,6 +8,7 @@ def try_endpoint(endpoint, logger):
     try:
         response = requests.head(endpoint)
         if response.status_code != 200:
+            logger.info("Error status code != 200. Status code: " + str(response.status_code) + " for URL: " + endpoint)
```
giorgiosironi: I think this can be raised to the `error` level, since it leads to a permanent failure for this run

giorgiosironi: would also use the standard format `logger.info("%s...%s", parameter1, parameter2)`

giorgiosironi: btw, this tries all the figures before giving an error, which I like as it gives more information back

Jenniferstrej: Aggreed.

PR #406, opened 2017-03-13
Elpp 2373

Jenniferstrej:


provider/glencoe_check.py
```diff
@@ -60,6 +62,30 @@ def metadata(msid, settings):
...
+    if '<media content-type="glencoe' in xml_str:
+        return True
+    return False
+
+
+def pad_msid(msid):
+    return str(int(msid)).zfill(5)
+
+
+def check_msid(msid):
```
giorgiosironi: nice to see that this module is attracting Glencoe-related functionality

activity/activity_CopyGlencoeStillImages.py
```diff
@@ -0,0 +1,77 @@
...
+                for jpg in jpgs:
+                    self.store_file(jpg, article_id)
+
+            self.emit_monitor_event(self.settings, article_id, version, run, self.pretty_name, "end",
+                                    "Finished Copying Glencoe still images to CDN. "
+                                    "Article: " + article_id)
+
+            return activity.activity.ACTIVITY_SUCCESS
+        except Exception as e:
+            self.logger.exception(str(e))
```
giorgiosironi: The pattern of the other bot activities suggests that this should be:
``
self.logger.exception("Exception when copying Glencoe still images")
```
as `logger.exception` should already capture message and stack trace of the exception

File: tests/provider/test_glencoe_check.py [hunk 3]
```diff
@@ -0,0 +1,23 @@
+import unittest
```
giorgiosironi: nice that unit tests also reflect the growing `glencoe_check` module

provider/glencoe_check.py
```diff
@@ -60,6 +62,30 @@ def metadata(msid, settings):
     
     return resp.json()
 
+
+def jpg_href_values(metadata):
```
giorgiosironi: are we doing any renaming of the files to a standard format (and is it even possible)? I don't know what is the final word on how the files should be named and whether the name that comes from the Glencoe API is ok

register.py
```diff
@@ -66,6 +66,7 @@ def start(ENV="dev"):
         print 'got response: \n%s' % json.dumps(response, sort_keys=True, indent=4)
 
     activity_names = []
+    activity_names.append("CopyGlencoeStillImages")
     activity_names.append("VerifyImageServer")
```
giorgiosironi: one thing that is missing is adding this file to the VerifyImageServer checks, if it's not already being picked up. This however can be done in another PR later

provider/glencoe_check.py
Jenniferstrej: We are not doing any remaining. The last 4 comments in the conversation in the Jira ticket make me understand that having the file name the same as the mp4 one but with .jpg extension is fine and Exeter is treating it as a pattern (see James comment) which makes me think I should just take the file name that Exeter provider in the glencoe metadata.

register.py
Jenniferstrej: Right, should it be part of the DoD of this task or a new task?

activity/activity_CopyGlencoeStillImages.py
Jenniferstrej: I didn't realise. I double checked the documentation and pushed changes.

register.py
giorgiosironi: (as agreed in standup, new PR but same task)

provider/glencoe_check.py
giorgiosironi: as in standup, we should fail fast if the filenames provided by Glencoe are not what we expect. This check is not blocking for this PR, but should be in the DOD

gnott:
Sorry for the late review - is there a test for test_jpg_href_values() glencoe provider, where the metadata is either None or blank or malformed? That complicated looking sequence and lambda function looks like it could use some tests to be sure it is performing as expected.

PR #408, opened 2017-03-15
Separating article id from image count

giorgiosironi:
On the dashboard, we see
```
Finished converting images for  426183516370560497025 images processed
```
but it should be
```
Finished converting images for 4261835163705604970: 25 images processed
```


Jenniferstrej:
👍

PR #410, opened 2017-03-15
Retry ApprovePublication

giorgiosironi:
Sometimes the old website goes 500 due to MySQL deadlocks. There's nothing we can do for its terminal illness, but we can retry to avoid giving up the run

### DIFF
--- a/activity/activity_ApprovePublication.py
+++ b/activity/activity_ApprovePublication.py
@@ -60,7 +60,10 @@ class activity_ApprovePublication(activity.activity):
                 auth = requests.auth.HTTPBasicAuth(self.settings.drupal_update_user,
                                                    self.settings.drupal_update_pass)
             r = requests.put(destination, data="{ \"publish\": \"1\" }", headers=headers, auth=auth)
-            self.logger.info("PUT response was %s" % r.status_code)
+            self.logger.info("PUT response was %s, retrying" % r.status_code)
+
+            if r.status_code == 500:
+                return activity.activity.ACTIVITY_TEMPORARY_FAILURE
 
             if r.status_code == 200:
                 self.set_monitor_property(self.settings, article_id, 'publication-status',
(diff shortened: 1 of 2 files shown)


Jenniferstrej:
Mind that if there is an actual problem and the old website keeps returning 500 we will only find out after it times out and the logs are full of tries, otherwise it's ok 👍

giorgiosironi:
If only it returned a better error, we could check for 503 or similar. Should we adjust the timeout for the activity downward, to just a few minutes?

Jenniferstrej:
Yes, I normally keep those three the same 
"heartbeat_timeout",
"schedule_to_close_timeout",
"start_to_close_timeout"
(I would change the three of them to the same time, just a few minutes as you said, since a heartbeat should be less or equal the start to close timeout, and start to close timeout should be less or equal to schedule to close timeout if you see what I mean. I will double check their exact meaning on Amazon and post here for clarity.) 

The change should be in this file:
https://github.com/elifesciences/elife-bot/blob/develop/workflow/workflow_ApproveArticlePublication.py#L68

Jenniferstrej:
See "Timeouts in Activity Tasks" for reference:
http://docs.aws.amazon.com/amazonswf/latest/awsflowguide/swf-timeout-types.html

gnott:
I think (going from memory) because we haven't implemented heartbeats, the "heartbeat_timeout" and "start_to_close_timeout" values should be the same, possibly "schedule_to_close_timeout" also the same - it's probably fine for all three to be the same at this stage.

giorgiosironi:
Just discovered from this test run:
https://end2end--ppp-dash.elifesciences.org/article/39579087945004177/1/9ef59af5-0d55-4e85-98b7-0df306be9101
that this activity is already being retried. The SWF console said `error: activity failed with result None` for the first failure, the activity returned `False` according to `worker.log`.

However, it reports an error to the dashboard for every failed attempt, which we don't need. This change will improve this and the end2end tests (which check no errors appear on the dashboard) will be happier. Adding the timeouts now.

giorgiosironi:
The timeout is at 5 minutes, which is the value I intended to set :thinking: so, merging, but if you want a stricter value feel free to change it later.

PR #411, opened 2017-03-16
Elpp 2373 validate jpgs x videos

Jenniferstrej:
These changes add CopyGlencoeStillImages activity to ProcessArticleZip workflow. It will run after DepositAssets and:
- Lookup for jpg_href key in glencoe metadata;
- save jpgs from glencoe metadata;
- list the jpgs and compare them with files that have the same name as them
        - if there isn't any match we will stop the workflow execution and display an error on the dashboard containing a list of the unmatched file names

I included tests for the activity and added a simpler version of a test for jpg_hrefs lookup.

The changes above should satisfty the checklist items:
"Copy Glencoe Still Images" activity + tests
Confirm if all jpgs have a video with the same name in CDN Bucket

What's next:
Add the video check under Verify Image Server
Create a workflow that will have only CopyGlencoeStillImages in order to run it standalone

@giorgiosironi - will you need to add anything further in the end2end tests to test this?


provider/glencoe_check.py
```diff
@@ -66,8 +66,8 @@ def metadata(msid, settings):
 def jpg_href_values(metadata):
 
     return list((seq(metadata.items())
-                .filter(lambda x: x[0].startswith('media') and 'jpg_href' in x[1])
-                .map(lambda y: y[1]['jpg_href'])))
+                .filter(lambda key_value: key_value[0].startswith('media') and 'jpg_href' in key_value[1])
```
gnott: startswith('media') might be limiting here, if you look at the new kitchen sink example http://movie-usa.glencoesoftware.com/metadata/10.7554/eLife.00666  key names include "fig3video1", "fig3video1", "app2video1"

giorgiosironi:
Comment on end2end: we use clones of 00230 and 00666 (the latter not being a real article) to test videos. With the usual id trick, a generated id like 12345600230 will retrieve the videos and the still images of 00230. There may be some sublety involved as these files should be copied on our side with the original id for consistency - they may be elife-00230-... on the other side but should be elife-12345600230 on our side.

activity/activity_CopyGlencoeStillImages.py
```diff
@@ -72,6 +85,36 @@ def s3_resource(self, path, article_id):
...
+                              self.settings.publishing_buckets_prefix + self.settings.ppp_cdn_bucket + "/" + \
+                              article_id
+        return storage_context.list_resources(article_path_in_cdn)
+
+    def validate_jpgs_against_cdn(self, files_in_cdn, jpgs):
+        jpgs_rep_no_extension = map(lambda filename: os.path.splitext(filename)[0], jpgs)
+        files_in_cdn_no_extention = map(lambda filename: os.path.splitext(filename)[0], files_in_cdn)
+        jpgs_without_video = []
+        for file_no_ext in jpgs_rep_no_extension:
+            if len(list(filter(lambda filename: filename == file_no_ext, files_in_cdn_no_extention))) != 2:
```
giorgiosironi: to run in end2end this check will have to work for elife-12345600230 on CDN vs elife-00230 on Glencoe

activity/activity_CopyGlencoeStillImages.py
```diff
@@ -72,6 +85,36 @@ def s3_resource(self, path, article_id):
 
     def store_file(self, path, article_id):
         storage_context = StorageContext(self.settings)
-        with open(path) as still_image:
-            storage_context.set_resource_from_file(self.s3_resource(path, article_id), still_image)
+        r = requests.get(path)
+        if r.status_code == 200:
+            storage_context.set_resource_from_string(self.s3_resource(path, article_id), r.content,
```
giorgiosironi: this would need a bit of renaming on the id contained in `path`: `elife-00230-...` => `elife-12345600230`

giorgiosironi: Here we have some flexibility to accept in input of what we should accept (badly named or mismatched files on Glencoe, Graham said) but strictness in output (a predictable URL for the bot-lax-adaptor to link to the CDN). So in general I would try to build the S3 filename ourselves, using as little as possible from the original name (just the suffix right before the extension?)

gnott: I think we're expecting the .jpg will have the same name in Glencoe as the .avi (or other extension) video file, because it's a 1 to 1 match. This activity so far is not looking the XML and comparing it, but that verification should be happening earlier when the Verify Glencoe happens prior to Lax ingest. Jen and I discussed these issues on our call and sorry I haven't added these notes elsewhere. The videos and the video workflow is deceptively complex.

activity/activity_CopyGlencoeStillImages.py
Jenniferstrej: I think this satisfies the requirements now https://github.com/elifesciences/elife-bot/pull/411/commits/5b9765c6a45436aa0b3aac9506b53bb5cda5ee50

provider/glencoe_check.py
Jenniferstrej: I'm not validating the name of the first level key in glencoe metadata anymore, so then if it's "video.." or anything else and there is a jpg_href, it will be included.
https://github.com/elifesciences/elife-bot/pull/411/commits/bc3832bfea54f4d4354b7d5185eacb34423aae06

PR #413, opened 2017-03-17
Elpp 2373 logs

Jenniferstrej:


Jenniferstrej:
I need to fix tests. one sec

Jenniferstrej:
@giorgiosironi please view test_glencoe_check.py changes (to see if I am doing the right thing)

activity/activity_CopyGlencoeStillImages.py
```diff
@@ -95,8 +91,17 @@ def do_activity(self, data=None):
...
+    def store_jpgs(self, jpgs, article_id):
+        jpg_filenames = []
+        for jpg in jpgs:
+                    jpg_filename = self.store_file(jpg, article_id)
+                    jpg_filenames.append(jpg_filename)
+        return jpg_filenames
+
     def s3_resource(self, path, article_id):
         filename = os.path.split(path)[1]
+        filename = glencoe_check.pad_article_for_end2end(filename, article_id)
```
giorgiosironi: would probably rename this method afterward but no problem

activity/activity_CopyGlencoeStillImages.py
```diff
@@ -105,8 +110,12 @@ def store_file(self, path, article_id):
         storage_context = StorageContext(self.settings)
         r = requests.get(path)
         if r.status_code == 200:
-            storage_context.set_resource_from_string(self.s3_resource(path, article_id), r.content,
+            resource = self.s3_resource(path, article_id)
+            self.logger.info("S3 resource: " + resource)
```
giorgiosironi: should use `,` for logging but not a blocker

activity/activity_CopyGlencoeStillImages.py
```diff
@@ -105,8 +110,12 @@ def store_file(self, path, article_id):
...
+            jpg_filename = os.path.split(path)[1]
+            storage_context.set_resource_from_string(resource, r.content,
                                                      content_type=r.headers['content-type'])
+            return jpg_filename
```
giorgiosironi: down the road, it may be easier to return `resource` for comparison as it will avoid further translation of article ids in the comparison between video files and still image files

provider/glencoe_check.py
```diff
@@ -85,11 +85,11 @@ def check_msid(msid):
         return pad_msid(msid[-5:])
     return pad_msid(msid)
 
-def pad_article_for_end2end(filename, article_id):
+def extend_article_for_end2end(filename, article_id):
```
giorgiosironi: this method is not being called, apart from the tests

activity/activity_CopyGlencoeStillImages.py
Jenniferstrej: yes, I renamed it but forgot to change it here. I am trying now but my editor and my computer are behaving strangely

activity/activity_CopyGlencoeStillImages.py
Jenniferstrej: I tried using , but log info thinks it's there are 2 arguments, and not one string only.

provider/glencoe_check.py
Jenniferstrej: changed now.

activity/activity_CopyGlencoeStillImages.py
```diff
@@ -117,14 +126,19 @@ def list_files_from_cdn(self, article_id):
         return storage_context.list_resources(article_path_in_cdn)
 
     def validate_jpgs_against_cdn(self, files_in_cdn, jpgs, article_id):
-        jpgs_rep_no_extension = map(lambda filename: os.path.splitext(filename)[0], jpgs)
-        files_in_cdn_no_extention = map(lambda filename: os.path.splitext(filename)[0], files_in_cdn)
-        files_in_cdn_article_padded = map(lambda filename: glencoe_check.pad_article_for_end2end(filename, article_id),
-                                          files_in_cdn_no_extention)
+        jpgs_rep_no_extension = list(map(lambda filename: os.path.splitext(filename)[0], jpgs))
```
giorgiosironi: at this point `jpgs_rep_no_extension` is `["elife-00666-media1.jpg"]` while `files_in_cdn_no_extension` is `["elife-1234500666-media1.jpg"]`. I think we should return the filename of `resource` from `store_file()` so that we only have to compare between video files and image files in the CDN; those files will all have the same article id, which makes us able here to drop the commented out part about padding

giorgiosironi: ah, I misread: we are already doing that.

giorgiosironi: would still renamed `files_in_cdn` and `jpgs` to `videos_in_cdn` and `video_still_jpgs_in_cdn` respectively

PR #414, opened 2017-03-17
Form still image path on CDN by using current article id

giorgiosironi:
No matter if the article id on Glencoe is the original one, Helps in
end2end testing and moves us toward building the filename ourselves as
much as possible.


Jenniferstrej:
Looks ok to me. 👍

PR #416, opened 2017-03-17
check if article is poa

Jenniferstrej:
if it’s poa end activity - no videos are added in poa’s


activity/activity_CopyGlencoeStillImages.py
```diff
@@ -46,6 +46,12 @@ def do_activity(self, data=None):
         self.emit_monitor_event(self.settings, article_id, version, run, self.pretty_name, "start",
                                 "Starting check/copy of Glencoe video still images " + article_id)
         try:
+            file_name = session.get_value(run, 'file_name')
+            if "poa" in file_name:
```
giorgiosironi: uhm this could be made safer, like `-poa-` or something similar

PR #418, opened 2017-03-17
Add still jpgs to IIIF verification

Jenniferstrej:
I plan to refactor is_media_file (change its name)
and apply the method in article_structure on ApplyVersionNumber since they do the same thing

These changes should allow still images to be verified with IIIF api.

This could be merged after we make sure everything works with the previous merges for ELPP-2373

### DIFF
--- a/provider/article_structure.py
+++ b/provider/article_structure.py
@@ -139,10 +139,17 @@ def file_parts(filename):
 
 
 def get_media_file_images(files):
-    return list(filter(lambda f: is_media_file(f) and has_extensions(f, ['jpg']), files))
+    return list(filter(lambda f: is_video_file(f) and has_extensions(f, ['jpg']), files))
 
 
-def is_media_file(filename):
+def is_video_file(filename):
+    """
+    Simple check for video file names
+    E.g. match True on elife-00005-media1.mov
+         match True on elife-99999-resp-media1.avi
+         match False on elife-00005-media1-code1.wrl
+    """
+
     (file_prefix, file_extension) = file_parts(filename)
     file_type_plus_index = file_prefix.split('-')[-1]
     if ("media" in file_type_plus_index) or ("video" in file_type_plus_index):
--- a/provider/article_structure.py
+++ b/provider/article_structure.py
@@ -128,17 +128,24 @@ def get_original_files(files):
 def get_figures_for_iiif(files):
     # should only be tif
     fs = [f for f in get_original_files(files) if (article_figure(f) and has_extensions(f, ['tif']))]
+    fs = originals_and_figures + get_media_files(files)
     return fs
 
+
 def file_parts(filename):
-        prefix = filename.split('.')[0]
-        extension = filename.split('.')[-1]
-        return (prefix, extension)
+    prefix = filename.split('.')[0]
+    extension = filename.split('.')[-1]
+    return prefix, extension
+
+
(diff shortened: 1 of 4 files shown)


provider/article_structure.py
```diff
@@ -125,19 +125,27 @@ def get_original_files(files):
...
+
+def get_media_files(files):
+    return list(filter(is_media_file, files))
+
 
 def is_media_file(filename):
     (file_prefix, file_extension) = file_parts(filename)
     file_type_plus_index = file_prefix.split('-')[-1]
-    if "media" in file_type_plus_index:
+    if ("media" in file_type_plus_index) or ("video" in file_type_plus_index):
```
gnott: A second look at this maybe instead 

if file_type_plus_index in ["media", "video"]:

Jenniferstrej:
Yes, i mentioned that because I intend to refactor. It's just because they
are 2 separate PRs now and once both pass I was going to leave both in the
article_structure file. But feel free to move them around while I'm on
holiday if you can :). If you can't, I'll do it when I'm back.

Jenniferstrej:
Yes, I actually had forgotten​ the best syntax at the time. Feel free to
change it if you can :-).

provider/article_structure.py
```diff
@@ -125,19 +125,27 @@ def get_original_files(files):
...
 def file_parts(filename):
-        prefix = filename.split('.')[0]
-        extension = filename.split('.')[-1]
-        return (prefix, extension)
+    prefix = filename.split('.')[0]
+    extension = filename.split('.')[-1]
+    return prefix, extension
+
+
+def get_media_files(files):
```
giorgiosironi: I think this needs to filter out everything that is not a `.jpg` from the `-media`/`-video` files (e.g. `elife-12345-media1.mpg` should not be considered); unless this is already done before passing in `files`.

It could be named `get_media_images`?

gnott: Is that ``elife-12345-media1.jpg`` should not be considered (and not ``.mpg``)?

If ``.jpg`` files are ignored in this function I think it will still support the additional use which is where it identifies which files to not add a version number in the ApplyVersionNumber activity.

giorgiosironi: This is used here https://github.com/elifesciences/elife-bot/pull/418/files#diff-bd63540024801b25141106fad53bab31R131 to generate the list of *images* to check through the IIIF server so it should not contain `.avi`, `.wmv`, `.mpg` files. Or they should be further filtered in `get_figures_for_iiif` if this function is used elsewhere.

Jenniferstrej: We need to find a way to get all tiff Figures AND all the jpg media/video. I could filter out only jpgs in "get_media_files" and change the name of the function to "get_media_file_images" as you mentioned @giorgiosironi . Then your changes under https://github.com/elifesciences/elife-bot/pull/425 could be merged and the problems would be fixed. I guess your changes can be merged first without affecting much. I just need to solve some conflicts with this one.

Jenniferstrej:
Changes related to ApplyVersionNumber refactoring of has_videos, now added to article_structure and change of function now get_media_file_images that will only get still glencoe images.

Jenniferstrej:
@giorgiosironi and @gnott : Is it ok to merge?

gnott:
These changes look really good to me :+1:

PR #419, opened 2017-03-20
Runs only continuum-tagged end2end tests

giorgiosironi:
Not useful to run tests that involve, for example, only journal-cms and search and not the bot's code.

Takes the occasion to switch to the new, clearer stages syntax.


giorgiosironi:
This should make it slightly simpler to go through end2end.

PR #425, opened 2017-03-28
VerifyImageServer should only check the .tif version of a figure

giorgiosironi:
We are falling back to the .jpg version if the .tif doesn't work into the IIIF server, but this is done transparently by the server itself. So I believe we don't need to also check the .jpg, and we can do half the work that this activity is doing for better reliability and speed.

### DIFF
--- a/lint.sh
+++ b/lint.sh
@@ -3,4 +3,12 @@ set -e
 source venv/bin/activate
 
 # intentionally only the script files in the root folder
-pylint -E *.py provider/storage_provider.py provider/lax_provider.py provider/imageresize.py activity/activity_Update*.py activity/activity_Pubmed*.py
+python -m pylint -E \
+    *.py \
+    activity/activity_Update*.py \
+    activity/activity_Pubmed*.py \
+    provider/article_structure.py \
+    provider/imageresize.py \
+    provider/lax_provider.py \
+    provider/storage_provider.py \
+    tests/provider/*.py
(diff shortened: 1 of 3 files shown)


giorgiosironi:
Added, I was surprised not to see a test fail after the change.

gnott:
:+1:

Jenniferstrej:
Are we not excluding the glencoe still images (they are only jpg) (can't remember. I can check tomorrow)?

gnott:
I think the glencoe stills will be ignored because the file name pattern does not match type "Figure". We could add another file name to the test scenario to be sure.

giorgiosironi:
Not sure whether they were included, but feel free to keep open until that is solved. This was causing some fragility in testing but it is not blocking.

Jenniferstrej:
It looks like that we could exclude the jpgs from that line of code, but we would need to first have the changes added in this PR https://github.com/elifesciences/elife-bot/pull/418/files#diff-bd63540024801b25141106fad53bab31R131. There are a few observations there, I see, but both PRs have something missing then. This one will still exclude the Glencoe images that need to be returned from get_figures_for_iiif function. The PR I linked in this comment would allow for the jpgs to be returned there (but it also has its own corrections to be made under is_media_file function for example)

Jenniferstrej:
@giorgiosironi - if you and @gnott agree with my latest comment under https://github.com/elifesciences/elife-bot/pull/418 I believe you can merge this PR 👍

giorgiosironi:
Agreed, let's get this in and consequently into the #418.

PR #426, opened 2017-03-28
Media video tweaks

gnott:
To support the new kitchen sink XML, the videos are expressed differently in the XML, with tests to support the check.

Also it looks like is_media_file() is no longer used, and is_video_file() is elsewhere and smarter now.

### DIFF
--- a/provider/glencoe_check.py
+++ b/provider/glencoe_check.py
@@ -1,5 +1,6 @@
 import sys, json
 import requests
+import re
 from functional import seq
 
 '''
@@ -71,7 +72,7 @@ def jpg_href_values(metadata):
 
 
 def has_videos(xml_str):
-    if '<media content-type="glencoe' in xml_str:
+    if re.search(ur'<media.*mimetype="video".*>', xml_str):
         return True
     return False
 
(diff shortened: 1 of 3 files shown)


provider/glencoe_check.py
```diff
@@ -71,7 +72,7 @@ def jpg_href_values(metadata):
 
 
 def has_videos(xml_str):
-    if '<media content-type="glencoe' in xml_str:
+    if re.search(ur'<media.*mimetype="video".*>', xml_str):
```
giorgiosironi: regexes for XML/HTML always give me the shivers, but this can be made safer by substituting `.*` with `[^>]*` which won't match `<media><another-tag mimetype="video">` (not likely to happen)

gnott:
Yes I also wanted to move "is_video_file" from the activity to article_structure, I can do that, I like it.

gnott:
I see, maybe you want to do that in https://github.com/elifesciences/elife-bot/pull/418/ and then I can merge and resolve conflicts from there into this PR.

Jenniferstrej:
The PR that moves/changes the is_media_file function has been merged. I guess you can remove that from your changes and only keep the glencoe has videos thing.

provider/glencoe_check.py
gnott: Good suggestion of course, it was just a quick way to pass the example test cases. I'm going to commit a revision to this.

gnott:
Any additional feedback - ok to merge?

PR #428, opened 2017-03-29
Synchronise Silent correction workflow

Jenniferstrej:
The IIIF and Glencoe verifications should also happen during Silent corrections.

### DIFF
--- a/workflow/workflow_SilentCorrectionsIngest.py
+++ b/workflow/workflow_SilentCorrectionsIngest.py
@@ -90,6 +90,17 @@ class workflow_SilentCorrectionsIngest(workflow.workflow):
                         "start_to_close_timeout": 60 * 10
                     },
                     {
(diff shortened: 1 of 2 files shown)


Jenniferstrej:
I wonder if @giorgiosironi would like to add further end2end tests related to this. Having the silent corrections workflow add an invalid image and see VerifyImageServer complaining for example.

giorgiosironi:
We do test the happy path of silent corrections in end2end. However, I always deemed error cases too slow to be tested there. I have the dream of providing enough test infrastructure to elife-bot--ci so that it can run full workflow tests there by faking (instead of mocking) external dependencies (https://github.com/atlassian/localstack/blob/master/README.md has been brought to my attention) but there's no fake implementation of SWF we could use so that would have to be real and prevent tests from running locally...

PR #430, opened 2017-03-31
Elpp 2497

Jenniferstrej:


starter/starter_CopyGlencoeStillImages.py
```diff
@@ -0,0 +1,103 @@
+import os
+import boto.swf.exceptions
+
+# Add parent directory for imports
```
giorgiosironi: argh, are these two lines something all starters do? Where do they come from?

starter/starter_CopyGlencoeStillImages.py
```diff
@@ -0,0 +1,103 @@
...
+            logger.info('got response: \n%s' % json.dumps(response, sort_keys=True, indent=4))
+
+        except NullRequiredDataException as e:
+            logger.exception(e.message)
+            raise
+
+        except boto.swf.exceptions.SWFWorkflowExecutionAlreadyStartedError:
+            # There is already a running workflow with that ID, cannot start another
+            message = 'SWFWorkflowExecutionAlreadyStartedError: There is already a running workflow with ID %s' % workflow_id
+            logger.info(message)
```
giorgiosironi: should probably be `logger.error`, we are requesting something to be done but nothing will be executed because of the already running workflow

starter/starter_CopyGlencoeStillImages.py
Jenniferstrej: Aren't you talking about lines 5 and 6? 
regarding boto.swf.exceptions, it's in order to catch an exception that often happens (especially when debuging). Regarding "import os" and lines 5 and 6 I am not sure, I can take a look to understand why it's there. Maybe @gnott remembers ?

giorgiosironi: Yes, line 5 and 6

gnott: It could be from old dependencies that included code for FluidInfo and/or the elife-poa-xml-generation repo. They may not be required anymore, I haven't checked.

giorgiosironi: I would say comment it out and if required, we will enable it otherwise drop it?

tests/starter/test_starter_copy_glencoe_still_images.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def test_copy_glencoe_still_images_starter_no_article(self):
+        self.assertRaises(NullRequiredDataException, self.stater_copy_glencoe_still_images.start,
+                          settings=settings_mock)
+
+    @patch('starter.starter_helper.get_starter_logger')
+    @patch('boto.swf.layer1.Layer1')
+    def test_copy_glencoe_still_images_starter_(self, fake_boto_conn, fake_logger):
+        fake_boto_conn.return_value = FakeBotoConnection()
+        self.stater_copy_glencoe_still_images.start(settings=settings_mock, article_id='00353')
```
giorgiosironi: small typo `stater`

giorgiosironi: at least an assertion would be useful, to be added later

PR #431, opened 2017-03-31
PMCDeposit activity to not fill up the disk when it fails repeatedly.

gnott:


gnott:
This should fix the instance disk filing up when the activity fails @giorgiosironi

gnott:
It will be as silent as before I think, because it would just time out, and we are not watching for timed out workflows carefully. Except, there would be no email notification sent (to Production and to PMC) on an FTP failure.

PR #436, opened 2017-04-13
Elpp 2495

Jenniferstrej:
Fixes the problem with no files being copied to the CDN bucket by DepositIIIFAssets and adds a couple of tests just to make sure there are no language errors.

### DIFF
--- a/tests/activity/classes_mock.py
+++ b/tests/activity/classes_mock.py
@@ -108,7 +108,7 @@ class FakeStorageContext:
         pass
 
     def list_resources(self, resource):
-        return ["elife-00353-fig1-v1.tif", "elife-00353-v1", "elife-00353-v1.xml"]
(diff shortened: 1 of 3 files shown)


tests/activity/classes_mock.py
```diff
@@ -107,6 +107,12 @@ def set_resource_from_filename(self, resource, file):
     def set_resource_from_string(self, resource, data, content_type=None):
         pass
 
+    def list_resources(self, resource):
+        return ["elife-00353-fig1-v1.tif", "elife-00353-v1", "elife-00353-v1.xml"]
```
giorgiosironi: the second file, without an extension, looks strange?

Jenniferstrej: oops, forgot that. Adding now..

PR #437, opened 2017-04-13
Fix: Videos were not being added to CDN, only images

Jenniferstrej:


giorgiosironi:
I understand the intent, I guess we need videos as CopyGlencoesStillImages depends on it later (even if videos itself are not accessed through IIIF, their video stills will be).

PR #440, opened 2017-04-19
Elpp 2495 - ConvertImagesToJPG extracted from ResizeImages

Jenniferstrej:
with content refactored into a provider.
ConvertImagesToJPG uses a provider and more StorageProvider functions. It's basically a refactored ResizeImages and excludes some formats (originally from formats.yaml)

### DIFF
--- a/activity/activity_ConvertImagesToJPG.py
+++ b/activity/activity_ConvertImagesToJPG.py
@@ -51,8 +51,7 @@ class activity_ConvertImagesToJPG(activity.activity):
             formats = {"Original": {
                             "sources": "tif",
                             "format": "jpg",
-                            "resolution": 96,
-                            "download": "yes"
+                            "resolution": 96
                         }}
 
             for file_name in figures:
@@ -72,7 +71,7 @@ class activity_ConvertImagesToJPG(activity.activity):
 
             self.emit_monitor_event(self.settings, article_id, version, run, self.pretty_name, "end",
                                     "Finished converting images for " + article_id + ": " +
-                                    str(len(files_in_bucket)) + " images processed ")
+                                    str(len(figures)) + " images processed ")
             return activity.activity.ACTIVITY_SUCCESS
 
         except Exception as e:
--- a/activity/activity_ConvertImagesToJPG.py
+++ b/activity/activity_ConvertImagesToJPG.py
@@ -23,7 +23,6 @@ class activity_ConvertImagesToJPG(activity.activity):
         self.default_task_start_to_close_timeout = 60 * 5
         self.description = "Converts article images to JPG"
         self.logger = logger
-        self.formats = self.load_formats()
 
     def do_activity(self, data=None):
         if self.logger:
@@ -49,14 +48,12 @@ class activity_ConvertImagesToJPG(activity.activity):
 
             figures = filter(article_structure.article_figure, files_in_bucket)
 
-            formats = [
-                    { "Original": {
-                        "sources": "tif",
-                        "format": "jpg",
-                        "resolution": 96,
-                        "download": "yes"
-                    }}
-                ]
+            formats = {"Original": {
+                            "sources": "tif",
+                            "format": "jpg",
+                            "resolution": 96,
+                            "download": "yes"
+                        }}
 
             for file_name in figures:
                 figure_resource = orig_resource + "/" + file_name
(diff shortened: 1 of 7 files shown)


activity/activity_ConvertImagesToJPG.py
```diff
@@ -0,0 +1,84 @@
...
+
+            storage_context = StorageContext(self.settings)
+            files_in_bucket = storage_context.list_resources(orig_resource)
+
+            figures = filter(article_structure.article_figure, files_in_bucket)
+
+            formats = {"Original": {
+                            "sources": "tif",
+                            "format": "jpg",
+                            "resolution": 96,
```
giorgiosironi: this `96` looks more like a `dpi` parameter than a `resolution` like `800x600`

activity/activity_ConvertImagesToJPG.py
```diff
@@ -0,0 +1,84 @@
...
+                add_resource_path = storage_provider + published_bucket_path + "/" + article_id + "/"
+
+                publish_locations = [cdn_resource_path, add_resource_path]
+
+                image_conversion.generate_images(self.settings, formats, file_pointer, article_structure.ArticleInfo(file_name),
+                                                 publish_locations, self.logger)
+
+            self.emit_monitor_event(self.settings, article_id, version, run, self.pretty_name, "end",
+                                    "Finished converting images for " + article_id + ": " +
+                                    str(len(files_in_bucket)) + " images processed ")
```
giorgiosironi: only `len(figures)` were actual images to be processed right?

activity/activity_ConvertImagesToJPG.py
```diff
@@ -0,0 +1,84 @@
...
+            storage_context = StorageContext(self.settings)
+            files_in_bucket = storage_context.list_resources(orig_resource)
+
+            figures = filter(article_structure.article_figure, files_in_bucket)
+
+            formats = {"Original": {
+                            "sources": "tif",
+                            "format": "jpg",
+                            "resolution": 96,
+                            "download": "yes"
```
giorgiosironi: I don't think we need the `download` version, in the 2.0 case; the images created here will be read only by the IIIF server

workflow/workflow_IngestArticleZip.py
```diff
@@ -90,6 +90,17 @@ def __init__(self, settings, logger, conn=None, token=None, decision=None,
                         "start_to_close_timeout": 60 * 15
                     },
                     {
+                        "activity_type": "ConvertImagesToJPG",
```
giorgiosironi: :+1: right place

provider/image_conversion.py
```diff
@@ -0,0 +1,42 @@
...
+                format_spec = formats[format_spec_name]
+                # if sources not present or includes file extension for this image
+                if 'sources' not in format_spec or info.extension in [
+                        x.strip() for x in format_spec['sources'].split(',')]:
+                    download = 'download' in format_spec and format_spec['download']
+                    fp.seek(0)  # rewind the tape
+                    filename, image = resizer.resize(format_spec, fp, info, logger)
+                    if filename is not None and image is not None:
+                        store_in_publish_locations(settings, filename, image, publish_locations, download)
+                        logger.info("Stored image %s as %s" % (filename, str(publish_locations)))
```
giorgiosironi: looks like if `filename` or `image` are `None` we have encountered a error; throwing an exception?

activity/activity_ConvertImagesToJPG.py
giorgiosironi: (the provider may support it, but we may not need to actually require it from this activity)

activity/activity_ConvertImagesToJPG.py
Jenniferstrej: True. In ResizeImages I believe there is no such filter. I'll change.

activity/activity_ConvertImagesToJPG.py
Jenniferstrej: Ok, I'll remove it from the formats definition only

activity/activity_ConvertImagesToJPG.py
Jenniferstrej: I wonder where this "resolution" is used...

Jenniferstrej: I'm removing it altogether from the formats dictionary. It doesn't seem to influence anything. Less confusing.

PR #441, opened 2017-04-21
ELPP-2495 - Silent correction in sync with normal workflows

Jenniferstrej:


workflow/workflow_ProcessArticleZip.py
```diff
@@ -112,6 +112,17 @@ def __init__(self, settings, logger, conn=None, token=None, decision=None,
                         "start_to_close_timeout": 60 * 5
                     },
                     {
+                        "activity_type": "CopyGlencoeStillImages",
```
giorgiosironi:
```
$ grep -rl CopyGlencoeStillImages workflow
workflow/workflow_CopyGlencoeStillImages.py
workflow/workflow_IngestArticleZip.py
workflow/workflow_SilentCorrectionsProcess.py
```
in how many places do we need this activity? I though it was only needed once per run?

workflow/workflow_SilentCorrectionsIngest.py
```diff
@@ -101,6 +101,50 @@ def __init__(self, settings, logger, conn=None, token=None, decision=None,
                         "start_to_close_timeout": 60 * 15
                     },
                     {
+                        "activity_type": "ConvertImagesToJPG",
```
giorgiosironi: ConvertImagesToJPG is consistent with `IngestArticleZip` workflow

workflow/workflow_SilentCorrectionsIngest.py
```diff
@@ -101,6 +101,50 @@ def __init__(self, settings, logger, conn=None, token=None, decision=None,
...
+                        "version": "1",
+                        "input": data,
+                        "control": None,
+                        "heartbeat_timeout": 60 * 15,
+                        "schedule_to_close_timeout": 60 * 15,
+                        "schedule_to_start_timeout": 300,
+                        "start_to_close_timeout": 60 * 15
+                    },
+                    {
+                        "activity_type": "DepositIIIFAssets",
```
giorgiosironi: DepositIIIFAssets is consistent with `IngestArticleZip` workflow too

workflow/workflow_SilentCorrectionsIngest.py
```diff
@@ -101,6 +101,50 @@ def __init__(self, settings, logger, conn=None, token=None, decision=None,
...
+                        "version": "1",
+                        "input": data,
+                        "control": None,
+                        "heartbeat_timeout": 60 * 15,
+                        "schedule_to_close_timeout": 60 * 15,
+                        "schedule_to_start_timeout": 300,
+                        "start_to_close_timeout": 60 * 15
+                    },
+                    {
+                        "activity_type": "VerifyImageServer",
```
giorgiosironi: same question as `CopyGlencoeStillImages`, what leads us to do this in both ingest and and process (may be necessary)

workflow/workflow_ProcessArticleZip.py
Jenniferstrej: Yes, you are right, I missed the removal of that one. Changing now.

workflow/workflow_SilentCorrectionsIngest.py
Jenniferstrej: It's necessary just for now as the same images are copied again in DepositAssets. When I remove this redundancy I will also remove VerifyImageServer from ProcessArticleZip and SilentCorrectionsProcess

giorgiosironi: I see this PR has been updated leaving ProcessArticleZip in the state we want it to be, but it doesn't have the removal of CopyGlencoeStillImages from SilentCorrectionsProcess (it's in there in `develop`)

Jenniferstrej: Sorry, my bad. fixed.

PR #442, opened 2017-04-21
removes redundancy from DepositAssets

Jenniferstrej:


giorgiosironi:
I see nothing strange but I'll go through this on Monday, checking the unit test coverage.

gnott:
I think the static method ``get_keys()`` might be not used anymore by this activity or any other and can probably be removed.

Otherwise looks clear and smart to exclude files from the full list of files and then copy those :+1:

activity/activity_DepositAssets.py
```diff
@@ -32,73 +34,72 @@ def do_activity(self, data=None):
         version = session.get_value(run, 'version')
         article_id = session.get_value(run, 'article_id')
 
-        self.emit_monitor_event(self.settings, article_id, version, run, "Deposit assets", "start",
+        self.emit_monitor_event(self.settings, article_id, version, run, self.pretty_name, "start",
                                 "Depositing assets for " + article_id)
```
giorgiosironi: The unit tests in `tests/activity/test_activity_deposit_assets.py` do not cover `do_activity`. I think two single positive and negative tests are necessary (one for the `return True` case and on for `ACTIVITY_PERMANENT_FAILURE`).

Jenniferstrej: Tests have been added.

PR #443, opened 2017-04-21
S3 consistency

giorgiosironi:
See https://github.com/elifesciences/builder-private/pull/17 and https://instrumentalapp.com/blog/why-you-should-stop-using-the-us-standard-region-in-s3-right-now/
Depends on https://github.com/elifesciences/bot-lax-adaptor/pull/179

### DIFF
--- a/provider/article.py
+++ b/provider/article.py
@@ -168,7 +168,7 @@ class article(object):
 
         # Download XML file via HTTP for now
         cdn_bucket_name = self.settings.publishing_buckets_prefix + self.settings.ppp_cdn_bucket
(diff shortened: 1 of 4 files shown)


gnott:
Side comment: I should check when the bot looks for article XML files it is using the newest CDN bucket location. It might be looking in the old CDN bucket.

giorgiosironi:
End2end should catch any host configuration problem

giorgiosironi:
And in fact, end2end caught a credentials problem caused by this in https://ci--alfred.elifesciences.org/job/test-elife-bot/234/console

giorgiosironi:
See https://github.com/elifesciences/bot-lax-adaptor/pull/184

PR #445, opened 2017-04-24
Original conversion is already happening under ConvertImagesToJPG

Jenniferstrej:
Last item of the checklist from ELPP-2495: Removes redundant copy of assets to CDN/published buckets from ResizeImages activity

Removes redundancy;
download: yes is being re-added to ConvertImagesToJPG because we don’t have an option to only copy `-download` images excluding the ones without the suffix. It could be added but is this additional complexity worth adding?

### DIFF
--- a/activity/activity_ConvertImagesToJPG.py
+++ b/activity/activity_ConvertImagesToJPG.py
@@ -48,6 +48,8 @@ class activity_ConvertImagesToJPG(activity.activity):
 
             figures = filter(article_structure.article_figure, files_in_bucket)
 
+            # download is not a IIIF asset but is currently kept for compatibility
+            # download may become obsolete in future
             formats = {"Original": {
                             "sources": "tif",
                             "format": "jpg",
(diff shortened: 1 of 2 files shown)


formats.yaml
```diff
@@ -79,11 +79,12 @@ Figure:
     format: gif
     resolution: 96
 
-  Original:
-    sources: tif
-    format: jpg
-    resolution: 96
-    download: yes
+#  Original images are being converted under ConvertImagesToJPG Activity
```
giorgiosironi: I agree that we should not convert again and update the S3 .jpg which has already been created before.

activity/activity_ConvertImagesToJPG.py
```diff
@@ -50,7 +50,8 @@ def do_activity(self, data=None):
 
             formats = {"Original": {
                             "sources": "tif",
-                            "format": "jpg"
+                            "format": "jpg",
+                            "download": "yes"
```
giorgiosironi: this is easy enough to do here, but I would add a comment explaining that `download` could probably be dropped in the future once no one links to that anymore

PR #446, opened 2017-04-24
VerifyImageServer only needs to run after Copying JPG and DepositIIIF…

Jenniferstrej:
…Assets

We’ve removed the redundancy, so we don’t need VerifyImageServer twice.


gnott:
Can the same duplication also be removed from https://github.com/elifesciences/elife-bot/blob/develop/workflow/workflow_SilentCorrectionsProcess.py?

Jenniferstrej:
(Face palm!) Silent Corrections, of course. New PR being shipped now!

PR #448, opened 2017-04-25
Remove PublishFullArticle workflow, the S3Monitor of the "full articl…

gnott:
…e zip" bucket, and the associated cron and starters.

As explained on https://elifesciences.atlassian.net/browse/ELPP-1942

For review, if I messed up somewhere deleting too much or too little - any comments / question?


giorgiosironi:
I am in favor of removing unused code, :+1: . I have however little knowledge of this workflow.

gnott:
If I deleted the right parts then we should not need these anymore. It was supposed to be temporary and the dependencies on unzipping the files to an old bucket are all gone now. I will merge and we can see what happens.

PR #450, opened 2017-05-03
Article provider to download XML from S3 from the new CDN bucket.

gnott:
https://elifesciences.atlassian.net/browse/ELPP-2604

### DIFF
--- a/provider/article.py
+++ b/provider/article.py
@@ -167,8 +167,8 @@ class article(object):
             return False
 
         # Download XML file via HTTP for now
-        cdn_bucket_name = self.settings.publishing_buckets_prefix + self.settings.ppp_cdn_bucket
-        xml_file_url = ('http://s3-external-1.amazonaws.com/' + cdn_bucket_name + '/'
+        published_bucket_name = self.settings.publishing_buckets_prefix + self.settings.published_bucket
+        xml_file_url = ('http://s3-external-1.amazonaws.com/' + published_bucket_name + '/articles/'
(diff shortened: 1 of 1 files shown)


giorgiosironi:
Depends on https://github.com/elifesciences/bot-lax-adaptor/pull/188 to be integrated in all environments first

provider/article.py
```diff
@@ -167,8 +167,8 @@ def download_article_xml_from_s3(self, doi_id=None):
             return False
 
         # Download XML file via HTTP for now
-        cdn_bucket_name = self.settings.publishing_buckets_prefix + self.settings.ppp_cdn_bucket
```
giorgiosironi: `ppp_cdn_bucket` is set to `elife-publishing-cdn`, which is the old CDN

provider/article.py
```diff
@@ -167,8 +167,8 @@ def download_article_xml_from_s3(self, doi_id=None):
...
-        xml_file_url = ('http://s3-external-1.amazonaws.com/' + cdn_bucket_name + '/'
+        published_bucket_name = self.settings.publishing_buckets_prefix + self.settings.published_bucket
```
giorgiosironi: `published_bucket` is `prod-elife-published`, but it sometimes set including the prefix and sometimes without it. Let me check the code.

giorgiosironi: Ok, it's consistently used with the prefix, this is the correct way to refer to it.

PR #451, opened 2017-05-05
Refactor to use storage provider and not conn.get_bucket

Jenniferstrej:
I would prefer to run a bot article run locally first to see if all these things work, it's difficult to unit test. I will do that on Monday.

I am observing the UpdateRepository temp file loading and I am not sure what it was used for and its use now. Maybe it became redundant. it looks like it's used for nothing.

### DIFF
--- a/provider/article.py
+++ b/provider/article.py
@@ -126,7 +126,7 @@ class article(object):
             self.article_type = parser.article_type(soup)
 
             self.authors = parser.authors(soup)
-            self.authors_string = self.authors_string(self.authors)
+            self.authors_string = self.get_authors_string(self.authors)
 
             self.related_articles = parser.related_article(soup)
 
@@ -174,7 +174,7 @@ class article(object):
         xml_filename = xml_file_url.split('/')[-1]
 
         r = requests.get(xml_file_url)
-        if r.status_code == requests.codes.ok:
+        if r.status_code == 200:
             filename_plus_path = self.get_tmp_dir() + os.sep + xml_filename
             f = open(filename_plus_path, "wb")
             f.write(r.content)
@@ -761,7 +761,7 @@ class article(object):
         #print lookup_url
 
         r = requests.head(lookup_url, allow_redirects=True)
-        if r.status_code == requests.codes.ok:
+        if r.status_code == 200:
             return r.url
         else:
             return None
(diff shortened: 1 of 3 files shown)


provider/article.py
```diff
@@ -808,17 +809,14 @@ def is_in_display_channel(self, display_channel):
 
     @staticmethod
     def get_bucket_files(settings, expanded_folder_name, xml_bucket):
```
giorgiosironi: looks like this method is only used inside here and in the module's unit tests, could it be renamed to `_get_bucket_files()` to imply its scope is limited?

provider/article.py
```diff
@@ -12,6 +12,7 @@
 import provider.s3lib as s3lib
```
giorgiosironi: there are a few warnings to solve to be able to add this file to `lint.sh`:
```
$ pylint -E provider/article.py 
No config file found, using default configuration
************* Module provider.article
E:176,28: Instance of 'LookupDict' has no 'ok' member (no-member)
E:763,28: Instance of 'LookupDict' has no 'ok' member (no-member)
E:776, 4: An attribute defined in provider.article line 128 hides this method (method-hidden)
```
First is using numbers (`200`) rather than a magic constant, and the other is a method to rename because it has the same name as a field. If we are able to put this under linting it will improve the confidence about its changes, as it's able to find undefined variables and other typos/basic errors.

Jenniferstrej: @giorgiosironi 
I did the changes that  are mentioned above. I saw another thing when running pylint:

E:814,26: Instance of 'StorageContext' has no 'list_resources' member (no-member)

The reason for that I believe is because the actual class that is instantiated at runtime is S3StorageContext. This was made in order to support other providers (File system for example). I could change S3StorageContext's name to StorageContext and remove the current StorageContext startup, but that wouldn't encourage people to use it for other providers... What do you think?

giorgiosironi: I saw it before, it's a more complex issue. In general I think the solution could be to slowly substitute the class `StorageContext` with a function `storage_context` that returns an instance of `S3StorageContext`: it's a strange result to instantiate a class and getting another, and this way we would keep the indirection.

However that's a story for another day, the use of `StorageContext` is widespread so it's a separate refactoring.

PR #453, opened 2017-05-08
activity success instead of return true

Jenniferstrej:


giorgiosironi:
If you rebase this should become green.

PR #455, opened 2017-05-09
Comparison fix

Jenniferstrej:
So I noticed in continuum some files saved in the published/cdn bucket and some had eLife as opposed to elife in the name and it seems to not to be making the comparison properly.
Will this change solve the problems or do you think we should look for a way to always save the file with lower case?


giorgiosironi:
Were do these wrong files come from? Are there wrong filenames on the Glencoe side? We have zero control on uploads to Glencoe as they happen outside of eLife, so would like to know more before we try to remedy here.

Jenniferstrej:
After I did this fix Melissa said she would ask Exeter to fix it, so it may not be necessary I suppose. I'll check and let you know.

PR #456, opened 2017-05-09
Removes reference to elife-published bucket by itself

Jenniferstrej:
second step of:
1. change `ppp_cdn_bucket` to the new value. the bot will write twice to the same place
2. change the bot removing every reference to `published_bucket`, which is not necessary anymore
3. rename the setting `ppp_cdn_bucket` to something longer-lasting (we create a new value, leave both old and new key available for a while, eventually removing the old one)

### DIFF
--- a/provider/storage_provider.py
+++ b/provider/storage_provider.py
@@ -6,12 +6,12 @@ import re
 import os
 
 
-def get_storage_context(*args):
+def StorageContext(*args):
     return S3StorageContext(args[0])
 
-class StorageContext(object):
-    def __new__(cls, *args):
-        return S3StorageContext(args[0])
+# class StorageContext(object):
+#     def __new__(cls, *args):
+#         return S3StorageContext(args[0])
 
 class S3StorageContext:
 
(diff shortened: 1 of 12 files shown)


giorgiosironi:
The removal looks good, could you try to add this activities to `lint.sh` if they are not already there? Progressively trying to get better feedback in development about changes to them, but I'm not sure if they have other weird things inside to make them passing.

Jenniferstrej:
@giorgiosironi - shall I add one by one? Or isn't better to do activity/activity_*.py? Is that because there are a lot of activities that we don't touch much and we would need to fix them first?

giorgiosironi:
You are correct, there are lots of warnings like
```
E: 52,32: Instance of 'Session' has no 'get_value' member (no-member)
E: 56, 8: Instance of 'Session' has no 'store_value' member (no-member)
```
on many activities when running it on `activity/*.py`, so adding them one at a time is easier to perform fixes.

provider/storage_provider.py
```diff
@@ -6,12 +6,12 @@
 import os
 
 
-def get_storage_context(*args):
+def StorageContext(*args):
```
giorgiosironi: ah, nice workaround

PR #457, opened 2017-05-10
SendQueuedEmail skipping invalid addresses

giorgiosironi:
Essentially we whitelist this kind of error to be logged with the right level, and go on sending to the other addresses. Previous behavior was to go into an apparently infinite loop (more than 12 hours at least), first of the activity inside the workflow and then of the workflow failing


lsh-0:
:+1:

giorgiosironi:
Need to merge this, as the daily update has removed my hotfix.

PR #460, opened 2017-05-16
removes published_bucket references

Jenniferstrej:
@gnott  - FYI, I am undoing what you did under article.py because we did the change the other way around, we changed the value of ppp_cdn_bucket to elife-published/articles so we are removing all references to the single bucket name value "elife-published".


gnott:
I see! thanks for the tip :+1:

PR #464, opened 2017-05-22
Publication email exception handling

Jenniferstrej:
@gnott - I added a try catch on the methods that showed error on New Relic.

### DIFF
--- a/activity/activity_PublicationEmail.py
+++ b/activity/activity_PublicationEmail.py
@@ -171,8 +171,9 @@ class activity_PublicationEmail(activity.activity):
 
             return True
         except Exception:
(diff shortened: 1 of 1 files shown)


gnott:
I think we approached this in different ways. I was troubleshooting PublicationEmail right now, and I didn't know you were working on it.

I have an alternate or additional solution (which I was running to clear the email backlog) https://github.com/elifesciences/elife-bot/pull/466

Jenniferstrej:
@gnott - I can close this PR. I didn't do much on it. I only added a try except around the do_activity and send_email. But it needs to be improved still

PR #466, opened 2017-05-23
Handle exception when email headers cannot be generated.

gnott:


activity/activity_PublicationEmail.py
```diff
@@ -603,16 +603,20 @@ def send_email(self, email_type, elife_id, author, article, authors):
...
+                article=article,
+                format="html")
+        except:
+            headers = None
 
         if not headers:
             log_info = ('Failed to load email headers for: doi_id: %s email_type: %s recipient_email: %s' %
                         (str(elife_id), str(email_type), str(author.e_mail)))
             self.admin_email_content += "\n" + log_info
+            return False
```
Jenniferstrej: return False: Do we want to retry the activity?

Jenniferstrej:
Are you going to catch exceptions that happen during do_activity? It would be good since New Relic has seen some errors happening on this activity but we don't catch or log those errors. Also I am not sure whether it needs to be a "permanent failure" or a retry one. (?)

gnott:
The new ``return False`` here would just mean for one article the emails will not get sent, and it will stop failing the entire activity when it encounters some bad data.

We can certainly add more exceptions around other places in the code in subsequent PR if you want to. 

Is this one ok to merge and then you can add more exceptions ?

activity/activity_PublicationEmail.py
gnott: As below, if we retry the activity it would still fail, if the bad data is still around. This return False just means this particular article from the list of articles it is trying to process will get ignored, and the activity can continue to reach a successful state so it will not churn forever.

PR #467, opened 2017-05-24
Adds PDF figures to cdn before ingest

Jenniferstrej:
Obs: 
Created a new file_type under article_structure to filter PDF figures;
Changed the name of DepositIIIFAssets to DepositIngestAssets to include PDF figures.

### DIFF
--- a/activity/activity_DepositAssets.py
+++ b/activity/activity_DepositAssets.py
@@ -51,8 +51,11 @@ class activity_DepositAssets(activity.activity):
 
             # filter figures that have already been copied (see DepositIIIFAssets activity)
             original_figures = article_structure.get_figures_for_iiif(files_in_bucket)
-            original_figures_and_videos = original_figures + article_structure.get_videos(files_in_bucket)
-            other_assets = filter(lambda asset: asset not in original_figures_and_videos, files_in_bucket)
+            iiif_assets = original_figures + article_structure.get_videos(files_in_bucket)
+            pdf_figures = article_structure.get_pdf_figures(files_in_bucket)
+            pre_ingest_assets = iiif_assets + pdf_figures
+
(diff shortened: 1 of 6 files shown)


activity/activity_DepositIngestAssets.py
```diff
@@ -5,22 +5,22 @@
 from provider import article_structure
 
 """
-DepositIIIFAssets.py activity
+DepositIngestAssets.py activity
```
giorgiosironi: good name

activity/activity_DepositAssets.py
```diff
@@ -51,8 +51,11 @@ def do_activity(self, data=None):
 
             # filter figures that have already been copied (see DepositIIIFAssets activity)
             original_figures = article_structure.get_figures_for_iiif(files_in_bucket)
-            original_figures_and_videos = original_figures + article_structure.get_videos(files_in_bucket)
-            other_assets = filter(lambda asset: asset not in original_figures_and_videos, files_in_bucket)
+            iiif_assets = original_figures + article_structure.get_videos(files_in_bucket)
+            pdf_figures = article_structure.get_pdf_figures(files_in_bucket)
+            pre_ingest_assets = iiif_assets + pdf_figures
```
giorgiosironi: this `pre_ingest_assets` seems duplicated between the two activities (one uses them, the other excludes them); seems a good candidate for extraction in `article_structure` too even if it's just a sum of two lists.

provider/article_structure.py
```diff
@@ -131,6 +138,9 @@ def get_figures_for_iiif(files):
     fs = originals_figures_tif + get_media_file_images(files)
     return fs
 
+def get_pdf_figures(files):
```
gnott: A naming suggestion I would use {{get_figures_pdfs}} which is a little more consistent with "figures PDF" terminology.

PR #468, opened 2017-05-26
Update elifetools version.

gnott:


gnott:
Tests pass I will merge now.

PR #470, opened 2017-05-31
Feature preserve xml jats doctype

gnott:
Should work now with elifetools 0.1.7 (will see whether the tests pass).

Takes advantage of the new feature of ``xmlio.py`` in elifetools to get the doctype from an XML file when it is parsed in, and then later add the same doctype back again when writing the altered XML out again.

It is to support the coming new XML format which is JATS v1.1, whereas before all XML files that were rewritten were given a JATS v1.1d3 doctype.

Not a rush but should be merged before the new eLife XML format is started to be delivered, which is I think still in a week or two at the earliest.


gnott:
Backwards compatible, this now taking advantage of the enhancement. The tests on the XML files with a different doctype seems to show it is working in practice - let's merge and hope that is true.

PR #471, opened 2017-06-02
more exception handling on Publication email

Jenniferstrej:


gnott:
If there's an exception, do we get no return value?

I'm curious whether if we have one email's headers fail, will the other emails get sent? I.e., does the exception raised in setting headers allow the exception upstream from it in do_activity() let it continue to send emails later in the list of articles that is iterated?

This might do with some further refactoring in the future, due to the complexity of what it does.

PR #472, opened 2017-06-02
verify version and last modified date on S3 bucket

Jenniferstrej:
The intention is to avoid the problem occurred on ELPP-2731 where there were more than one archive zip files

### DIFF
--- a/activity/activity_PubRouterDeposit.py
+++ b/activity/activity_PubRouterDeposit.py
@@ -233,8 +233,10 @@ class activity_PubRouterDeposit(activity.activity):
                                self.settings.aws_secret_access_key)
         bucket = s3_conn.lookup(bucket_name)
 
+        s3_keys_in_bucket = s3lib.get_s3_keys_from_bucket(bucket=bucket)
+
         s3_keys = []
-        for key in bucket:
+        for key in s3_keys_in_bucket:
             s3_keys.append({"name": key.name, "last_modified": key.last_modified})
 
         return self.latest_archive_zip_revision(article.doi_id, s3_keys, self.journal, status)
--- a/activity/activity_PubRouterDeposit.py
+++ b/activity/activity_PubRouterDeposit.py
@@ -14,6 +14,7 @@ import provider.article as articlelib
 import provider.s3lib as s3lib
 import provider.blacklist as blacklist
 
+import dateutil.parser
 """
 PubRouterDeposit activity
 """
@@ -233,11 +234,15 @@ class activity_PubRouterDeposit(activity.activity):
                                self.settings.aws_secret_access_key)
         bucket = s3_conn.lookup(bucket_name)
 
-        s3_key_names = s3lib.get_s3_key_names_from_bucket(bucket=bucket)
+        s3_keys = []
+        for key in bucket:
+            s3_keys.append({"name": key.name, "last_modified": key.last_modified})
 
-        return self.latest_archive_zip_revision(article.doi_id, s3_key_names, self.journal, status)
+        #s3_key_names = s3lib.get_s3_key_names_from_bucket(bucket=bucket)
 
-    def latest_archive_zip_revision(self, doi_id, s3_key_names, journal, status):
+        return self.latest_archive_zip_revision(article.doi_id, s3_keys, self.journal, status)
+
+    def latest_archive_zip_revision(self, doi_id, s3_keys, journal, status):
         """
         Get the most recent version of the article zip file from the
         list of bucket key names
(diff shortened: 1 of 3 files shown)


tests/activity/test_activity_pub_router_deposit.py
```diff
@@ -0,0 +1,26 @@
...
+import settings_mock
+from ddt import ddt, data, unpack
+
+@ddt
+class TestPubRouterDeposit(unittest.TestCase):
+    def setUp(self):
+        self.pubrouterdeposit = activity_PubRouterDeposit(settings_mock, None, None, None, None)
+
+    @unpack
+    @data({"input": [{"name": "elife-16747-vor-v1-20160831000000.zip", "last_modified": "2017-05-18T09:04:11.000Z"},
```
giorgiosironi: could you add a comment for each of the datasets here? I think I understand what it does, but we'll forget quickly

Jenniferstrej: I'm not sure how you mean, but I added a comment at the top of the test. Please let me know if that helps. Thanks.

activity/activity_PubRouterDeposit.py
```diff
@@ -225,19 +226,20 @@ def archive_zip_file_name(self, article, status='vor'):
...
         bucket_name = self.archive_bucket
 
         # Connect to S3 and bucket
         s3_conn = S3Connection(self.settings.aws_access_key_id,
                                self.settings.aws_secret_access_key)
         bucket = s3_conn.lookup(bucket_name)
 
-        s3_key_names = s3lib.get_s3_key_names_from_bucket(bucket=bucket)
+        s3_keys = []
+        for key in bucket:
```
gnott: This may need to be
```
for key in bucket.list():
```

?

Jenniferstrej:
@giorgiosironi - I made the changes requested by @gnott . Is that possible to add on end2end the bit that runs PubRouterDeposit (what do you think @gnott ?).
It would be fairly straightforward for example to test archive_zip_file_name function as that would need a couple of zip files added to the archive bucket, then check if the returned value was the most recent file and highest version.
@giorgiosironi - where is the best place to do this? I could have a go with putting that on end2end, but I would just need a bit of direction as I never touched that area.

giorgiosironi:
Somewhere in https://github.com/elifesciences/elife-spectrum/blob/master/spectrum/test_article.py although the project could need a large `app.cfg` file to run locally (directed to end2end). Some questions:
1. this is a previously not tested workflow, who triggers it? Usually workflows are triggered by S3 buckets or by the dashboard. elife-spectrum uses econtools if that can help? 
2. it seems it depends on an outbox, which workflows populate it?
3. it depends on an FTP server, we can provide that on end2end to simulate Scopus/Cencage/etc.
From the end2end perspective we do not have access to the Python processes that the bot runs, so we can only run large tests rather then checking a single function like `archive_zip_file_name`. That is more easily tested at the unit level.

gnott:
I will attempt to answer the questions about testing

1. ``PubRouterDeposit`` workflow of the ``PMC`` type is triggered by the cron, and reads the outbox contents to figure out which articles it should attempt to process
2. The outbox gets populated as part of the ``ScheduleDownstream`` activity as part of the ``PostPerfectPublication`` workflow
3. FTP would be a great addition to end2end tests (I wasn't aware we had it now). Testing the PubRouterDeposit ability to locate the latest zip file to transform it into a PMC zip doesn't require FTP though, that is just the final step after finding the input zip, transforming the XML and the contents, and then uploading it to S3 as a final product.

Is this something that you can help with @giorgiosironi? Or, please can you provide more instructions since I haven't tried to create any end2end tests (if you'd like someone else to try and create it).

Or, are we happy to merge this and just let it run on production. I am ok with that too, if I know when it gets deployed so we can monitor it.

giorgiosironi:
Long story short I can help, but:
- not in a short timeframe as I am leaving for a conference tomorrow, so you may want to deploy this and monitor it manually
- I need a way to trigger the `PubRouterDeposit` workflow on-demand from outside the elife-bot instance, since the tests cannot wait for it to run every N minutes/hours. A message in a queue can do this? Calling the SWF API directly? (but I would need parameters)

giorgiosironi:
Moving the end2end testing part to https://elifesciences.atlassian.net/browse/ELPP-2851

gnott:
On demand PubRouterDeposit is possible using its starter, for example

```
starter/starter_PubRouterDeposit.py -e {{env_name}} -w PMC
```

The above would replicate what ``cron.py`` does once a day, and the workflow would run what it finds in the S3 outbox.

I'm ok with merging and we can monitor for errors.

PR #473, opened 2017-06-20
Version should not be limited to a single digit

giorgiosironi:
We have v10 for the kitchen sink in the testing environments


gnott:
Good find @giorgiosironi, I don't think we'd encounter a v10 elsewhere than the test environment.

PR #475, opened 2017-07-04
More retries in VerifyImageServer

giorgiosironi:
This activity sometimes fail because of an image that times out when downloaded from S3 inside the IIIF server. We do our best to detect this and retry once


gnott:
It seems reasonable to retry after a 404 error :+1:

PR #476, opened 2017-07-12
Article check_was_ever_poa_by_lax() function, with tests and replaced…

gnott:
… in activities that use it.

https://elifesciences.atlassian.net/browse/ELPP-2266

Basically, using Lax data will bypass the article provider's ``get_was_poa_doi_ids()`` function, which was checking PoA bucket contents to find all the articles that were ever PoA'ed. Using Lax data should speed things up, and also fix a bug where the old function was returning a false positive for articles that were prepared as a PoA but were never published as a PoA, cascading into a PubMed deposit ``<Replaces>`` tag bug.

Appreciate if someone can check I didn't mess up anywhere.

### DIFF
--- a/provider/article.py
+++ b/provider/article.py
@@ -574,26 +574,6 @@ class article(object):
         else:
             return False
 
-    def check_was_ever_poa_by_lax(self, doi):
-        """
-        Use Lax data to check if the article was ever a PoA article
-        Return true or false if Lax has status_code 200, otherwise return None
-        """
-        doi_id = int(self.get_doi_id(doi))
-        article_id = str(doi_id).zfill(5)
-        # work around circular dependency article/lax_provider
-        from lax_provider import article_versions, poa_vor_status
-        status_code, data = article_versions(article_id, self.settings)
-
-        if status_code == 200:
-            poa_status, vor_status = poa_vor_status(data)
-            if poa_status is True:
-                return True
-            else:
-                return False
-        else:
-            return None
-
     def get_doi_id_from_poa_s3_key_name(self, s3_key_name):
         """
         Extract just the integer doi_id value from the S3 key name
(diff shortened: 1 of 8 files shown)


provider/article.py
```diff
@@ -574,6 +574,26 @@ def check_was_ever_poa(self, doi):
...
 
+    def check_was_ever_poa_by_lax(self, doi):
+        """
+        Use Lax data to check if the article was ever a PoA article
+        Return true or false if Lax has status_code 200, otherwise return None
+        """
+        doi_id = int(self.get_doi_id(doi))
+        article_id = str(doi_id).zfill(5)
+        # work around circular dependency article/lax_provider
+        from lax_provider import article_versions, poa_vor_status
```
giorgiosironi: to break the circular dependency, this function could go to into `lax_provider` right? After all, it is dependent on Lax so it would fit there

gnott: How this function is used it is a good idea to move it to ``lax_provider``, I will look at it more carefully. The doi value gets passed as an argument and the ``article`` object is not asking itself "was I ever PoA?".

There's aother function in the ``article`` provider that imports ``lax_provider`` can possibly get refactored too in the future.

provider/lax_provider.py
```diff
@@ -103,6 +103,18 @@ def poa_vor_status(data):
     return poa_status, vor_status
 
 
+def was_ever_poa(article_id, settings):
+    "Use Lax data to check if the article was ever a PoA article"
+    status_code, data = article_versions(article_id, settings)
+    if status_code == 200:
+        poa_status, vor_status = poa_vor_status(data)
+        if poa_status is True:
```
giorgiosironi: could be just `return poa_status` maybe?

gnott: I think by my reading of {{poa_vor_status()}} is it will return {{None}} if there is no PoA, and I'd like this new function to return False in that case. {{None}} is reserved for when something has gone wrong.

PR #479, opened 2017-07-28
PDF covers generation, prod url and adapting response

Jenniferstrej:
These changes are assuming that the PDF covers request response is according to @garyttierney 's pointers on https://elifesciences.atlassian.net/browse/ELPP-2047.

If those are ok, we need to add the settings variable pdf_cover_generator before merging (cc @giorgiosironi).

### DIFF
--- a/activity/activity_GeneratePDFCovers.py
+++ b/activity/activity_GeneratePDFCovers.py
@@ -35,8 +35,16 @@ class activity_GeneratePDFCovers(activity.activity):
 
             article = articlelib.article()
 
-            assert hasattr(self.settings, "pdf_cover_generator"), "pdf_cover_generator variable is missing from " \
-                                                                  "settings file!"
+            if not hasattr(self.settings, "pdf_cover_generator") or \
+                    (hasattr(self.settings, "pdf_cover_generator") and self.settings.pdf_cover_generator == None) or \
+                    (hasattr(self.settings, "pdf_cover_generator") and len(self.settings.pdf_cover_generator) < 1):
+
+                self.emit_monitor_event(self.settings, article_id, version, run,
+                                        self.pretty_name, "start", "pdf_cover_generator variable is missing from "
+                                                          "settings file. PDF not generated but flag is set "
+                                                          "for the activity to succeed.")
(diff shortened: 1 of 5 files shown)


activity/activity_GeneratePDFCovers.py
```diff
@@ -34,13 +34,19 @@ def do_activity(self, data):
                                     self.pretty_name, "start", "Starting check for generation of pdf cover.")
 
             article = articlelib.article()
-            pdf_cover_a4 = article.get_pdf_cover_link(self.logger, self.settings, article_id, "a4")
-            pdf_cover_letter = article.get_pdf_cover_link(self.logger, self.settings, article_id, "letter")
+
+            assert hasattr(self.settings, "pdf_cover_generator"), "pdf_cover_generator variable is missing from " \
```
giorgiosironi: can we switch on the presence of `pdf_cover_generator` in the `settings` and only perform these actions if the settings is present and not `None`|`False`? We would then be able to merge the code but keep the feature behind a feature flag, disabling it if it's unstable

Jenniferstrej: There is a check as you pointed out. Are you saying I should conclude the activity as a *Success* if there is no ```pdf_cover_generator``` in ```settings```? Currently the activity just errors when the value isn't present.

giorgiosironi: I think so, so that we don't stop publication if something goes wrong (at the moment, we will make it stricter later)

Could you also add it to the settings-example.py?

Jenniferstrej: Yes, I will. Thanks!

giorgiosironi: Added the setting to `end2end` and `prod`, creating `personalised-covers--continuumtest` to add it there too.

Jenniferstrej:
@giorgiosironi - changes made. setting-example.py already had the pdf covers settings there https://github.com/elifesciences/elife-bot/blob/develop/settings-example.py#L191-L193

Jenniferstrej:
👍

PR #483, opened 2017-08-02
LensArticle automated test cases, a little code clean up and make it …

gnott:
…easier to test.

### DIFF
--- a/activity/activity_LensArticle.py
+++ b/activity/activity_LensArticle.py
@@ -5,8 +5,8 @@ import codecs
 
 import activity
 
-import boto.s3
 from boto.s3.connection import S3Connection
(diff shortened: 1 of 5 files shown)


gnott:
In preparation to remove the filesystem provider, this activity has test cases now. Removing filesystem from it later should be trivial anyway, and hopefully the test case will make more sure it is ok.

giorgiosironi:
The changes are simple and ok. The tests are failing though?

gnott:
Thanks for taking a look. I think I just reused a value with a leading slash for confirming an s3 key name as well as writing to the disk during the test. The test temp directory does not like leading slashes, I just replaced the filename to write with ``"index.html"`` and is equivalent.

PR #487, opened 2017-08-07
Elpp 2963 - InvalidateCdn Activity

Jenniferstrej:


activity/activity_InvalidateCdn.py
```diff
@@ -0,0 +1,67 @@
...
+"""
+activity_InvalidateCdn.py activity
+"""
+
+class activity_InvalidateCdn(activity.activity):
+    def __init__(self, settings, logger, conn=None, token=None, activity_task=None):
+        activity.activity.__init__(self, settings, logger, conn, token, activity_task)
+
+        self.name = "InvalidateCdn"
+        self.pretty_name = "CF Invalidate Cdn"
```
giorgiosironi: `CloudFront` is prettier than `CF` imho. However users of the dashboard may find another name even prettier (assuming this appears on the dashboard) @sianroderick what do you want written there?

activity/activity_InvalidateCdn.py
```diff
@@ -0,0 +1,67 @@
...
+        self.version = "1"
+        self.default_task_heartbeat_timeout = 30
+        self.default_task_schedule_to_close_timeout = 60 * 5
+        self.default_task_schedule_to_start_timeout = 30
+        self.default_task_start_to_close_timeout = 60 * 5
+        self.description = "Invalidate Cdn todo"
```
giorgiosironi: where does the description end up? and what should we write there?

activity/activity_InvalidateCdn.py
```diff
@@ -0,0 +1,67 @@
...
+
+            ### If we want to run Invalidation only if CDN has been previously populated
+            # if "files_in_cdn" in data:
+            #     if data["files_in_cdn"] == True:
+            #         cloudfront_provider.create_invalidation(article_id)
+            #         dashboard_message = "CloudFront Invalidation command sent for article %s." % str(article_id)
+            #     else:
+            #         dashboard_message = "CloudFront Invalidation was not necessary for article %s." % str(article_id)
+
+            cloudfront_provider.create_invalidation(article_id, self.settings)
```
giorgiosironi: I like this interface, it's minimal

activity/activity_InvalidateCdn.py
```diff
@@ -0,0 +1,67 @@
...
+
+            dashboard_message = "CloudFront Invalidation command sent for article %s." % str(article_id)
+            self.emit_monitor_event(self.settings, article_id, version, run,
+                                    self.pretty_name, "end", dashboard_message)
+            return activity.activity.ACTIVITY_SUCCESS
+
+        except AssertionError as err:
```
giorgiosironi: can the two `except` blocks be unified since there is no difference between them?

provider/cloudfront_provider.py
```diff
@@ -0,0 +1,8 @@
+import boto.cloudfront as cloudfront
+
+
+def create_invalidation(article, settings):
```
giorgiosironi: possible alternative: `create_invalidation(article, distribution_id)`, would avoid dependency on `settings` here but I'm not sure how it's done in other providers

settings-example.py
```diff
@@ -335,6 +337,17 @@ class dev():
...
+     # PDF cover
+    pdf_cover_generator = "http://localhost:8082/personalcover/generate/"
+    pdf_cover_landing_page = "http://localhost:8082/personalcover/landing/"
+
+    # IIIF
+    path_to_iiif_server = "https://pathto--iiif.elifesciences.org/"
+    iiif_resolver = "{article_id}/{article_fig}/full/full/0/default.jpg"
+
+    # CloudFront
+    cloudfront_distribution_id_cdn = "DISTRIBUTIONID"
```
giorgiosironi: I will add this to the deployed settings.py

activity/activity_InvalidateCdn.py
Jenniferstrej: Sorry, I forgot to edit. It's just for documentation and it may appear on the console view when the activity is registered.

provider/cloudfront_provider.py
Jenniferstrej: I first thought I didn't want something specific to the API to be especified on the activity, but since the activity is just for cloudfront purposes I believe passing the distribution id is not a problem.

giorgiosironi:
Alright, I missed silent corrections but they should include this as well.

PR #495, opened 2017-08-21
for logging of resp.json

Jenniferstrej:
Any suggestion for further log?

### DIFF
--- a/provider/article.py
+++ b/provider/article.py
@@ -243,8 +243,8 @@ class article(object):
         assert (resp.status_code == 200), "unhandled status code from PDF cover service: %s . " \
                                           "Format: %s - url requested: %s" % \
                                           (resp.status_code, format, url)
-        logger.info("PDF Generator Response %s", str(resp))
         data = resp.json()
+        logger.info("PDF Generator Response %s", str(data))
         return data['formats']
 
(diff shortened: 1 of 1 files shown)


provider/article.py
```diff
@@ -243,8 +243,8 @@ def get_pdf_cover_link(self, pdf_cover_generator_url ,doi_id, logger):
         assert (resp.status_code == 200), "unhandled status code from PDF cover service: %s . " \
                                           "Format: %s - url requested: %s" % \
                                           (resp.status_code, format, url)
-        logger.info("PDF Generator Response %s", str(resp))
         data = resp.json()
+        logger.info("PDF Generator Response %s", str(data))
```
giorgiosironi: perhaps `resp.content` in case the response is not a JSON (when catching `ValueError: No JSON object could be decoded`)

Jenniferstrej: The "JSON object could not be decoded" error only has happened when the PDF Covers API returned an empty body but with 200 status code. 
The new error "formats" is the one that is strange and I haven't been able to tell what it is, since the covers API returned a valid body this time.

giorgiosironi: Useful to log the whole parsed response, can't think of additional logs for that problem in particular

PR #499, opened 2017-09-07
CNPIEC deliveries and FTPArticle refactoring

gnott:
These changes started out as wanting to add a new recipient of article files, CNPIEC. 

I found the FTPArticle activity coding to be outdated, because if we did not have a PMC zip file available to use as the input for the delivery, it was downloading old zip files from the old buckets.

To get this working better, it will download an archive zip file if no PMC zip file is available, then repackage the archive zip into PMC zip file format.

There was logic in the PMCDeposit activity that could be reused, so I created a new provider named ``article_processing.py`` which holds these shared functions (later to be used in PMCDeposit activity when it can be refactored).

I also added new tests for the FTPArticle activity and for the s3lib provider which we didn't have yet.

Internally this is related to ticket https://elifesciences.atlassian.net/browse/ELPP-1357. More steps can come later to start sending to CNPIEC automatically, but first I would like to merge and deploy this code and then start to send older eLife articles with manually started workflows before the full automation is turned on.


provider/article_processing.py
```diff
@@ -0,0 +1,140 @@
...
+        else:
+            renamed_list.append(k)
+
+    return (verified, renamed_list, not_renamed_list)
+
+
+def new_pmc_zip_filename(journal, volume, fid, revision=None):
+    filename = journal
+    filename = filename + '-' + str(volume).zfill(2)
+    filename = filename + '-' + str(fid).zfill(5)
```
giorgiosironi: to avoid the `5` magic number, the `str(fid).zfill(5)` can be extracted in a `padded_id` (or similar name) function; it is duplicated in a few places

provider/article_processing.py
```diff
@@ -0,0 +1,140 @@
...
+
+    highest = 0
+    for key in s3_keys:
+        if key["name"].startswith(name_prefix_to_match):
+            version_and_date = None
+            try:
+                parts = key["name"].split(name_prefix_to_match)
+                version = parts[1].split('-')[0]
+                date_formatted = dateutil.parser.parse(key["last_modified"])
+                date_part = date_formatted.strftime('%Y%m%d%H%M%S')
```
giorgiosironi: also this format looks like a constant that should be extracted with a good name e.g. `S3_DATE_FORMAT` or `ELIFE_DATE_FORMAT`

provider/article_processing.py
gnott: Agreed, I've added a ``utils.py`` provider to the elife-bot with a ``pad_msid()`` and ``pad_volume()``, using python ``string.format()`` method

provider/article_processing.py
gnott: The value in this situation should be constant since it is just comparing date values in string format. It also would probably not vary by environment or by journal, so ``settings.py`` does not seem to be a good place for it. I just added it to the new ``utils.py`` provider since that would be widely available when it is imported by other parts of the overall system. Is this a good place for this value for now?

provider/utils.py
```diff
@@ -0,0 +1,8 @@
+
+S3_DATE_FORMAT = '%Y%m%d%H%M%S'
```
giorgiosironi: yes, this is a good place as in `setting.py` it would be very opaque (putting it in `builder-private` rather than in the repository), and doesn't really need to be customized as it is a wide convention

PR #501, opened 2017-09-14
Refactor filesystem provider

gnott:
Goal of this PR is to remove the ``filesystem.py`` provider completely. It was half-baked and not used much (fortunately), where we rely now on temporary directories and the activities write files directly to those.

This adds automated tests for providers that had no tests before, ``ejp.py`` and ``templates.py``.

Then the filesystem provider was removed from those two ejp and templates providers.

filesystem provider in the LensArticle activity was not really in use and just deleted.

A minor change to the PublicationEmail activity for an argument name change done in the ejp provider to make the document name more clear.

With the tests backing these I'm fairly certain there are no bugs. Where they may happen is where files are downloaded from S3 which I didn't test yet. Not all of these are covered by end2end tests either. I'd like to know when this is deployed after merging in case there are issues.

Relevant internal ticket https://elifesciences.atlassian.net/browse/CR-23


provider/ejp.py
```diff
@@ -76,22 +72,20 @@ def get_s3key(self, s3_key_name, bucket=None):
...
         return s3key
 
+    def write_content_to_file(self, filename, content, mode="wb"):
+        "write the content to a file in the tmp_dir"
+        document = os.path.join(self.get_tmp_dir(), filename)
+        with open(document, 'wb') as fp:
+            fp.write(content)
+            return document
+        # default if it does not work
+        return None
```
giorgiosironi: silent failure? Would an exception be more explicit if a file cannot be written?

giorgiosironi: although I think the `return document` won't be reached only in case an exception is raised before it, so this `return None` line may be unreachable code

tests/features/filesystem_steps.py
```diff
@@ -36,31 +35,10 @@ def i_get_the_tmp_dir_from_the_world(step):
   if(tmp_dir == ""):
     tmp_dir = None
   world.tmp_dir = tmp_dir
+  # also try to create the directory if it does not exist
+  try:
+    os.mkdir(world.tmp_dir)
+  except OSError as e:
```
giorgiosironi: `OSError` has an `errno` field that tells which kind of error has happened. The error about an already existing directory can be ignored, but other kind of `OSError` exceptions should be re-reraised as they are actual failure (e.g. permissions problems). See https://stackoverflow.com/a/600612/91590

provider/ejp.py
gnott: I did create that function quickly to get tests to pass and to make writing content to disk more general purpose after removing the filesystem provider. I added some more exceptions and some specific tests for it. I hope it is an improvement!

tests/features/filesystem_steps.py
gnott: I like what you are describing about differentiating between directory exists and other undesirable disk issues. My suggestion is we create a new ticket for this to fix ones that are not in the lettuce tests (which will not be around for long). There are at least 5 similar occurrences of this OSError in activity code which could be improved in a separate coding task.

giorgiosironi: Sure, if the code is duplicated it is worth extracting it into a utility function

gnott: Cool, I created a JIRA ticket for myself https://elifesciences.atlassian.net/browse/ELPP-3156

gnott:
I'll merge this in now and we can try it out!

PR #504, opened 2017-09-22
Test case for DepositCrossref activity.

gnott:
Adding tests for DepositCrossref activity, where there were no tests. I wanted to do this and merge it in before I start refactoring it to use the new Crossref generation library.


tests/activity/test_activity_deposit_crossref.py
```diff
@@ -0,0 +1,85 @@
...
+from activity.activity_DepositCrossref import activity_DepositCrossref
+import shutil
+from mock import patch
+import tests.activity.settings_mock as settings_mock
+from provider.article import article
+from provider.simpleDB import SimpleDB
+
+import os
+# Add parent directory for imports, so activity classes can use elife-poa-xml-generation
+PARENTDIR = os.path.dirname(os.path.dirname(
```
giorgiosironi: does `os.path.realpath(os.path.dirname(os.path.abspath(__file__)) + '/../../')` simplify this?

tests/activity/test_activity_deposit_crossref.py
```diff
@@ -0,0 +1,85 @@
...
+        # Open the crossref XML and check some of its contents
+        with open(crossref_xml_filename_path, 'rb') as fp:
+            crossref_xml = fp.read()
+            for expected in expected_crossref_xml_contains:
+                try:
+                    self.assertTrue(expected in crossref_xml)
+                except AssertionError:
+                    print expected, ' not found in crossref_xml'
+                    raise
+
```
giorgiosironi: long
empty
space? :)

tests/activity/test_activity_deposit_crossref.py
gnott: It seems to with ``'/../../../'`` at least locally. I used the other method years ago with a concern it would not be Windows compatible but the forward slashes in this case seem to be ok.

The refactoring I'm up to will eventually no longer need to import anything from the parent directory, but it is still at least a couple months away before it is entirely finished, I would estimate.

tests/activity/test_activity_deposit_crossref.py
gnott: I like lots of space, don't you? I will trim a little of it before merging.

PR #505, opened 2017-09-26
Schedule crossref tests

gnott:
As part of integrating the new Crossref generation library, I noticed the ScheduleCrossref activity did not have test cases, and I first added them here.

Reviewing it, I noticed it imported a static method from the ConvertJATS activity in order to find the article XML key from the expanded article bucket. To refactor this to a more logical place, I put the function into the ``article_structure.py`` provider, since it is closely related to knowing what the article XML file name is.

There are at least three other activities that use a ``get_article_xml_key()`` function somewhere, and those can be refactored in the future as part of a new PR, presuming this PR gets merged in.

### DIFF
--- a/activity/activity_ScheduleCrossref.py
+++ b/activity/activity_ScheduleCrossref.py
@@ -5,7 +5,7 @@ import activity
 from boto.s3.connection import S3Connection
 
 from provider.execution_context import Session
-from activity_ConvertJATS import activity_ConvertJATS as ConvertJATS
+from provider.article_structure import get_article_xml_key
 
 """
 ScheduleCrossref.py activity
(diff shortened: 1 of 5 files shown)


activity/activity_ScheduleCrossref.py
```diff
@@ -5,7 +5,7 @@
 from boto.s3.connection import S3Connection
 
 from provider.execution_context import Session
-from activity_ConvertJATS import activity_ConvertJATS as ConvertJATS
+from provider.article_structure import get_article_xml_key
```
giorgiosironi: I agree putting the function on a provider makes it more reusable (and testable in isolation)

provider/article_structure.py
```diff
@@ -180,6 +180,17 @@ def pre_ingest_assets(files):
     return list(set(iiif_assets + pdf_figures))
 
 
+def get_article_xml_key(bucket, expanded_folder_name):
+    "locate the article XML file in the expanded article bucket on S3"
```
giorgiosironi: the docstring could say two values are returned, rather than only the key

gnott: Good point, thanks @giorgiosironi!

PR #506, opened 2017-09-28
Retry on SSL timeout from Github, 2nd attempt

giorgiosironi:
This is a particular error we encounter when load is too high in pushing to a Github repository.

We also add tests for the activity, which go for the happy path and both the permanent/temporary failure cases.

### DIFF
--- a/provider/storage_provider.py
+++ b/provider/storage_provider.py
@@ -13,6 +13,9 @@ def StorageContext(*args):
 #     def __new__(cls, *args):
 #         return S3StorageContext(args[0])
 
+def storage_context(*args):
+    return S3StorageContext(*args)
+
 class S3StorageContext:
 
     def __init__(self, settings):
--- a/tests/activity/classes_mock.py
+++ b/tests/activity/classes_mock.py
@@ -97,9 +97,8 @@ class FakeStorageContext:
     def get_resource_to_file(self, resource, filelike):
         bucket_name, s3_key = self.get_bucket_and_key(resource)
         src = data.ExpandArticle_files_source_folder + s3_key
-        if type(filelike) == file:
-            with open(src) as fsrc:
-                copyfileobj(fsrc, filelike)
+        with open(src) as fsrc:
+            copyfileobj(fsrc, filelike)
 
     def get_resource_as_string(self, origin):
         return '<mock><media content-type="glencoe play-in-place height-250 width-310" id="media1" mime-subtype="wmv" mimetype="video" xlink:href="elife-00569-media1.wmv"></media></mock>'
--- a/tests/activity/classes_mock.py
+++ b/tests/activity/classes_mock.py
@@ -1,6 +1,6 @@
 from testfixtures import TempDirectory
 import test_activity_data as data
-from shutil import copyfile
+from shutil import copyfile, copyfileobj
 from shutil import copy
 import shutil
 import re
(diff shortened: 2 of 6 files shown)


provider/storage_provider.py
```diff
@@ -13,6 +13,9 @@ def StorageContext(*args):
 #     def __new__(cls, *args):
 #         return S3StorageContext(args[0])
 
+def storage_context(*args):
```
gnott: I'm wondering if ``def StorageContext(*args)`` is equivalent to this (further up in this provider) and whether this is duplication of the same function?

tests/activity/classes_mock.py
```diff
@@ -93,9 +94,12 @@ def get_bucket_and_key(self, resource):
         s3_key = match.group(3)
         return bucket_name, s3_key
 
-    def get_resource_to_file(self, resource, file):
+    def get_resource_to_file(self, resource, filelike):
         bucket_name, s3_key = self.get_bucket_and_key(resource)
-        copyfile(data.ExpandArticle_files_source_folder + s3_key, file.name)
+        src = data.ExpandArticle_files_source_folder + s3_key
+        if type(filelike) == file:
```
gnott: Is there a case in the test cases where a non-file is supplied to this function? Maybe in some end2end live bucket testing?

provider/storage_provider.py
giorgiosironi: afaik it's equivalent for the existing use case as only the first element of `args` is passed:
```
$ grep -r "\bStorageContext(" .
./activity/activity_VerifyGlencoe.py:            storage_context = StorageContext(self.settings)
./activity/activity_CopyGlencoeStillImages.py:        storage_context = StorageContext(self.settings)
./activity/activity_CopyGlencoeStillImages.py:        storage_context = StorageContext(self.settings)
./activity/activity_ConvertImagesToJPG.py:            storage_context = StorageContext(self.settings)
./activity/activity_ExpandArticle.py:        storage_context = StorageContext(self.settings)
./activity/activity_VerifyImageServer.py:            storage_context = StorageContext(self.settings)
./activity/activity_SetEIFPublish.py:            storage_context = StorageContext(self.settings)
./activity/activity_DepositAssets.py:            storage_context = StorageContext(self.settings)
./activity/activity_DepositIngestAssets.py:            storage_context = StorageContext(self.settings)
./provider/storage_provider.py:def StorageContext(*args):
./provider/storage_provider.py:# class StorageContext(object):
./provider/article.py:        storage_context = StorageContext(settings)
./provider/image_conversion.py:            storage_context = StorageContext(settings)
```
It is a bit more general so if someone starts passing more than one argument in the future, it should error out rather than silently ignoring the additional arguments.

giorgiosironi: The reason for the duplication is Pylint (run through `lint.sh`) was detecting a problem.
The class `StorageContext` (now commented out) was returning a `S3StorageContext` object when instantiated, which was not even a subclass of itself but a fully different class.
Later on it seems `StorageContext` was turned from a class into a function, which solves this problem. However the conventions still require function names to_be_in_snake_case so `StorageContext` and `storage_context` should be unified to `storage_context` only.
Would do that as a follow-up PR, not to create confusion in this one.

tests/activity/classes_mock.py
giorgiosironi: `classes_mock` is used only in `tests`, which should be used only in `ci` by the `project_tests.sh` script running all the unit tests it can find. 
After running all tests in the Vagrant VM, `filelike` was always a `file`, so I'll remove the `if` to make it fail more clearly if it is not.

provider/storage_provider.py
gnott: Yeah cool, if it is refactored in the future to have just one method to instantiate that sounds great!

gnott:
:+1:

provider/storage_provider.py
giorgiosironi: Follow up in https://github.com/elifesciences/elife-bot/pull/519

PR #507, opened 2017-10-02
Feature crossref xml generation

gnott:
In summary, this branch changes the DepositCrossref activity to use the new elife-crossref-xml-generation library.

Changes to DepositCrossref include
* no longer needs to load the older elife-poa-xml-generation code from an exteral folder
* in ``requirements.txt`` for the Python requirements, it now includes the elifecrossref library at the elifearticle library on which it depends. A sepecific commit of elifetools is used because it has some new features required by elifecrossref that are not part of a new numbered release of the library yet.
* The bot ``settings-example.py`` has two new values for setting the elifecrossref configuration values: ``elifecrossref_config_file`` and ``elifecrossref_config``, which is the config section to use from the config file
* DepositCrossref, renamed one of the activity directories to INPUT_DIR which is a more logical name, and how these directory names are specified, and the TMP_DIR in the elife-crossref-xml-generation library is reset
* some older (I presume unused) code that would attempt to set the article version from the XML file name is changed: since elife-crossref-xml-generation will already set the version number from the file name, DepositCrossref is changed so if the article does not have a version value for some reason then it will get the most recent version number from Lax. The version number is important in Crossref deposits because the URLs to the XML and PDF files for text and data mining are dependent on knowing the article version number.
* in ScheduleCrossref, it is changed to use the original article XML file name when adding the file to the S3 outbox, since the filename includes the version number, which we want now.
* Automated tests are improved and expanded: more scenarios for DepositCrossref with combinations of multiple files, bad files, etc. Removed an older test case for DepositCrossref. ScheduleCrossref is changed to reflect the new file name copied to the S3 outbox
* A change to SilentCorrectionsProcess workflow, where CrossrefDeposit will no longer be "scheduled" as part of silent corrections

When time allows if you could please review and comments are welcome @giorgiosironi 

This is not ready to merge just yet, awaiting a final approval from Melissa and production when we're ready to deploy the new Crossref deposit format. Aiming for in the next week or so or less.


activity/activity_DepositCrossref.py
```diff
@@ -17,6 +16,11 @@
 import provider.simpleDB as dblib
 import provider.article as articlelib
 import provider.s3lib as s3lib
+import provider.lax_provider as lax_provider
```
giorgiosironi: possibly `from provider import lax_provider` is equivalent and in line with the other imports that follow?

requirements.txt
```diff
@@ -10,7 +10,9 @@ requests==2.5.3
 wsgiref==0.1.2
 lxml==3.4.1
 xlrd==0.9.3
-elifetools==0.1.7
+git+https://github.com/elifesciences/elife-tools.git@dacd55fb18885d0462b2266cdd3ba1b11a003aca#egg=elifetools
+git+https://github.com/elifesciences/elife-article.git@e86fd697880673ded634c8eae0624227160f05ce#egg=elifearticle
+git+https://github.com/elifesciences/elife-crossref-xml-generation.git@3b12d754e21fb2954ad2d63cca16ad5ee823b0ff#egg=elifecrossref
```
giorgiosironi: perfect pinning for a reproducible build

settings-example.py
```diff
@@ -118,6 +118,10 @@ class exp():
     # Templates S3 settings
     templates_bucket = 'elife-bot-dev'
 
+    # Crossref generation
+    elifecrossref_config_file = 'crossref.cfg'
+    elifecrossref_config = 'elife'
```
giorgiosironi: perhaps `elifecrossref_config_section` would make clear to casual readers where this value is used?

tests/activity/test_activity_deposit_crossref.py
```diff
@@ -45,36 +46,95 @@ def fake_download_files_from_s3_outbox(self, document):
...
+        {
+            "article_xml_filenames": ['elife-18753-v1.xml'],
+            "deposit_files_return_value": False,
+            "expected_result": True,
+            "expected_approve_status": True,
+            "expected_generate_status": True,
+            "expected_publish_status": False,
+            "expected_activity_status": False,
+            "expected_file_count": 1,
+        },
```
giorgiosironi: good and expandable data set

tests/activity/test_activity_deposit_crossref.py
```diff
@@ -45,36 +46,95 @@ def fake_download_files_from_s3_outbox(self, document):
...
+        file_count = len(os.listdir(self.tmp_dir()))
+        self.assertEqual(file_count, test_data.get("expected_file_count"))
+        if file_count > 0 and test_data.get("expected_crossref_xml_contains"):
+            # Open the first crossref XML and check some of its contents
+            crossref_xml_filename_path = os.path.join(self.tmp_dir(), os.listdir(self.tmp_dir())[0])
+            with open(crossref_xml_filename_path, 'rb') as fp:
+                crossref_xml = fp.read()
+                for expected in test_data.get("expected_crossref_xml_contains"):
+                    try:
+                        self.assertTrue(expected in crossref_xml)
```
giorgiosironi: `self.assertTrue()` should accept a second argument with an error message, where `"%s not found in crossref_xml` % expected` could be put

giorgiosironi: to avoid the additional `try/except` block

tests/activity/test_activity_deposit_crossref.py
```diff
@@ -45,36 +46,95 @@ def fake_download_files_from_s3_outbox(self, document):
...
+                        raise
+
+
+    @patch('provider.lax_provider.article_versions')
+    def test_parse_article_xml(self, mock_lax_provider_article_versions):
+        "example where there is not pub date and no version for an article"
+        mock_lax_provider_article_versions.return_value = 200, test_case_data.lax_article_versions_response_data
+        articles = self.activity.parse_article_xml(['tests/test_data/crossref/elife_poa_e03977.xml'])
+        article = articles[0]
+        self.assertIsNotNone(article.get_date('pub'))
```
giorgiosironi: also here a second argument with an error message could be helpful

activity/activity_DepositCrossref.py
gnott: Yes, I thought pylint complained about this but your suggestion is good, thanks!

settings-example.py
gnott: I like it, I will rename it.

tests/activity/test_activity_deposit_crossref.py
gnott: Making use of the msg argument I like :+1:

tests/activity/test_activity_deposit_crossref.py
gnott: As above, I will add a more descriptive msg for if the test case does not pass.

gnott:
Thanks for the input @giorgiosironi, it seems there is nothing stopping the merge now, here goes!

PR #510, opened 2017-10-06
Padding id value returned by Lax in the lax_response_adapter.py with …

gnott:
…a test case.

Uses utils.pad_msid() to "zerofill" the id into article_id, and a test case based on the quirky silent correction that failed to complete the PostPerfectPublication workflow, article 00269.


gnott:
Thanks, I will get this merged in now.

PR #513, opened 2017-10-18
In install.sh uninstall elifearticle and elifecrossref when installing.

gnott:
I tried deploying a bug fix to elifecrossref. After deploying, the bug is still there. I believe the library did not update to the commit as specified in ``requirements.txt``. This addition to ``install.sh`` should follow the similar pattern to what is done with elifetools: to uninstall it first, then it will result in installing the desired commit of each library.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -11,5 +11,11 @@ source venv/bin/activate
 if pip list | grep elifetools; then
     pip uninstall -y elifetools
 fi
+if pip list | grep elifearticle; then
+    pip uninstall -y elifearticle
+fi
+if pip list | grep elifecrossref; then
+    pip uninstall -y elifecrossref
+fi
 pip install -r requirements.txt
 


giorgiosironi:
Here's one for @seanwiseman why does this happen (and hence we are forced to go with the brute force solution uninstall+install)?

gnott:
I think if we did ``pip install --upgrade elifecrossref`` then we might get the right commit we are looking for. I'm not sure if the ``requirements.txt`` method of install makes that happen.

I may try that command or manually uninstall reinstall that library today because I want to continue with delivering to Crossref.

gnott:
Tested a working combination now doing it manually

1. pip uninstall -y elifecrossref
2. pip install -r requirements.txt

Then, it still wasn't working until I restarted all of the bot processes.

giorgiosironi:
Go ahead with this change if it solves the problem for now, we'll look into how to do this without uninstalling with @seanwiseman.
It is correct that you have to restart the processes, as if they have already loaded the Python modules from disk the fastest way they have to see the new code is to restart from a blank slate.

PR #519, opened 2017-10-23
StorageContext factory cleanup

giorgiosironi:
We started switching from
- `StorageContext()`, a function named like a class created for backward compatibility
to
- `storage_context()`, a function properly named and more general

Since you never know if you have found all usages, I inserted a deprecated log file to be written if the old function is still called.


giorgiosironi:
Indeed, consolidating on a single provider makes it easy to mock it in tests. and shows the dependency of an activity towards S3 very clearly.

PR #522, opened 2017-10-31
Refactor use of the get_article_xml_key() to use the function from th…

gnott:
…e provider.

JIRA ticket https://elifesciences.atlassian.net/browse/ELPP-3177  refactoring to not use the staticmethod from ConvertJATS we need to do before removing the ConvertJATS activity.


gnott:
Yes, I thought there was a test (odd I didn't add one earlier). I will add one!

gnott:
I think there are tests already, https://github.com/elifesciences/elife-bot/blob/feature-refactor-get_article_xml_key/tests/provider/test_article_structure.py#L232, I seemed to remember adding the test (if it was me).

``article_structure.py`` has 100% coverage too!

PR #524, opened 2017-11-01
Deliver files to CNPIEC daily at 23:30 UTC.

gnott:


giorgiosironi:
I often thought on how we can make this more configurable, but it's definitely easier to understand when this is kept into the Python code than to add and remove manual crons from the formula. Perhaps in the future a cron.cfg file could be useful if we want to make this generic.

gnott:
I agree, the cron functions could be formalised with what workflow to start, data to process, and the time of day to process it. The list of cron jobs you'd like to run can be extracted to a new file, leaving the cron functionality general purpose. This of course is just to get CNPIEC files hooked up. We can review the cron strategy in a separate initiative.

gnott:
Added a ticket describing cron improvements we can do in https://elifesciences.atlassian.net/browse/ELPP-3252.

PR #525, opened 2017-11-06
Feature dashboard properties

gnott:
Started by pulling in the new activity code provided by @jhroot, thanks John!

A first attempt at refactoring ConvertJATS I went a little too far. It turned out we still need to keep the EIF format active while the old site is still generating eTOC content. Also, there are more steps to remove the EIF support entirely, involving multiple activities, ``shimmy.py``, updating data in test scenarios, etc.

Here, it is 
* introducing the new activity SendDashboardProperties
* removing the functions that were sending properties to the dashboard from ConvertJATS activity
* insert SendDashboardProperties into the appropriate workflows, following ExpandArticle activity


workflow/workflow_IngestArticleZip.py
```diff
@@ -68,6 +68,17 @@ def __init__(self, settings, logger, conn=None, token=None, decision=None,
                         "start_to_close_timeout": 60 * 15
                     },
                     {
+                        "activity_type": "SendDashboardProperties",
```
giorgiosironi: it seems like the right place, since the extracted XML is available to pull the values of the properties, and doing it as early as possible helps providing the dashboard with debugging information in case later activities fail.

tests/activity/test_activity_send_dashboard_properties.py
```diff
@@ -0,0 +1,55 @@
...
+        fake_emit_monitor_property.assert_any_call(ANY, '00353', 'status', u'VOR', 'text', version='1')
+        fake_emit_monitor_property.assert_any_call(ANY, '00353', 'publication-date', u'2012-12-13', 'text', version='1')
+        fake_emit_monitor_property.assert_any_call(ANY, '00353', 'article-type', u'discussion', 'text', version='1')
+        fake_emit_monitor_property.assert_any_call(ANY, '00353', 'corresponding-authors', u'Eve Marder', 'text', version='1')
+        fake_emit_monitor_property.assert_any_call(ANY, '00353', 'authors', u'Eve Marder', 'text', version='1')
+
+        fake_emit_monitor_event.assert_any_call(ANY, '00353', '1', '1ee54f9a-cb28-4c8e-8232-4b317cf4beda',
+                                                'Send dashboard properties', 'start',
+                                                'Starting send of article properties to dashboard for article 00353')
+        fake_emit_monitor_event.assert_any_call(ANY, '00353', '1', '1ee54f9a-cb28-4c8e-8232-4b317cf4beda',
```
giorgiosironi: Mock-heavy but with good coverage

activity/activity_SendDashboardProperties.py
```diff
@@ -0,0 +1,133 @@
...
+
+        title = utils.tidy_whitespace(parser.full_title(soup))
+        self.set_monitor_property(self.settings, article_id, "title", title,
+                                  "text", version=version)
+
+        status = utils.article_status(parser.is_poa(soup))
+        self.set_monitor_property(self.settings, article_id, "status", status,
+                                  "text", version=version)
+
+        pub_date = datetime.fromtimestamp(time.mktime(parser.pub_date(soup))).strftime("%Y-%m-%d")
```
giorgiosironi: this magic generation line is worth a `utils` function or to be put into a `provider`

gnott: I think so, I simplified it and moved created PUB_DATE_FORMAT in ``utils.py`` in commit https://github.com/elifesciences/elife-bot/commit/dbe1762190bb1fc38cb59c19c04e2396a27804f5

PR #530, opened 2017-11-06
Remove old NewS3File workflow and starter.

gnott:
Uncovered in the attempt to remove the ConvertJATS activity, the NewS3File workflow is no longer used and can be deleted, as well as its starter.


giorgiosironi:
A quick grep shows also this commented out code that could be deleted:
https://github.com/elifesciences/elife-bot/blob/develop/test_publish_message.py#L10-L17

gnott:
I discussed with @jhroot about the ``test_publish_message.py`` and I planned to leave it unchanged, if that is ok as part of this PR.

giorgiosironi:
No problem for me, it's commented code

PR #531, opened 2017-11-07
Consider non-versioned XML file as an ArticleXML to fix …

gnott:
… SendDashboardProperties.

end2end tests seem to be stopping when reaching SendDashboardProperties. If it looks for the ArticleXML in the expanded bucket before ApplyVersionNumber happens, then it needs to find the non-versioned XML file name and consider that a valid ArticleXML as according to the ``article_structure.py`` provider logic.

I think this doesn't have too many negative consequences.


jhroot:
This change looks good to me

gnott:
Thanks, I will merge and see if it fixes the end2end tests.

PR #533, opened 2017-11-08
Update elifetools version is use, and also elifecrossref and …

gnott:
… elifearticle.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,9 +10,9 @@ requests==2.5.3
 wsgiref==0.1.2
 lxml==3.4.1
 xlrd==0.9.3
-git+https://github.com/elifesciences/elife-tools.git@05def8780aaf7ffb247c409ad77d62d186f1abdd#egg=elifetools
-git+https://github.com/elifesciences/elife-article.git@3c770c1c934d381cc01a7d5d984d13849704c784#egg=elifearticle
-git+https://github.com/elifesciences/elife-crossref-xml-generation.git@288f0bc8d1148eb1795c8ae18a3985d30ba38cd5#egg=elifecrossref
+git+https://github.com/elifesciences/elife-tools.git@7fbcef83fbe4ff960a6f44e1d143588b96c79325#egg=elifetools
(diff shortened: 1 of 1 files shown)


giorgiosironi:
Thinking: can we automate this update like in https://alfred.elifesciences.org/job/dependencies-elife-bot-update-elife-poa-xml-generation/
What's the most common use case?
- update a single one of these 3 projects
- update a subset of them
- update all 3 together
And when should we do it (the automated opening of this PR)?
- when there is a new `master` commit on one of these 3 project
- only when some of them have a new `master` commit, not for all 3
- every day at 10 AM
- every week

gnott:
We're not committing to master branch on any of these projects on a regular basis, not that we could not do that.

elifetools may get updated on its own. elifearticle and elifecrossref are closely tied together, and they would probably get updated at the same time to the latest version if any new commits happened on them.

In this case, I tested the elifecrossref project with the latest elifetools and elifearticle separately so I could be sure the test cases passed. I didn't expect them not to pass since the changes were non-breaking, but you never know.

The commit to master branch might be the best option to automatically update projects and then run tests to be sure it doesn't break anything.

gnott:
I will merge this PR so we can get the elifetools rewrites into production.

giorgiosironi:
No blockers on this PR.

> We're not committing to master branch on any of these projects on a regular basis

Is it because elife-bot follows their `develop` branches instead (if they exist for some of them)? Same reasoning stands, but tracking their `develop` branches rather than `master` if we update from them. But of course if we can merge to `master` on passing tests, it's safer to avoid accidentally importing versions that are red.

> elifetools may get updated on its own.

I'll start from that one and do the same routine that opens PRs like https://github.com/elifesciences/bot-lax-adaptor/pull/350

In that case, the update is triggered when a new `develop` version passes the tests: 
https://github.com/elifesciences/elife-tools/blob/develop/Jenkinsfile#L10-L15
but there is no automated merge in `master`. It's a "bit of both" solution.

PR #534, opened 2017-11-08
SWFmeta provider fix and tests

gnott:
Still a work in progress, there was a bug I fixed in an unused part of the ``swfmeta.py`` provider when I started to feed files by FTP to a downstream recipient many weeks ago. I was depending on the SWF meta info to count open FTPArticle workflows, but the ``all_infos`` variable was not set and caused an error.

So far I've added tests that were in lettuce, and I've simulated a ``nextPageToken`` value and a mock SWF connection. It's a little strange because it has to have a ``nextPageToken`` in the first infos it provides, and then make sure then second infos does not have a ``nextPageToken``, otherwise the pagination will loop forever.

I can describe it in person if time allows and is maybe a good time to refactor this part a tiny bit too.

I also want to add tests for pagination on the ``get_closed_workflow_executionInfos()`` function, and also I think the lettuce tests for this provider can be deleted once these new unittests are introduced.


tests/provider/test_swfmeta.py
```diff
@@ -0,0 +1,150 @@
...
+        self.infos_counter = 0
+
+    def connect(self):
+        pass
+
+    def add_infos(self, infos):
+        "add an infos, to allow it to return more than one infos in succession"
+        self.infos.append(infos)
+
+    def list_open_workflow_executions(self, domain=None, workflow_id=None, workflow_name=None,
```
giorgiosironi: it may not help in this case and in general for a hand-rolled stub like this gives you the most flexibility; in some cases you can use a `mock` and assign an iterable to `side_effect` to make it return different things on different calls https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.side_effect

gnott: ``side_effect`` looks interesting, maybe for simpler tests and not all wrapped up in this mock SWF Connection object. I think this is the only place the swfmeta provider is tested with test fixtures, so it might be ok to use it as it is now, rather than spending more time changing it. It is ugly, but it is passing.

PR #536, opened 2017-11-09
update Session api

jhroot:
including S3 back end and simpler get/set calls

@gnott for discussion

### DIFF
--- a/activity/activity_VerifyImageServer.py
+++ b/activity/activity_VerifyImageServer.py
@@ -1,6 +1,6 @@
 import activity
 import json
-from provider.execution_context import Session
+from provider.execution_context import get_session
 from provider.storage_provider import storage_context
 import provider.article_structure as article_structure
 import provider.iiif as iiif
@@ -37,9 +37,9 @@ class activity_VerifyImageServer(activity.activity):
                 self.logger.info('data: %s' % json.dumps(data, sort_keys=True, indent=4))
 
             run = data['run']
-            session = Session(self.settings)
-            article_id = session.get_value(run, 'article_id')
-            version = session.get_value(run, 'version')
+            session = get_session(self.settings, data, run)
+            article_id = session.get_value('article_id')
+            version = session.get_value('version')
 
         except Exception as e:
             self.logger.exception(str(e))
--- a/activity/activity_VerifyPublishResponse.py
+++ b/activity/activity_VerifyPublishResponse.py
@@ -1,10 +1,10 @@
 import activity
 import json
-from provider.execution_context import get_session
 
 """
 activity_VerifyPublishResponse.py activity
 """
+
 
 class activity_VerifyPublishResponse(activity.activity):
 
--- a/activity/activity_VerifyPublishResponse.py
+++ b/activity/activity_VerifyPublishResponse.py
@@ -1,6 +1,6 @@
 import activity
 import json
-from provider.execution_context import Session
+from provider.execution_context import get_session
 
 """
 activity_VerifyPublishResponse.py activity
--- a/provider/execution_context.py
+++ b/provider/execution_context.py
@@ -1,56 +1,107 @@
 import redis
+from boto.s3.connection import S3Connection
+from boto.s3.key import Key
+from boto.exception import S3ResponseError
 from pydoc import locate
+import json
 
 
-def Session(settings):
+def get_session(settings, input_data, session_key):
     settings_session_class = "RedisSession"  # Default
     if hasattr(settings, 'session_class'):
         settings_session_class = settings.session_class
 
     session_class = locate('provider.execution_context.' + settings_session_class)
-    return session_class(settings)
-
-# class Session(object):
-#     def __new__(self, settings):
-#         settings_session_class = "RedisSession"  # Default
-#         if hasattr(settings, 'session_class'):
-#             settings_session_class = settings.session_class
-#
-#         session_class = locate('provider.execution_context.' + settings_session_class)
-#         return session_class(settings)
+    return session_class(settings, input_data, session_key)
 
 
 class FileSession(object):
 
     # TODO : replace with better implementation - e.g. use Redis/Elasticache
 
-    def __init__(self, settings):
-        self.settings = settings
+    def __init__(self, settings, input_data, session_key):
 
-    def store_value(self, execution_id, key, value):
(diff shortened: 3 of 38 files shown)


provider/execution_context.py
```diff
@@ -1,56 +1,107 @@
...
-    def store_value(self, execution_id, key, value):
-        self.r.hset(execution_id, key, value)
-        self.r.expire(execution_id, self.expire_key)
+    def store_value(self, key, value):
+
+        value = json.dumps(value)
+        self.r.hset(self.session_key, key, value)
+        self.r.expire(self.session_key, self.expire_key)
 
     def get_value(self, execution_id, key):
```
gnott: execution_id here, should it be converted to use self.session_key?

provider/execution_context.py
```diff
@@ -1,56 +1,107 @@
...
+    def get_value(self, key):
+
+        s3_key = Key(self.bucket)
+        s3_key.key = self.get_full_key(key)
+        value = None
+        try:
+            value = s3_key.get_contents_as_string()
+            value = json.loads(value)
+        except S3ResponseError:
+            if key in self.input_data:
```
gnott: Is it possible input_data could be None? Maybe

``if self.input_data and key in self.input_data:``

activity/activity_VerifyPublishResponse.py
```diff
@@ -1,6 +1,6 @@
 import activity
 import json
-from provider.execution_context import Session
+from provider.execution_context import get_session
```
giorgiosironi: can be removed from activities that do not use it?

giorgiosironi: (it seems unused due to the lack of a diff in the rest of the file, but perhaps it is used somewhere in this activity)

activity/activity_VerifyImageServer.py
```diff
@@ -37,9 +37,9 @@ def do_activity(self, data=None):
                 self.logger.info('data: %s' % json.dumps(data, sort_keys=True, indent=4))
 
             run = data['run']
-            session = Session(self.settings)
```
giorgiosironi: This and many other activities look like they weren't passing `data` during the Session object creation, only the `run` when calling `get_value`. While passing the `run` on creation sounds like a good lifecycle (specifying it at the start of the activity and then it's not necessary anymore), `data` means any activity could be overwriting any data in the session, which is much more than what they were doing before. I think it's valuable to retain the original limitation and only passing in `data` in activities that need to overwrite the session, if that's what the argument is doing.

gnott: Perhaps only use the value from data if it doesn't already exist from the session? So no values are overwritten automatically.

provider/execution_context.py
```diff
@@ -38,14 +38,14 @@ def get_value(self, key):
             f = open(self.settings.workflow_context_path + self.get_full_key(self.session_key, key), 'r')
             value = json.loads(f.readline())
         except:
-            if key in self.input_data:
+            if self.input_data is not None and key in self.input_data:
                 value = self.input_data[key]
         return value
 
     @staticmethod
```
gnott: One more comment looking closely - I wonder if this were not a staticmethod, would it still work? Its interface is a little different than the ``get_full_key()`` in the S3 session because it uses the self.session_key instead of getting the session_key in arguments. It might have implications in tests and activities so changing it may have repercussions. @jhroot

activity/activity_VerifyImageServer.py
giorgiosironi: But if the activities didn't pass in `data` before, why they need top pass it in now? It seems that activities that write data use `session.store_value()` so this parameter should never be used?

gnott: I think this is generally so an activity can receive its values from data or the session. For example, LensArticle activity would get it from the session when part of a publication workflow, but when it is run on its own (to regenerate Lens pages) then it would be passed in as data, because no session would yet exist.

giorgiosironi: Ok, now we're talking about an explicit use case and the resulting trade-offs.

Do we need to support every activity potentially being run on its own? Or could we call YAGNI and select activities like `LensArticle` to contain this change? If it's done through a workflow like https://github.com/elifesciences/elife-bot/blob/develop/workflow/workflow_LensArticlePublish.py an additional workflow creation (plus a starter?) is needed anyway to support this standalone execution.

gnott: I think not all activities need to be able to run on their own, some do. We could possibly document which ones must be part of a session workflow, and which can do both, or which support data only. Since a discussion with Sean yesterday, it is a good question whether you can determine where an activity gets the data it requires, and whether it can be compatible to be added to another workflow or not.

giorgiosironi: That seems to push towards not adding a blanket support for the `data` parameter, but rather only for the activities which we want to actually support that use case. It can have a default like `{}`?

jhroot: This was something discussed in detail in the meetings last week. Its purpose is to enable great flexibility about moving activities around as they won't need to care about if the data came via the input data or session. At the moment when we split a workflow or move an actvity from one workflow to another there are a lot of change required that are really incidental to the purpose of the acitivities. 

The support for activities running "on its own" is important but not the reason behind this change.

activity/activity_VerifyPublishResponse.py
jhroot: can be removed from here

activity/activity_VerifyImageServer.py
jhroot: @giorgiosironi  "data means any activity could be overwriting any data in the session, which is much more than what they were doing before"

No, values in the session always take precedence over values in data. The input data is immutable through the processing of the workflow so to allow the unified view of the state to be altered the, and writes are to the session and these take precendence over the input data.

provider/execution_context.py
jhroot: I don't see any issue with this but I will change it for consistency.

jhroot:
updated for comments above

activity/activity_VerifyImageServer.py
giorgiosironi: :+1: on the explicit writes, clarified

PR #552, opened 2017-12-07
Deposit inline figure files as part of IIIF assets.

gnott:
Should fix the issues with trying to include an inline graphic into a JSON image block, where filling in the image details from the IIIF fails since the inf file has not been copied to the bucket yet prior to ingest to Lax. In reference to https://elifesciences.atlassian.net/browse/ELPP-3312.

### DIFF
--- a/tests/provider/test_article_structure.py
+++ b/tests/provider/test_article_structure.py
@@ -160,6 +160,22 @@ class TestArticleStructure(unittest.TestCase):
                     'elife-07398-media1.jpg']
         self.assertListEqual(article_structure.get_figures_for_iiif(files), expected)
 
+    def test_get_inline_figures_for_iiif(self):
+        "should return all inline figure files"
+        files = ['elife-00666-app1-fig1-figsupp1-v1.tif',
+                 'elife-00666-fig2-figsupp2-v1.tif',
+                 'elife-00666-fig2-figsupp2-v1.jpg',
+                 'elife-00666-inf001-v1.jpg',
+                 'elife-00666-inf001-v1-80w.jpg',
+                 'elife-00666-table3-data1-v1.xlsx',
+                 'elife-07702-vor-r4.zip',
+                 'elife-6148691793723703318-fig10-v1.gif',
+                 'elife-9204580859652100230-fig2-data1-v1.xls',
+                 'elife-00666-video2.jpg',
+                 'elife-07398-media1.jpg']
+        expected = ['elife-00666-inf001-v1.jpg']
+        self.assertListEqual(article_structure.get_inline_figures_for_iiif(files), expected)
+
     # see https://github.com/elifesciences/elife-continuum-documentation/blob/master/file-naming/file_naming_spec.md
     def test_get_figures_pdfs(self):
         files = ['elife-07398-media1.jpg',
@@ -216,6 +232,7 @@ class TestArticleStructure(unittest.TestCase):
                  'elife-13273-media1.mp4']
         expected = ['elife-00666-app1-fig1-figsupp1-v1.tif',
                     'elife-00666-fig2-figsupp2-v1.tif',
+                    'elife-00666-inf001-v1.jpg',
                     'elife-00666-video2.jpg',
                     'elife-07398-media1.jpg',
                     'elife-13273-media1.mp4',
--- a/tests/provider/test_article_structure.py
+++ b/tests/provider/test_article_structure.py
@@ -117,22 +117,21 @@ class TestArticleStructure(unittest.TestCase):
     def test_get_original_files(self):
         files = ['elife-00666-fig2-figsupp2-v1.tif',
                  'elife-00666-fig2-figsupp2-v10.tif',
-                 'elife-00666-inf001-v1.jpg',
-                 'elife-00666-inf001-v1-80w.jpg',
+                 'elife-00666-inf001-v1.tif',
                  'elife-00666-table3-data1-v1.xlsx',
                  'elife-07702-vor-r4.zip',
                  'elife-07398-media1.jpg']
         expected = ['elife-00666-fig2-figsupp2-v1.tif',
                     'elife-00666-fig2-figsupp2-v10.tif',
-                    'elife-00666-inf001-v1.jpg',
+                    'elife-00666-inf001-v1.tif',
                     'elife-00666-table3-data1-v1.xlsx']
 
         self.assertListEqual(article_structure.get_original_files(files), expected)
 
     def test_get_media_file_images(self):
         files = ['elife-00666-fig2-figsupp2-v1.tif',
+                 'elife-00666-inf001-v1.tif',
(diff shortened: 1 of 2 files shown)


tests/provider/test_article_structure.py
```diff
@@ -160,6 +160,22 @@ def test_get_figures_for_iiif(self):
...
+                 'elife-00666-fig2-figsupp2-v1.jpg',
+                 'elife-00666-inf001-v1.jpg',
+                 'elife-00666-inf001-v1-80w.jpg',
+                 'elife-00666-table3-data1-v1.xlsx',
+                 'elife-07702-vor-r4.zip',
+                 'elife-6148691793723703318-fig10-v1.gif',
+                 'elife-9204580859652100230-fig2-data1-v1.xls',
+                 'elife-00666-video2.jpg',
+                 'elife-07398-media1.jpg']
+        expected = ['elife-00666-inf001-v1.jpg']
```
giorgiosironi: as far as I can tell, `jpg` versions of `inf*` images were already deposited pre-ingest. What we need however, is to also deposit their `.tif` as that is needed for IIIF to work. `elife-00666-inf001-v1.tif` should be added to `files` and `expected` in this test.

(detail: IIIF is able to fall back from .tif to .jpg as the source image, but only if the .tif causes errors during the conversion, not if it's absent and giving a 404)

tests/provider/test_article_structure.py
```diff
@@ -216,6 +232,7 @@ def test_pre_ingest_assets(self):
                  'elife-13273-media1.mp4']
         expected = ['elife-00666-app1-fig1-figsupp1-v1.tif',
                     'elife-00666-fig2-figsupp2-v1.tif',
+                    'elife-00666-inf001-v1.jpg',
```
giorgiosironi: same here, add the `.tif` for clarity

gnott:
Very perceptive comments @giorgiosironi thanks, I change it so inline graphics for IIIF are also only the .tif files.

giorgiosironi:
I don't know where the `inf` `jpg` came from when we saw them, but the intention should be to have both `tif` and `jpg` for each of them so that IIIF can fall back from the former to the latter if needed.

If it's not clear (to me it's not) where the `inf` `jpg` were coming from, they should be included as well in this list and its tests (I'd rather risk the `inf` `jpg` being uploaded twice that not uploaded at all).

gnott:
I can review the test data against what would normally be produced. Would we also want figure JPG files deposited as a fall back for IIIF? All the images will have been converted from tif to jpg by the point DepositIngestAssets activity is run.

giorgiosironi:
> Would we also want figure JPG files deposited as a fall back for IIIF? 

Yes, for all figures

> All the images will have been converted from tif to jpg by the point DepositIngestAssets activity is run.

I guess it's another activity then that produces the `.jpg`, if that can be identified this activity should only deal with the `.tif`. The important end result is that when Lax ingests the XML, all `tif` and `jpg` versions are there in the `*-elife-published` bucket.

gnott:
I cleaned up some example data in the tests cases for the ``article_structure.py`` provider.

I'm not sure how to express these better right now. Following is a summary of my understanding of how the activities proceed in the IngestArticleZip workflow (with respect to these test cases):

1. ``ExpandArticle`` activity, unzips all the ``.tif`` files from the zip into the expanded bucket
2. ``ConvertImagesToJPG`` converts the ``.tif`` files to ``.jpg`` files, saving the ``.jpg`` files into the CDN bucket. We only get one size of jpg per tif file, so I removed the test data that refers to jpg files with alternate dimensions we used to resize into multiple sizes
3. ``DepositIngestAssets`` copies files from the expanded bucket to the CDN bucket. The expanded bucket only has ``.tif`` files and video original files (``.avi``, `.mpg``, etc.) and does not contain any video still files (which would be ``.jpg``). So, the data for testing ``get_figures_for_iiif()`` and ``get_inline_figures_for_iiif()`` would only ever see ``.tif`` files or video files. Hence, I removed all the ``.jpg`` files from the test data here - they would never exist (for eLife articles - a separate point is in the future if non-tif files are part of a zip package, we'd want to relax the rules, which is a valid point, but we should probably create a new test for non-eLife data to be clear? Sorry if this is confusing, I wanted to comment on this situation).
4. ``CopyGlencoeStillImages`` activity happens after, which will add video still ``.jpg`` files directly to the CDN bucket. These files are never added to the expanded bucket, which is why I removed them from the test data for ``get_figures_for_iiif()`` and ``get_inline_figures_for_iiif()``.
5. Then, ``VerifyImageServer`` activity is run, where it should find all the ``.tif`` and ``.jpg`` files in the CDN bucket!
6. Finally, ``IngestToLax`` activity is run, and Lax should also find all the image assets available from the IIIF server, because all the image assets are in the CDN bucket.

@giorgiosironi I hope this helps for you to comment on whether this is sufficient to copy inf tif files over to where they should be, and maybe if you propose some changes to the logic of the above steps we could tackle it in a new PR after planning it out?

giorgiosironi:
Am I right in thinking that if we do a silent correction to fix one of the existing articles that was having the ELPP-3312 problem, that will also populate the missing `.tif`?

gnott:
Good point about ``.tif`` files will not be present in the CDN bucket for the articles. Rather than causing more noise by silent corrections, there are 5 or 6 articles I identified that will have an ``inline-graphic`` converted to an ``image`` block, and for those I will manually copy the ``-inf`` files from the existing expanded bucket over to the CDN bucket. That should hopefully uncover any articles where the images are still missing, in cases I wasn't able to identify specifically.

gnott:
I checked the four articles I was able to find where the ``boxed-text`` ``inline-graphic`` will get converted to an image block, and the ``.tif`` and ``.jpg`` files on the CDN are the following:

```
https://cdn.elifesciences.org/articles/00351/elife-00351-inf1-v1.tif
https://cdn.elifesciences.org/articles/00351/elife-00351-inf1-v1.jpg

https://cdn.elifesciences.org/articles/00646/elife-00646-inf1-v1.tif
https://cdn.elifesciences.org/articles/00646/elife-00646-inf1-v1.jpg

https://cdn.elifesciences.org/articles/02658/elife-02658-inf1-v1.tif
https://cdn.elifesciences.org/articles/02658/elife-02658-inf1-v1.jpg

https://cdn.elifesciences.org/articles/07364/elife-07364-inf1-v1.tif
https://cdn.elifesciences.org/articles/07364/elife-07364-inf1-v1.jpg
```

I think it could be just an end2end or future article ingest issue this new code will solve. At the time the workflows expect to find a ``.tif``, it is not present in the CDN bucket (with the currently running code). Later on, all the ``inf`` ``.tif`` files get copied over the the CDN bucket, but after the IIIF server verification happens. So, for all the previously published articles, the ``.tif`` files should already be available for IIIF to access if a Lax backfill happens.

I'll merge this code now and hopefully end2end tests will pass once the new elife-tools code is also integrated.

PR #553, opened 2017-12-08
Work in progress for comment - Simple cut contact with elife-website.

gnott:
The first commit here may be a simplistic start, but hopefully it is enough to illustrate an idea and attract some comments please.

The motivation is: if elife-website is turned off, to have no adverse effects on the bot, and to allow publication of articles to happen normally.

In looking at the ``PreparePostEIF`` activity, ``shimmy.py``, ``ArticleInformationSupplier`` workflow, and ``PostEIFBridge`` activity and how they interconnect, it could be simply to substitute ``PreparePostEIF`` activity with the ``PostEIFBridge`` activity.

``PostEIFBridge`` seems to load variables from the data passed to it. The data doesn't include too many items that elife-website would have provided, other than an ``updated`` date (which is probably not used, or taken from Lax), and a ``published`` value, where we prefer the value Lax gives the bot.

Open for comment!

It would also be interesting to see if this variation were run on and end2end workflow whether everything would work as expected. That could be a quick way to determine the blind spots.

Ideally something like this needs to be merged and deployed before elife-website is turned off.

### DIFF
--- a/workflow/workflow_ApproveArticlePublication.py
+++ b/workflow/workflow_ApproveArticlePublication.py
@@ -56,19 +56,6 @@ class workflow_ApproveArticlePublication(workflow.workflow):
                         "schedule_to_start_timeout": 300,
                         "start_to_close_timeout": 60 * 5
                     },
-                    {
-                        "activity_type": "ApprovePublication",
-                        "activity_id": "ApprovePublication",
-                        "version": "1",
-                        "input": data,
-                        "control": None,
-                        "heartbeat_timeout": 60 * 5,
-                        "schedule_to_close_timeout": 60 * 5,
-                        "schedule_to_start_timeout": 300,
-                        "start_to_close_timeout": 60 * 5
-                    },
-
-
                 ],
 
             "finish":
--- a/workflow/workflow_ProcessArticleZip.py
+++ b/workflow/workflow_ProcessArticleZip.py
@@ -112,15 +112,15 @@ class workflow_ProcessArticleZip(workflow.workflow):
                         "start_to_close_timeout": 60 * 5
                     },
                     {
-                        "activity_type": "PreparePostEIF",
-                        "activity_id": "PreparePostEIF",
+                        "activity_type": "PostEIFBridge",
+                        "activity_id": "PostEIFBridge",
                         "version": "1",
                         "input": data,
                         "control": None,
-                        "heartbeat_timeout": 60 * 5,
-                        "schedule_to_close_timeout": 60 * 5,
+                        "heartbeat_timeout": 300,
+                        "schedule_to_close_timeout": 300,
                         "schedule_to_start_timeout": 300,
-                        "start_to_close_timeout": 60 * 5
+                        "start_to_close_timeout": 300
                     },
 
                 ],
--- a/workflow/workflow_SilentCorrectionsProcess.py
+++ b/workflow/workflow_SilentCorrectionsProcess.py
@@ -112,15 +112,15 @@ class workflow_SilentCorrectionsProcess(workflow.workflow):
                         "start_to_close_timeout": 60 * 5
                     },
                     {
-                        "activity_type": "PreparePostEIF",
-                        "activity_id": "PreparePostEIF",
+                        "activity_type": "PostEIFBridge",
+                        "activity_id": "PostEIFBridge",
                         "version": "1",
                         "input": data,
                         "control": None,
-                        "heartbeat_timeout": 60 * 5,
-                        "schedule_to_close_timeout": 60 * 5,
+                        "heartbeat_timeout": 300,
+                        "schedule_to_close_timeout": 300,
                         "schedule_to_start_timeout": 300,
-                        "start_to_close_timeout": 60 * 5
+                        "start_to_close_timeout": 300
                     },
                     {
                         "activity_type": "PublishToLax",


workflow/workflow_ApproveArticlePublication.py
```diff
@@ -56,19 +56,6 @@ def __init__(self, settings, logger, conn=None, token=None, decision=None,
                         "schedule_to_start_timeout": 300,
                         "start_to_close_timeout": 60 * 5
                     },
-                    {
```
giorgiosironi: note: in this workflow, `PublishToLax` covers publication on 2.0 while `ApprovePublication` covers publication on 1.0 which is what we want to skip

workflow/workflow_ProcessArticleZip.py
```diff
@@ -112,15 +112,15 @@ def __init__(self, settings, logger, conn=None, token=None, decision=None,
                         "start_to_close_timeout": 60 * 5
                     },
                     {
-                        "activity_type": "PreparePostEIF",
-                        "activity_id": "PreparePostEIF",
+                        "activity_type": "PostEIFBridge",
```
giorgiosironi: the job of this activity seems to collect some data and start the `PostPerfectPublication` workflow. However we have 2 `PostPerfectPublication` workflows running concurrently as far as I can tell:

- one run after `shimmy`
- one run after `lax_response_adapter`

The one coming from `shimmy` and this activity usually exits because the publishing authority is now `journal`: 
https://github.com/elifesciences/elife-bot/blob/develop/activity/activity_VerifyPublishResponse.py#L45-L64
Do we need to run this activity at all or could we just drop `PreparePostEIF` and end the workflow?

giorgiosironi:
Check my related https://github.com/elifesciences/elife-spectrum/pull/73 which will be a dependency of this.

workflow/workflow_ProcessArticleZip.py
gnott: In the finer details my interpretation of these workflows may be inaccurate, though what I was seeing (and I think I still see) is:

``PostEIFBridge`` activity is run after ``shimmy.py`` gets a response from elife-website (as part of the ``ArticleInformationSupplier`` workflow). 

``PostEIFBridge`` is important to run because it is depositing the data into the dashboard (which is base64 encoded) to say, if someone clicks the "Publish" button on the dashboard, then run the ``PostPerfectPublication`` workflow with the session data you stored, that was given by the ``PostEIFBridge`` activity.

In ``PostEIFBridge``, if the session attribute ``published`` is True, then it will not bother waiting for someone to click buttons on the dashboard and instead it will immediately continue with the ``PostPerfectPublication`` workflow. I think this only happens when a silent correction workflow happens, and in that case ``published`` is set to True in the session as part of the silent correction logic.

It is a bit strange, looking at it, because I would thinking clicking the button on the dashboard should actually start a ``ApproveArticlePublication`` workflow. That would in turn run the ``PublishToLax`` activity, and then the ``lax_response_adapter`` would start a ``PostPerfectPublication`` workflow.

To circle back to my earlier point, I believe the ``PreparePostEIF`` activity needs to send some data to the dashboard to make clicking the "Publish" button there to start an appropriate workflow for the right article. That is the bit we need to keep in some form.

Please let me know if this helps you run any end2end tests on these and hopefully we can track down what part of this rewiring is incorrect.

giorgiosironi: The flow makes sense to me, and the most conservative thing seems the current proposal of skipping `shimmy` (which represents the interaction with `elife-website`) and continuing to run whatever was being run after shimmy before (which is just the `PostEIFBridge` activity).

gnott:
The elife-spectrum checks should not conflict with this now, correct @giorgiosironi?

Is there a way to test this before merging it into the develop branch, or should we just merge it and see if the tests pass?

giorgiosironi:
No easy way to test it outside of end2end AFAIK.

giorgiosironi:
elife-spectrum has been updated with https://github.com/elifesciences/elife-spectrum/pull/73 not to depend on 1.0

giorgiosironi:
We have the go ahead from @sianroderick so will merge as soon as `test-elife-bot` is green again.

giorgiosironi:
Reverted after end2end failure

workflow/workflow_ProcessArticleZip.py
Jenniferstrej: Sorry for not seeing this before. On the continuum version I eliminate EIF publication completely. If that's what we want for elife-bot now, we need to extract the code bit that @gnott mentioned that is on PostEIFBridge that "makes the Publish button appear in the dashboard" and add it to the last activity we want to run or have another activity just for that. The bit of code is:

https://github.com/elifesciences/elife-bot/blob/master/activity/activity_PostEIFBridge.py#L92-L93

I extracted into a method like this (I added it to the last workflow, but probably an activity with it would be a good call for separation):

https://github.com/digirati-co-uk/elife-bot/blob/master/activity/activity_DepositAssets.py#L115-L136

Regarding Silent Corrections, `PostEIFBridge`  only supports the old website silent corrections. So if we extract that part of the code that sends the right property to the dashboard we can remove that activity completely.

Silent corrections on the new website happen by sending a `force:true` to Lax on PublishToLax activity.

Jenniferstrej:
I would personally extract the code that is sending the property to the dashboard and eliminate the PostEIFBridge completely. 

The only complexity is to get `eif_location` that is in the session I think. It looks like it's currently added to the session in ConvertJATS (so I guess we should extract that from there too - it's the only information that's important for the new site that is being extracted on old site only code) the https://github.com/elifesciences/elife-bot/blob/master/activity/activity_ConvertJATS.py#L91

Please let me know if my explanation is not clear or if you think I am confusing things. Not sure what the aim is here exactly.

giorgiosironi:
The aim is a minimal change to remove the dependency on the 1.0 website, which will be turned off next week. It will be followed up by removal of EIF, but it's not strictly necessary to remove it now.

I am continuing with https://github.com/elifesciences/elife-bot/pull/596 but as you have guessed `eif_location` is a problem. I'll try to read that from the session, since `ConvertJATS` is still being executed.

PR #554, opened 2017-12-11
Updated elifetools: pinned elifetools in requirements.txt to 6a83ccdeae75f4575769375636cd4b240850369e

elife-alfred-user:
I have run https://alfred.elifesciences.org/job/dependencies-elife-bot-update-elife-tools/20/display/redirect which resulted in this pull request.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,7 @@ requests==2.5.3
 wsgiref==0.1.2
 lxml==3.4.1
 xlrd==0.9.3
-git+https://github.com/elifesciences/elife-tools.git@5eecbc2d94905b028b4ae339b7a7475bf1a71bf4#egg=elifetools
+git+https://github.com/elifesciences/elife-tools.git@6a83ccdeae75f4575769375636cd4b240850369e#egg=elifetools
 git+https://github.com/elifesciences/elife-article.git@2b4e261579a8905ed02181bd96ccbcbb55850606#egg=elifearticle
 git+https://github.com/elifesciences/elife-crossref-xml-generation.git@c0af3ba85b244d620dd189175913e964a9e5772a#egg=elifecrossref
 PyYAML==3.11


giorgiosironi:
@gnott @seanwiseman the latest versions of `elifetools` seem to break the build of elife-bot

seanwiseman:
I am currently unable to successfully replicate the failure locally (due to local setup issues).

@gnott I can run the unit tests fine but I am not having much luck running the full `feature` test suite locally (where the failures seem to be). Most succeed but I get errors relating to importing settings:
```
boto_steps.py", line 16, in import_settings_module
    "Settings module was imported"
AssertionError: Settings module was imported
```
As a side note is there any other setup steps to go through apart from the `install.sh` and running the tests to get them working locally? if not I will just continue investigating, but thought I better ask in case there is something I am missing.

giorgiosironi:
I think copying `settings-example.py` to `settings.py` (not sure if there are actual values to be filled in).
That README is probably out of date. The build runs on `elife-bot--ci` where a `setting.py` file is added by Salt. It's probably easier to replicate the Travis build because it's only a subset of the full build, since it doesn't have a settings.py file and also fails (possibly for the same underlying cause).

gnott:
I guess there are still some lettuce tests not covered by other tests (which I thought could be true, I just haven't seen confirmation of that yet).

The issue seems to be in ``article`` provider ``parse_article_xml()`` function (because lettuce ``features/009_provider_article.feature`` scenario 3 is failing, which parses an XML file using the article provider).

When I comment out the try catch statements, I get this error

```
    FeatureNotFound: Couldn't find a tree builder with the features you requested: lxml-xml. Do you need to install a parser library?
```

I think the BeautifulSoup and/or lxml version in elife-bot ``requirements.txt`` will need be to upgraded to be compatible with the latest ``elife-tools`` code.

I will attempt this on a new branch and see if the tests will pass.

gnott:
PR #555 fixed this I think, and I will close this PR.

PR #558, opened 2017-12-13
Tests for PubmedArticleDeposit activity, prior to refactoring it.

gnott:
If anyone wants to take a look, this is just a basic test so there are some tests before I set about introducing the the PubMed code library and refactoring the activity. The test should still be compatible with the code as it is now and can be merged before I start refactoring.

I based this on the tests in https://github.com/elifesciences/elife-bot/blob/develop/tests/activity/test_activity_deposit_crossref.py which are fairly similar functionally.


seanwiseman:
Looks good.

Could the large test case could be broken down into smaller test cases? each focusing on a few assertions? I suspect the setup and a large amount of mocking required would make that less feasible.

Are there any failure scenarios that could be tested?

tests/activity/test_activity_pubmed_article_deposit.py
```diff
@@ -0,0 +1,134 @@
...
+    @patch.object(SimpleDB, 'elife_add_email_to_email_queue')
+    @patch.object(activity_PubmedArticleDeposit, 'upload_pubmed_xml_to_s3')
+    @patch.object(activity_PubmedArticleDeposit, 'clean_outbox')
+    @patch.object(activity_PubmedArticleDeposit, 'ftp_files_to_endpoint')
+    @patch.object(article, 'get_article_bucket_pub_date')
+    @patch.object(lax_provider, 'was_ever_poa')
+    @patch.object(lax_provider, 'published_considering_poa_status')
+    @patch.object(lax_provider, 'article_highest_version')
+    @patch.object(activity_PubmedArticleDeposit, 'get_outbox_s3_key_names')
+    @patch.object(activity_PubmedArticleDeposit, 'download_files_from_s3_outbox')
```
giorgiosironi: it's a lot of stubbing, but better than having no tests before refactoring :+1:

gnott:
There could be some error states I didn't explore too much here. Most of it will silently ignore files or handle exceptions if errors happen for any individual file processed right now.

I can look at breaking apart some logic into more easily testable parts when I refactor to incorporate the new code library. I didn't want to spend too much time on tests with the current state of things since in the refactoring I might move things around anyway.

seanwiseman:
OK understood. I understand its probably not worth it in this instance.

PR #560, opened 2017-12-13
Allow alpha characters in article_id of file names in …

gnott:
…article_structure.py

Relaxes the ``article_structure.py`` provider when it is extracting the ``article_id`` value to allow alpha as well as numeric characters, expanding the pattern to ``[a-zA-Z0-9]``.

This was the result of a quick test I tried for a non-eLife sample in the cotinuumtest enviroment which allowed a sample file with a non-numeric article id in the filename to get beyond the ``ExpandArticle`` activity.

I think this is compatible with eLife files and possibly wouldn't hurt to include it for making testing easier with non-eLife files.

I added a test case to compare ``ArticleInfo`` object attributes. More example of files can be added now or in the future.

Including @Jenniferstrej to be aware of this and it might be useful in testing other files.

### DIFF
--- a/provider/article_structure.py
+++ b/provider/article_structure.py
@@ -37,7 +37,7 @@ class ArticleInfo(object):
         (self.filename, self.extension) = full_filename.rsplit('.', 1)
         parts = self.filename.split('-')
 
-        match = re.match('.*?-[0-9]+?-(poa|vor)(-*?(v|r)[0-9]+?)?(-([0-9]+))?\.zip', self.full_filename, re.IGNORECASE)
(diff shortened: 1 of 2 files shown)


gnott:
There are probably a few more places where the zerofill values exist, at least this one is the first encountered and we can work our way along.

In ``provider/utils.py`` if we use the ``pad_msid()`` consistently, then it could be made more flexible or customised for other journals.

When refactoring the jats-generator library code, I started using a format that supports string and integers too, e.g.

```
"{msid:0>5}".format(msid=msid)
```

works for integers and strings.

PR #570, opened 2017-12-15
Moves pdf cover (landing) page function to article.py

Jenniferstrej:
Revisited the code after almost a year and it makes more sense to have the article pdf cover link function inside the article object. Added to tests.

This code can be merged without us having the real pdf cover link as it will only be used my the email templates that are in this PR https://github.com/elifesciences/elife-email-templates/pull/2/files


tests/activity/test_activity_publication_email.py
```diff
@@ -325,6 +325,7 @@ def test_template_get_email_body_00353(self, fake_download_email_templates_from_
 
         article_object = article()
         article_object.parse_article_file("tests/test_data/elife-00353-v1.xml")
+        article.pdf_cover_link = article_object.get_pdf_cover_page(article_object.doi_id, self.activity.settings, self.activity.logger)
```
giorgiosironi: can this be made a constant value? If needed, by mocking? The test is focused on the activity so the article_object is an additional piece

Jenniferstrej: I see that my intention was for article.pdf_cover_link to be article_object.pdf_cover_link, I'm changing that. Is that what you mean? I'm note sure I understand?

giorgiosironi: I mean this is a test, so can a constant string be assigned here rather than also re-executing `article_object` code?

Jenniferstrej: I made a change to this line, I am supposed to mutate `article_object` and not `article`
https://github.com/elifesciences/elife-bot/pull/570/commits/dbd8503086d6fc17f71ad375f1ca19cf9e7366f7

Jenniferstrej: on the instantiation of article you mean? have article_object populate the attribute `pdf_cover_link` as we run article_object = article()?

tests/activity/test_activity_publication_email.py
```diff
@@ -325,6 +325,7 @@ def test_template_get_email_body_00353(self, fake_download_email_templates_from_
 
         article_object = article()
         article_object.parse_article_file("tests/test_data/elife-00353-v1.xml")
+        article_object.pdf_cover_link = article_object.get_pdf_cover_page(article_object.doi_id, self.activity.settings, self.activity.logger)
```
giorgiosironi: my proposal: change to
```
article_object.pdf_cover_link = "https://localhost.org/download-your-cover/00353"
```
so that if `article_object.get_pdf_cover_page` breaks, only the tests of `provider.article` fail; this way, this test only fail if there is a problem with the activity code rather than a problem with the provider (which should be caught by the provider's tests, so that we don't have to retest all providers in all activities.)

Jenniferstrej: Ok, I understand. The way it is above I am also testing the `get_pdf_cover_page` function. So in order to have the same coverage, I'd need to add a new test just for that function. I see that if we have them split, we know more quickly where the problem is, with that test testing specifically one thing. I can add that.

Jenniferstrej: I think I will also be losing the article_object.doi verification value when spliting this test.

Jenniferstrej: Made the change @giorgiosironi - https://github.com/elifesciences/elife-bot/pull/570/commits/14680f51240d75436221a64ccc895e544fb3abde

PR #574, opened 2017-12-20
Feature PubMed

gnott:
This is to implement the deployment of the new code in issue https://github.com/elifesciences/elife-pubmed-feed/issues/52.

The main feature is to use the https://github.com/elifesciences/elife-pubmed-xml-generation library when generating PubMed deposits, instead of the old library.

The tests originally merged as part of https://github.com/elifesciences/elife-bot/pull/558 are improved here, and the number mocks were reduced.

There was some duplication in FTP sending code in other activities, so I created an FTP provider which is used in the ``PubmedArticleDeposit`` activity. The tests do not test it very well without actually running the workflows, and I tested this using a dev environment to be sure. FTP could be tested in an end2end test, but I don't think we're ready for that yet, @giorgiosironi?

I converted all the points where the activity contacts S3 to instead use the ``storage_context`` provider.

One thing I added to the ``storage_context`` provider is the ``delete_resource()`` method, something maybe @Jenniferstrej would want to note. I tested this via a real workflow on my dev environment which uncovered some fixes I had to make to the bucket paths.

This is probably not planned to be deployed until the New Year. I wanted to open a PR for this now to allow chances to review it if time allows between now and 2018.


provider/ftp.py
```diff
@@ -0,0 +1,65 @@
...
+"""
+
+"""
+
+class FTP(object):
+
+    def __init__(self, logger=None):
+        self.logger = logger
+
+    def ftp_connect(self, uri, username, password, passive=True):
```
giorgiosironi: I had attempted before to provide a local FTP server on the elife-bot testing. If we keep the URL configurable, as part of the `settings`, we can upload stuff to `localhost` with some fake user and password.

The real problem in the past was triggering the workflow. The starter is called [from the cron now](https://github.com/elifesciences/elife-bot/blob/develop/cron.py#L149-L153), but we need a way to trigger it on demand so that the test can execute:
- ingest
- publish
- trigger pubmed (or other cron-based workflows)
- check.. that the dashboard has reported something? that the file is in the right place on the `elife-bot--end2end` local FTP server using username and password?

https://github.com/elifesciences/econ-tools/blob/master/econtools/econ_article_feeder.py is integrated into elife-spectrum so I can extend that to put an SQS message in some queue to solve the triggering problem?

giorgiosironi: @gnott let me know if there's a way to trigger the workflow described here through SQS. Being end2end tests can only be added after this is merged anyway.

gnott: With a very small tweak, the econ feeder can run a workflow. It is not using its intended purpose, but this works (after my tweak)

```
python econtools/econ_article_feeder.py -p elife-00666-vor-r1.zip -r 1  ct-elife-production-final ct-workflow-starter-queue PubmedArticleDeposit
```

``PubmedArticleDeposit`` workflow starter doesn't require any input data, just the desired environment name. The input data is in the PubMed outbox on S3. The econ feeder is starting a workflow because it finds a file on S3 matching the string ``elife-00666-vor-r1.zip`` but that is basically ignored.

giorgiosironi: :+1: https://elifesciences.atlassian.net/browse/ELPP-3360

gnott: Wonderful to see the passing tests @giorgiosironi thanks!

gnott:
Tests are running and we have approval to deploy the feature when ready. I will merge now and get it deployed.

PR #581, opened 2017-12-29
IngestArticleZip workflow reduce the timeout from 5 hours to 1 hour.

gnott:
A workflow today when encountering an error in ``SendDashboardProperties`` results in retrying the same activity for 5 hours. The logs would fill up slower if it only retried for 1 hour maximum. If an ``IngestArticleZip`` workflow times out due to too much load, it can be re-run.


giorgiosironi:
Tests are broken in `develop`, checking `elife-bot--ci`.

PR #595, opened 2018-01-11
Simple cut contact with elife-website, 2nd attempt

giorgiosironi:
Trying again https://github.com/elifesciences/elife-bot/pull/553/commits

Problem encountered on end2end:

```
2018-01-11T11:19:21Z INFO worker_10673 activityType: PostEIFBridge
2018-01-11T11:19:21Z ERROR worker_10673 Exception after submitting article EIF
  File "/opt/elife-bot/activity/activity_PostEIFBridge.py", line 34, in do_activity
    article_path = data['article_path']
KeyError: 'article_path'
```
Input to the activity/workflow:
```
2018-01-11T11:19:12Z INFO worker_10656 got activity: 
{
    "activityId": "PostEIFBridge", 
    "activityType": {
        "name": "PostEIFBridge", 
        "version": "1"
    }, 
    "input": "{\"status\": \"vor\", \"update_date\": \"2018-01-11T11:18:05Z\", \"run\": \"f10d37bf-4807-4dca-a95d-6f1c3ebe33d0\", \"expanded_folder\": \"1157624208528401239.1/f10d37bf-4807-4dca-a95d-6f1c3ebe33d0\", \"version\": \"1\", \"result\": \"ingested\", \"eif_location\": \"\", \"article_id\": \"1157624208528401239\", \"message\": null, \"requested_action\": \"ingest\"}", 
    "startedEventId": 84, 
    "taskToken": "AAAAKgAAAAIAAAAAAAAAAlGShZjxVfwsDplkbt9csxmyO6oE6Doou8AEMYa8w4HrhnrV1kYwh5bw41iw8NPpdatWmKtNDYXmiVdql3V2QQY3jlJicYhYJDmSWOMDiPGXRpQtTJJXIRDX1gQ/RrVpF/B/DN2mFjIoWfBlMk4+4xAFik+Ld39mNBg3ty9tNkHg0EsNR0gdjRk7gjOB5tC6+C0InV0cz9/rYWNfKLagSMhlG0MEP7GMbH+XAHojlaU3ckfHDphduHMVLCptxhYC3lsG0CodUXUJq0+S1mGc/IKP2aPsjsF23pSUf1fcXgzfpkN4GonFdlPK2+ebz+PUZ6+ZdHK+la22mQacl/GexV4=", 
    "workflowExecution": {
        "runId": "226EEXuLGKK9+iIl5yL9THvtOzN47jgUGa5uCFoBD05do=", 
        "workflowId": "ProcessArticleZip_1157624208528401239"
    }
}
```
Not only `article_path` but also `published` is missing. Both were provided by the shimmy, which we are skipping now https://github.com/elifesciences/elife-bot/blob/develop/shimmy.py#L91-L102

Are there defaults we can use?
- `article_path` is a value similar to `content/4/e822311538705904637v1` so it can probably be changed to `/articles/...` to work as the preview link?
- `published` is a boolean, can we deduce if it's false or true?

Alternatively, we can still go through the shimmy, and make up some values. I'll evaluate that option too.

### DIFF
--- a/activity/activity_PostEIFBridge.py
+++ b/activity/activity_PostEIFBridge.py
@@ -43,7 +43,7 @@ class activity_PostEIFBridge(activity.activity):
             self.emit_monitor_event(self.settings, article_id, version, run, "Post EIF Bridge", "start",
                                 "Starting " + article_id)
 
-            published = data['published']
+            published = session.get_value('published')
 
             # assemble data to start post-publication workflow
             expanded_folder = data['expanded_folder']
--- a/activity/activity_PostEIFBridge.py
+++ b/activity/activity_PostEIFBridge.py
@@ -31,10 +31,11 @@ class activity_PostEIFBridge(activity.activity):
     def do_activity(self, data=None):
         try:
 
-            article_path = data['article_path']
             article_id = data['article_id']
             version = data['version']
             run = data['run']
+            session = get_session(self.settings, data, run)
+            article_path = session.get_value('article_path')
             self.set_monitor_property(self.settings, article_id, 'path',
                                   article_path, 'text', version=version)
 
--- a/activity/activity_VerifyLaxResponse.py
+++ b/activity/activity_VerifyLaxResponse.py
@@ -40,13 +40,18 @@ class activity_VerifyLaxResponse(activity.activity):
         article_id = data['article_id']
         run = data['run']
         version = data['version']
+        force = data['force']
+        session = get_session(self.settings, data, run)
 
         self.emit_monitor_event(self.settings, article_id, version, run, self.pretty_name, "start",
                                 "Starting verification of Lax response " + article_id)
 
         try:
             if data['result'] == "ingested":
-
+                if force is True:
+                    session.store_value('published', True)
+                else:
+                    session.store_value('published', False)
                 self.emit_monitor_event(self.settings, article_id, version, run, self.pretty_name, "end",
                                         " Finished Verification. Lax has responded with result: ingested."
                                         " Article: " + article_id)
--- a/lax_response_adapter.py
+++ b/lax_response_adapter.py
@@ -83,6 +83,7 @@ class LaxResponseAdapter:
             expanded_folder = token['expanded_folder']
             status = token['status']
             eif_location = token['eif_location'] #support for old journal
+            force = token['force']
 
             workflow_data = {
                 "run": run,
@@ -94,7 +95,8 @@ class LaxResponseAdapter:
                 "result": result,
                 "message": response_message,
                 "update_date": date_time,
-                "requested_action": operation
+                "requested_action": operation,
+                "force": force
             }
 
             if operation == "ingest":
--- a/tests/activity/test_activity_post_eif_bridge.py
+++ b/tests/activity/test_activity_post_eif_bridge.py
@@ -8,6 +8,7 @@ from classes_mock import FakeSQSConn
 from classes_mock import FakeSQSMessage
 from classes_mock import FakeSQSQueue
 from classes_mock import FakeLogger
+from classes_mock import FakeSession
 from testfixtures import TempDirectory
 import json
 import base64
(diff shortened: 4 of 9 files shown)


giorgiosironi:
Experimented with the shimmy, but don't know how to give a value to `published` there:
```
+        if True:
 
-            update_date = self.extract_update_date(passthrough, response.json())
             ingest_publish = response.json().get('publish')
             workflow_data = {
                 'eif_location': filename,
@@ -97,7 +89,7 @@ class Shimmy:
                 'article_path': passthrough.get("article_path"),
                 'expanded_folder': passthrough.get("expanded_folder"),
                 'status': passthrough.get("status"),
-                'update_date': update_date,
+                'update_date': passthrough.get("update_date"),
                 'published': ingest_publish
             }
             response_message = {
```

gnott:
I followed the data as carefully as I could through the workflows. 

The first commit (https://github.com/elifesciences/elife-bot/pull/595/commits/7a4c52983106da14c5bac9260ca090467e348fc0) should now shift the responsibility of setting ``publish`` to be ``True`` or ``False`` from the Lax ingest response, and whether ``force`` is ``True`` or ``False``. This is / was previously set by the ``SetPublicationStatus`` activity, which read the EIF file or the response from the website, which would be incorrect or unavailable now or in the future.

The second commit (https://github.com/elifesciences/elife-bot/pull/595/commits/455bac62c697034a34c6db4fb78df561f2556537) I think will solve the missing ``article_path`` attribute by taking it from the session instead of from the workflow data. It is added to the session in the ``ConvertJATS`` activity, and in the absence of data from ``shimmy.py`` in this quick code shortcut in the workflows, it was not getting added to data.

It looks like the ``SetPublicationStatus`` activity can be removed very soon, because I think it is only concerned with adding the ``published`` value to the EIF file. Then we won't need the ``article_path`` session value added by ``ConvertJATS``, which is an activity that will also get removed.

``PostEIFBridge`` can later be renamed perhaps and instead of getting data from the EIF file (which will no longer be produced) can get data from the session or from Lax to populate the dashboard and SQS messages. I haven't narrowed down what its future purpose is yet.

I think this is ready for another try if you want to merge @giorgiosironi, otherwise I can try it out tomorrow and revert if it still isn't passing.

giorgiosironi:
> The first commit (7a4c529) should now shift the responsibility of setting publish to be True or False from the Lax ingest response, and whether force is True or False. This is / was previously set by the SetPublicationStatus activity, which read the EIF file or the response from the website, which would be incorrect or unavailable now or in the future.

Seems correct, but we still need to read `published` from the session rather than the workflow data in `PostEIFBridge`, right?

> The second commit (455bac6) I think will solve the missing article_path attribute by taking it from the session instead of from the workflow data. It is added to the session in the ConvertJATS activity, and in the absence of data from shimmy.py in this quick code shortcut in the workflows, it was not getting added to data.

Cross-checking `ConvertJATS`, makes sense.

giorgiosironi:
Added the session read for `published`. I'll merge and try it out on end2end now.

giorgiosironi:
```
2018-01-12T11:18:35Z ERROR queue_workflow_starter_3825 Exception while processing message
  File "queue_workflow_starter.py", line 80, in process_message
    start_workflow(name, data)
  File "/opt/elife-bot/venv/local/lib/python2.7/site-packages/newrelic-2.72.0.52/newrelic/api/background_task.py", line 102, in wrapper
    return wrapped(*args, **kwargs)
  File "queue_workflow_starter.py", line 99, in start_workflow
    s.start(settings=settings, **workflow_data)
TypeError: start() got an unexpected keyword argument 'force'
```
a small integration problem to solve

Jenniferstrej:
If we want to remove completely reference from the old elife-website, we can remove all: ConvertJats, shimmy, PostEIFBridge, PreparePostEIF. There are only 2 extractions to do from those it seems, the session variable called eif_location and the property change in the dashboard. Silent corrections will work because the publish force is set earlier on on the SilentCorrections workflow.

Jenniferstrej:
I posted some more details on https://github.com/elifesciences/elife-bot/pull/553

gnott:
Regarding the ``published`` property and taking it from the session: I'm rethinking how I set ``published`` in ``VerifyLaxResponse`` based on the ``force`` value.

It was a simple way to set the ``published`` value which I think came from the elife-website. Since the communication with it was shut off, the next activity that is run is ``VerifyLaxResponse`` so I set the ``published`` value there.

The ``published`` value may only be used in EIF files (since it is the value returned by elife-website) and it could be it is not used elsewhere. Lax would also return a published value and I'm not sure how that relates to the session value.

A careful review of the ``published`` session value should be done when the code cleanup happens, and the logic I added to ``VerifyLaxResponse`` may need to be removed later (if ``published`` is no longer required or is set later in the workflows).

giorgiosironi:
> The published value may only be used in EIF files (since it is the value returned by elife-website) and it could be it is not used elsewhere. Lax would also return a published value and I'm not sure how that relates to the session value.

I think `published` ends up in the dashboard, that decides whether to show the `Publish` button or not for that run. It should use the value returned by Lax I guess from now on I guess.

gnott:
That sounds important!

I may have not been completely wrong then in the logic I added (if @Jenniferstrej gets a chance to review). In silent correction situations, it only invokes the ``IngestToLax`` activity, and never the ``PublishToLax`` activity (which is in the ``ApproveArticlePublication`` workflow). So, when ``VerifyLaxResponse`` activity happens, if ``force`` was ``True`` then ``publish`` should be ``True``.

PR #602, opened 2018-01-17
Run precise subset of end2end tests

giorgiosironi:
Only the end2end tests that involve the bot are run. See https://github.com/elifesciences/elife-spectrum/commit/e55d6bd6778f0c11df38fd09f71567bad0746c7a where we add the new tag, which was born to mark the pubmed test.

We don't run tests that:
- do not involve the bot
- even if involving the bot, do not add anything new with respect to the core bot tests that publish several articles and versions

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -27,7 +27,7 @@ elifePipeline {
                     revision: commit,
                     folder: '/opt/elife-bot',
                 ],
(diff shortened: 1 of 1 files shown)


giorgiosironi:
Will merge this when green, to start running the Pubmed test from ELPP-3360 which is only tagged with `bot`.

PR #607, opened 2018-01-23
Elpp 2698 - remove eif references

Jenniferstrej:
Not all references to eif_location have been removed as the activities that use it haven't been deleted yet, only disabled. However eif_location has been removed from everywhere else and a new activity, ReadyToPublish is doing the job of sending the message to the dashboard.

Once this works and is deployed, we can start deleting the unused activities.


giorgiosironi:
The build is missing something in `settings.py`?
```
[54.147.211.162] out: ************* Module activity.activity_ReadyToPublish
[54.147.211.162] out: E: 24,18: No value for argument 'session_key' in function call (no-value-for-parameter)
```

Jenniferstrej:
This looks like some error I had on the test, where I don't mock the session. I ran it locally though, strange.
The change isn't big, except for the eif_location being removed, which is a dependency in many places, so the change is going to be 'big' eventually. I did not remove the dependency from the activities themselves as there are tests for them still active and I haven't removed them completely.

Jenniferstrej:
@giorgiosironi - one way of removing this dependency slowly is to change the meaning of the EIF activities, by removing the dependency on eif, but some of them do access this variable and depend on it, so the smallest step I could think of was to remove those activities from the workflow definition and remove the variable. I agree with what you say but in this case, for example, lax_response_adapter is used between the two workflows (ingest and before publish), so if I remove the dependency in the last one, I still will have it 'complaining' on the first. Having `eif_location` dependency this way is causing  all those problems and trying to remove it slowly will be more time consuming and more prone to errors in my opinion.

Jenniferstrej:
by the way, there was a settings file missing. The session interface has been changed and I didn't know exactly how it was and got confused. It's fixed now.

workflow/workflow_SilentCorrectionsProcess.py
```diff
@@ -100,28 +78,6 @@ def __init__(self, settings, logger, conn=None, token=None, decision=None,
                         "schedule_to_start_timeout": 300,
                         "start_to_close_timeout": 60 * 5
                     },
-                    {
```
giorgiosironi: I'll review these to see if they set anything in the session/S3/etc that could become missing in downstream activities

Jenniferstrej: That's good to double check with @gnott also.

gnott: I cannot see anything EIF related where these cannot be removed, though end2end may discover something.

gnott:
I like the modifications here as a partial clean-up.

workflow/workflow_PostPerfectPublication.py
```diff
@@ -55,17 +55,6 @@ def __init__(self, settings, logger, conn=None, token=None, decision=None,
                         "schedule_to_start_timeout": 300,
                         "start_to_close_timeout": 60 * 5
                     },
-                    {
```
giorgiosironi: This activity emits a `Rewrite EIF` monitor event, which will disappear as expected; it rewrites the EIF in a bucket changing the update date, which won't happen anymore. Seems safe to remove.

workflow/workflow_SilentCorrectionsProcess.py
```diff
@@ -100,28 +78,6 @@ def __init__(self, settings, logger, conn=None, token=None, decision=None,
...
-                        "version": "1",
-                        "input": data,
-                        "control": None,
-                        "heartbeat_timeout": 60 * 5,
-                        "schedule_to_close_timeout": 60 * 5,
-                        "schedule_to_start_timeout": 300,
-                        "start_to_close_timeout": 60 * 5
-                    },
-                    {
-                        "activity_type": "PostEIFBridge",
```
giorgiosironi: In this workflow, is the `if published is True` check of `PostEIFBridge` true or not? In both cases the activity was executing `set_monitor_property` and/or sending a message to the dashboard, which may not be doing anymore?

Jenniferstrej: it won't be doing it anymore and the `set_monitor_property` is now being done (it was extracted into) by the `ReadyToPublish` activity. 
The `if published is True` is checking for a possible silent correction, but only on the EIF context. Which we will not be supporting anymore

Jenniferstrej: The new website silent correction is done elsewhere.

Jenniferstrej: SilentCorrections won't have a ready to publish status ever, so we don't need to add that activity in the Silen Correction workflows.

Jenniferstrej: On the new website the `published` `set_monitor_property` is set here https://github.com/elifesciences/elife-bot/blob/develop/activity/activity_VerifyPublishResponse.py#L29

workflow/workflow_SilentCorrectionsProcess.py
```diff
@@ -100,28 +78,6 @@ def __init__(self, settings, logger, conn=None, token=None, decision=None,
                         "schedule_to_start_timeout": 300,
                         "start_to_close_timeout": 60 * 5
                     },
-                    {
-                        "activity_type": "SetEIFPublish",
```
giorgiosironi: this activity doesn't look to be writing anything apart from EIF, should be safe to remove

workflow/workflow_SilentCorrectionsProcess.py
gnott: A good point @giorgiosironi, which would be not easily noticed. I think in this case the "published" session data originated from the elife-website EIF interaction, and it was added to allow the previous EIF short-circuiting change we did earlier, to remove dependence on the elife-website.

If I search activity code for ``session.get_value('published')`` it only appears in ``PostEIFBridge``.

I'm not sure what the dashboard does with the value of ``published`` though, maybe @Jenniferstrej can comment.

workflow/workflow_SilentCorrectionsProcess.py
```diff
@@ -56,28 +56,6 @@ def __init__(self, settings, logger, conn=None, token=None, decision=None,
...
-                        "version": "1",
-                        "input": data,
-                        "control": None,
-                        "heartbeat_timeout": 60 * 5,
-                        "schedule_to_close_timeout": 60 * 5,
-                        "schedule_to_start_timeout": 300,
-                        "start_to_close_timeout": 60 * 5
-                    },
-                    {
-                        "activity_type": "SetPublicationStatus",
```
giorgiosironi: also seems to only touch the EIF file on S3, should be safe to remove

workflow/workflow_SilentCorrectionsProcess.py
```diff
@@ -56,28 +56,6 @@ def __init__(self, settings, logger, conn=None, token=None, decision=None,
                         "schedule_to_start_timeout": 300,
                         "start_to_close_timeout": 60 * 5
                     },
-                    {
-                        "activity_type": "ConvertJATS",
```
giorgiosironi: also seems to only touch the EIF file on S3, should be safe to remove

workflow/workflow_SilentCorrectionsProcess.py
Jenniferstrej: This still seems only relevant to me for EIF, on the new website and lax publication, the session is set here https://github.com/elifesciences/elife-bot/blob/5f94f979a99c977d62bc8a35b40d7a6616cdc826/activity/activity_VerifyLaxResponse.py#L51-L54

provider/lax_provider.py
```diff
@@ -162,12 +162,11 @@ def get_xml_file_name(settings, expanded_folder, xml_bucket, version=None):
     return xml_file_name
 
 
-def lax_token(run, version, expanded_folder, status, eif_location, force=False):
+def lax_token(run, version, expanded_folder, status, force=False):
     token = {
         'run': run, 
         'version': version,
         'expanded_folder': expanded_folder,
-        'eif_location': eif_location,
```
giorgiosironi: for posterity: this deletion is what correlated with the starters removal of the `eif_location` parameter

workflow/workflow_SilentCorrectionsProcess.py
Jenniferstrej: actually, this session value is also not relevant but for EIF cases.

PR #611, opened 2018-01-24
Removes PostEIFBridge reference

Jenniferstrej:
I was thinking of creating one commit per activity reference deleted and have them all together on a PR, but decided to make one PR with one activity only and check with you guys your preference.


giorgiosironi:
These activities should not be used in workflows anymore, so at this point they should be independent. Would still go for merging one as soon as it is removed, because if any unknown problem happens in end2end or prod it's easy to trace it back to a few removed files.

workflow/workflow_ArticleInformationSupplier.py
```diff
@@ -33,17 +33,6 @@ def __init__(self, settings, logger, conn=None, token=None, decision=None,
 
             "steps":
```
giorgiosironi: is this workflow still executed? If so shouldn't it be already broken because of the lack of the other EIF activities for new article runs?

Jenniferstrej: It won't be used, I was going to first delete the activities then the empty workflows, if we are going to merge them gradually.

giorgiosironi: no need to remove it yet if it's unused

PR #612, opened 2018-01-24
article path support outside eif json

Jenniferstrej:


tests/activity/settings_mock.py
```diff
@@ -138,4 +138,7 @@
 pdf_cover_generator = "https://localhost/personalised-covers/"
 
 elifepubmed_config_file = 'tests/activity/pubmed.cfg'
-elifepubmed_config_section = 'elife'
\ No newline at end of file
+elifepubmed_config_section = 'elife'
+
+# Article path
+article_path_pattern = "/articles/{id}v{version}"
```
giorgiosironi: @thewilkybarkid

activity/activity_ReadyToPublish.py
```diff
@@ -51,7 +53,13 @@ def do_activity(self, data=None):
 
         return activity.activity.ACTIVITY_SUCCESS
 
-    def prepare_ready_to_publish_message(self, article_id, version, run, expanded_folder, status, update_date):
+    def preview_path(self, article_path_pattern, article_id, version):
+        path = article_path_pattern.replace('{id}', article_id)
+        path = path.replace('{version}', version)
```
giorgiosironi: I checked that also in Python 2 you can still write:
```
>>> "/articles/{id}v{version}".format(id=42, version=3)
'/articles/42v3'
```

Jenniferstrej: Oopsy! Cool!

tests/activity/settings_mock.py
giorgiosironi: added to the bot settings (I assume that's always right in any condition)

thewilkybarkid: For viewing a version of an article on Journal? Yeah. (If you go to the latest version it doesn't redirect, but does have `<link rel="canonical">` etc correctly and generates links without the version.)

PR #617, opened 2018-01-25
SetEIFPublish full removal

Jenniferstrej:


giorgiosironi:
A quick grep also finds `register.py`?

Jenniferstrej:
oops, missed from that one, will add it. Thanks.

PR #619, opened 2018-01-29
If Pubmed FTP delivery fails, then return a permanent failure.

gnott:
This should cut down on the repeated attempts of the PubMed deposits if there is an FTP error. Not retrying the FTP transfer if the first attempt fails is probably ok in this case. The files it processed in that case should remain in the outbox, and the deposit will be attempted again in about an hour (since the ``cron.py`` attempts to start a workflow each hour).

### DIFF
--- a/activity/activity_PubmedArticleDeposit.py
+++ b/activity/activity_PubmedArticleDeposit.py
@@ -119,7 +119,11 @@ class activity_PubmedArticleDeposit(activity.activity):
         if len(self.article_published_file_names) > 0:
             self.add_email_to_queue()
 
-        return self.activity_status
+        # return a value based on the activity_status
+        if self.activity_status is True:
+            return True
+        else:
+            return activity.activity.ACTIVITY_PERMANENT_FAILURE
 
     def set_datestamp(self):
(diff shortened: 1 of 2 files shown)


activity/activity_PubmedArticleDeposit.py
```diff
@@ -119,7 +119,11 @@ def do_activity(self, data=None):
         if len(self.article_published_file_names) > 0:
             self.add_email_to_queue()
 
-        return self.activity_status
+        # return a value based on the activity_status
+        if self.activity_status is True:
```
giorgiosironi: could just be `if self.activity_status:` unless strange values like `0` are used?

gnott: Probably yes in this case. There are a bunch of status values all set originally to ``None`` and not all of them get set to ``True`` or ``False`` such as the ``publish_status`` one. I guess my thinking is ``is True`` follows the same pattern as the ``is False`` comparisons, which respond differently if the value is ``None``. Null versus not null values. I'll leave it in if that's ok, but I appreciate the careful review!

PR #634, opened 2018-02-13
Activity to modify article subjects

gnott:
New activity ``ModifyArticleSubjects`` to modify the subjects of an article.

Basically,
* download the article XML
* parse its DOI
* download the article subjects data CSV file (if specified and if it exists)
* extract the subjects data that pertains to the specific DOI
* if there is subjects data to replace existing tags, use the function from the elifetools library ``xmlio.py`` which has a function to rewrite the subjects in XML
* if there were any replacements done, upload the article XML back to the bucket

It should return ``True`` (or activity success) if there are no settings for the activity, there is no data, there is no data for the particular DOI, etc.

This initial draft may be a little rough, and it hasn't been tested in a real workflow yet. The test coverage, I think, is fairly good, abstracting all the bucket operations with the ``storage_context`` provider.


activity/activity_ModifyArticleSubjects.py
```diff
@@ -126,7 +126,7 @@ def download_article_xml(self, expanded_bucket_name, expanded_folder_name):
         if not expanded_bucket_name or not expanded_folder_name:
             if self.logger:
                 self.logger.info('could not download article xml without bucket settings')
-            return False
+            return None
```
seanwiseman: You can remove `return None` altogether. The function will return `None` by default if you do not specify a return value.

workflow/workflow_SilentCorrectionsIngest.py
```diff
@@ -100,6 +100,17 @@ def __init__(self, settings, logger, conn=None, token=None, decision=None,
                         "schedule_to_start_timeout": 300,
                         "start_to_close_timeout": 60 * 10
                     },
+                    {
+                        "activity_type": "ModifyArticleSubjects",
+                        "activity_id": "ModifyArticleSubjects",
+                        "version": "1",
+                        "input": data,
+                        "control": None,
+                        "heartbeat_timeout": 60 * 10,
```
seanwiseman: Could be put into a constant as its commonly used here:
 `DEFAULT_TIMEOUT = 60 * 10`
or
`TEN_MINUTES = 60 * 10`

tests/activity/test_activity_modify_article_subjects.py
```diff
@@ -0,0 +1,353 @@
+import unittest
+import shutil
+import os
+from io import StringIO
+from collections import OrderedDict
+from ddt import ddt, data, unpack
+from activity.activity_ModifyArticleSubjects import activity_ModifyArticleSubjects
+from mock import patch
+import tests.activity.test_activity_data as test_data
+from tests.activity.classes_mock import FakeLogger
```
seanwiseman: Could group these 3 imports?

tests/activity/test_activity_modify_article_subjects.py
```diff
@@ -0,0 +1,353 @@
...
+    def clean_directories(self):
+        self.activity.clean_tmp_dir()
+        helpers.delete_files_in_folder('tests/tmp', filter_out=['.keepme'])
+        #helpers.delete_files_in_folder(test_data.ExpandArticle_files_dest_folder)
+
+    @patch('activity.activity_ModifyArticleSubjects.storage_context')
+    @patch.object(activity_ModifyArticleSubjects, 'emit_monitor_event')
+    @patch('activity.activity_ModifyArticleSubjects.get_session')
+    @patch.object(FakeStorageContext, 'list_resources')
+    @data(
```
seanwiseman: Parameterized data looks nice an clean here, easy to follow even though its a busy test 👍

tests/activity/test_activity_modify_article_subjects.py
```diff
@@ -0,0 +1,353 @@
...
+        xml_file_name = os.path.join(self.test_files_dir_name, 'elife-29353-v1.xml')
+        file_name = xml_file_name.split('/')[-1]
+        shutil.copy(xml_file_name, os.path.join(self.activity.get_tmp_dir(), file_name))
+        expected = '10.7554/eLife.29353'
+        doi = self.activity.article_doi(xml_file_name)
+        self.assertEqual(doi, expected)
+
+
+    def test_data_settings(self):
+        "test for when no settings are provided"
```
seanwiseman: I think some of these comments could just become the test name as they seem to explain the test case better.

tests/activity/test_activity_modify_article_subjects.py
```diff
@@ -0,0 +1,353 @@
...
+         'modify_article_subjects_snippet_01.xml'
+        ),
+        )
+    @unpack
+    def test_modify_article_subjects(self, xml_file_name, subjects_map,
+                                     expected_total, expected_snippet_file):
+        "test rewriting the XML file"
+        # copy the XML file in
+        article_xml_file = os.path.join(self.activity.get_tmp_dir(), xml_file_name)
+        shutil.copy(os.path.join(self.test_files_dir_name, xml_file_name),
```
seanwiseman: If there were more of these file copy operations required in this suite then it might be worth rethinking how these are set up to thin out the actual tests, but it looks isolated at present.

activity/activity_ModifyArticleSubjects.py
```diff
@@ -0,0 +1,273 @@
...
+        try:
+            expanded_folder_name = session.get_value('expanded_folder')
+            expanded_bucket_name = (self.settings.publishing_buckets_prefix
+                                    + self.settings.expanded_bucket)
+
+            # main
+            article_xml_file = self.download_article_xml(expanded_bucket_name,
+                                                         expanded_folder_name)
+            if not article_xml_file:
+                if self.logger:
```
seanwiseman: This pattern of checking if the instance has a truthy `self.logger` attribute before attempting to log seems a little problematic. 

I understand it's down to the caller of the class, in this case, to pass in a `logger` instance (or something at least as there is no default) but you could remove all these checks throughout the class if you just initialized a default `logger` if the passed in value was not true/a logger instance.

Also, there are checks before calling `self.logger` in most places, but in the exception handlers its called without these checks.

activity/activity_ModifyArticleSubjects.py
```diff
@@ -0,0 +1,273 @@
...
+                                    " message:" + str(e.message))
+            return activity.activity.ACTIVITY_PERMANENT_FAILURE
+
+        return activity.activity.ACTIVITY_SUCCESS
+
+
+    def load_subjects_data(self):
+        "download and parse the subjects data from the CSV in the bucket"
+        subjects_data = None
+        (data_bucket_name, data_file_name) = self.data_settings()
```
seanwiseman: Can remove the parenthesis here, tuple unpacking will assign each value:
can be: `data_bucket_name, data_file_name = self.data_settings()`

activity/activity_ModifyArticleSubjects.py
```diff
@@ -0,0 +1,273 @@
...
+
+    def data_settings(self):
+        "check if there are settings to specify where the article subjects data is stored"
+        data_bucket_name = None
+        data_file_name = None
+        if (hasattr(self.settings, 'article_subjects_data_bucket') and
+                hasattr(self.settings, 'article_subjects_data_file')):
+            data_bucket_name = self.settings.article_subjects_data_bucket
+            data_file_name = self.settings.article_subjects_data_file
+        return (data_bucket_name, data_file_name)
```
seanwiseman: Can remove the parenthesis here, when returning comma separated values python will wrap them in a tuple automatically.

activity/activity_ModifyArticleSubjects.py
```diff
@@ -0,0 +1,273 @@
...
+        xmlio.register_xmlns()
+
+        root, doctype_dict = xmlio.parse(article_xml_file, return_doctype_dict=True)
+
+        # Modify subject values
+        total = xmlio.rewrite_subject_group(root, subjects, subject_group_type)
+
+        # Start the file output
+        reparsed_string = xmlio.output(root, type=None, doctype_dict=doctype_dict)
+        open_file = open(article_xml_file, 'wb')
```
seanwiseman: Safer to use `with` context manager as will always close the file even if exceptions are raised.

workflow/workflow_SilentCorrectionsIngest.py
```diff
@@ -100,6 +100,17 @@ def __init__(self, settings, logger, conn=None, token=None, decision=None,
                         "schedule_to_start_timeout": 300,
                         "start_to_close_timeout": 60 * 10
                     },
+                    {
+                        "activity_type": "ModifyArticleSubjects",
```
giorgiosironi: Are we planning for all silent corrections to always execute this? May be temporary, but I'd like to make sure that if there is an unrelated silent correction it can still go through (albeit it could take some seconds more because of the additional activity.)

activity/activity_ModifyArticleSubjects.py
```diff
@@ -0,0 +1,273 @@
...
+        "check if there are settings to specify where the article subjects data is stored"
+        data_bucket_name = None
+        data_file_name = None
+        if (hasattr(self.settings, 'article_subjects_data_bucket') and
+                hasattr(self.settings, 'article_subjects_data_file')):
+            data_bucket_name = self.settings.article_subjects_data_bucket
+            data_file_name = self.settings.article_subjects_data_file
+        return (data_bucket_name, data_file_name)
+
+    def download_subjects_file(self, data_bucket_name, data_file_name):
```
giorgiosironi: There may be a concurrency problem here if multiple processes try to write to the same path on the filesystem, right?
Perhaps the easier way to sidestep this problem would be to keep the content of the file in memory rather than in the temp folder (e.g. `StringIO` or similar)? I am imagining it as ~1,000 lines so it should fit easily?

activity/activity_ModifyArticleSubjects.py
```diff
@@ -0,0 +1,273 @@
...
+from provider.storage_provider import storage_context
+from provider.execution_context import get_session
+from provider.article_structure import ArticleInfo
+from elifetools import xmlio
+
+"""
+ModifyArticleSubjects.py activity
+"""
+
+class activity_ModifyArticleSubjects(activity.activity):
```
seanwiseman: Not sure of the reason for it being like this, but this violates pythons class naming convention twice by including an `_` and not starting with a capital.

If something starts lowercase you should be able to safely assume its a function, likewise starting with a capital and being camelcase should indicate a class.

activity/activity_ModifyArticleSubjects.py
gnott: I did not know that one, thanks!

workflow/workflow_SilentCorrectionsIngest.py
gnott: For this one I should have set five minutes anyway instead of ten, not that it would have too much trouble either way.

I would be open to adding some defaults perhaps at a later time when we can do it throughout all the workflow definitions? Perhaps a new JIRA ticket for bot cleanup. There's probably one in there for refactoring the workflow object too that would set lots of defaults so we do not need to specify them.

tests/activity/test_activity_modify_article_subjects.py
gnott: I like it, thanks.

tests/activity/test_activity_modify_article_subjects.py
gnott: Bad comments on my part, because this test is for a successful read of the settings. The one below it is when no settings are available. Do you mean to change ``test_data_settings()`` to something like ``test_data_settings_success()``?

tests/activity/test_activity_modify_article_subjects.py
gnott: Good suggestion, I think I started out with one test copying a file and in the end I ended up with three tests. I'm refactoring the file copying into its own function.

activity/activity_ModifyArticleSubjects.py
gnott: I was almost going to refactor the logging line but not yet. Maybe we could add this in a later PR where the logger value of an activity object is set even if you do not pass it in? Then we can go throughout all the activities and drop the conditional check.

activity/activity_ModifyArticleSubjects.py
gnott: :+1:

activity/activity_ModifyArticleSubjects.py
gnott: :+1:

activity/activity_ModifyArticleSubjects.py
gnott: I almost did that too, but I left it. I took it from ``ApplyVersionNumber`` which is similar to this activity. I will improve it here

workflow/workflow_SilentCorrectionsIngest.py
gnott: The first use of the activity I intended to be as part of silent corrections. If the CSV data in the bucket is limited to only the article we want to alter, I think it should be ok. I can remove the file after doing the silent corrections for now too, since there is also some manual clean up I will need to do.

There's been some discussion how this activity may remain enabled for all silent corrections with data present, so as to not silently correct an article with the old subject values. Also, to possibly add this activity to the regular ingest workflow, to catch any future versions of the article does not use old subject values. These are not decided yet.

Putting it in the silent correction workflow was a suggestion as opposed to my original plan which was to create a new workflow specifically for changing subject values. I think allowing this activity to be added to a workflow at any time will provide flexibility in categorising or re-categorising the articles.

We could add an additional flag in the settings to skip all the logic of this activity, or we could make a data flag when the workflow is started whether to run this activity or not. That could solve the potential complications and performance issues you are describing.

activity/activity_ModifyArticleSubjects.py
gnott: I think this will be ok as far as OS level file concurrency, since each activity writes data to its own separate directory in the bot temp dir. Performance shouldn't be too much worse than downloading any other file from a bucket. S3 should be able to support concurrent downloading of the same object by multiple activities at least at the several workflows we run at once. Maybe if 200 workers all performed this activity at the same time it would cause some delays or timeouts.

activity/activity_ModifyArticleSubjects.py
gnott: These object names are in violation! I think here again if we could tackle the object renaming as a separate PR and rename them all at once, it might be the most orderly way to do it. I think this has been on the pending tasks for a couple years now. Let's do it! (soon)

gnott:
Much appreciated to receive the detailed notes here you provided. I left some comments about some tasks which could be done later as part of a more comprehensive refactoring effort. 

If there's anything you would like to change before merging, please feel free to edit it yourself, or leave a follow-up comment. If you want to merge and test it before I'm online you are welcome to.

activity/activity_ModifyArticleSubjects.py
seanwiseman: Yes, the establishment (any linter 😉 ) will not like it...

workflow/workflow_SilentCorrectionsIngest.py
giorgiosironi: No problem for me in always executing this activity as long as it works for all articles (don't think performance is going to be a big problem).

PR #635, opened 2018-02-14
Accept 202 from personalised-covers

giorgiosironi:
Spawned from a randomly failing end2end test.

https://github.com/elifesciences/personalised-covers/blob/cceb10b59caf19218dcfac59a3da9abe9ec3da8e/src/App/DefaultController.php#L127 shows either 200 or 202 can be returned, both are successes for the activity.


giorgiosironi:
Yes, there is no clear-cut specification, but from the referenced file we can see the body is always there.

PR #641, opened 2018-02-19
Create more unique workflow execution ID values in silent correction …

gnott:
…starters.

Small changes as described in https://elifesciences.atlassian.net/browse/ELPP-3460 to make it easier to run concurrent silent corrections of the same article_id, by adding ".[version]" to the workflow execution ID value.

### DIFF
--- a/starter/starter_SilentCorrectionsProcess.py
+++ b/starter/starter_SilentCorrectionsProcess.py
@@ -47,7 +47,7 @@ class starter_SilentCorrectionsProcess():
         child_policy, \
         execution_start_to_close_timeout, \
         workflow_input = helper.set_workflow_information(self.const_name, "1", None, input,
-                                                         article_id + "." + str(version))
+                                                         "%s.%s" % (article_id, version))
 
         # Simple connect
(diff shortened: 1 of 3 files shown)


starter/starter_SilentCorrectionsProcess.py
```diff
@@ -46,7 +46,8 @@ def start(self, settings, article_id, version, requested_action, force, result,
         workflow_version, \
         child_policy, \
         execution_start_to_close_timeout, \
-        workflow_input = helper.set_workflow_information(self.const_name, "1", None, input, article_id)
+        workflow_input = helper.set_workflow_information(self.const_name, "1", None, input,
+                                                         article_id + "." + str(version))
```
giorgiosironi: I'd go for the standard interpolation `"%s.%s" % (article_id, version)` that avoids having to remember to convert with `str()`

gnott: Sure, I will do these here thanks. Recently I've been preferring to use the ``.format()`` style, and I don't know if we should prefer one over the other in the future?

PR #653, opened 2018-03-08
altered workflow called from S3 notification

jhroot:
and ammended IngestArticleZip starter as no longer recieves S3 notification


jhroot:
@gnott please review, particularly the swf workflow id naming change

giorgiosironi:
Closed?

PR #654, opened 2018-03-09
Feature/version reason

jhroot:
modifications to starter mechanism and test for ingest workflow

### DIFF
--- a/starter/starter_IngestArticleZip.py
+++ b/starter/starter_IngestArticleZip.py
@@ -21,7 +21,7 @@ class starter_IngestArticleZip():
     def __init__(self):
         self.const_name = "IngestArticleZip"
         
-    def start(self, settings, run, article_id, version_reason, scheduled_publication_date):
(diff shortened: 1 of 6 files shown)


starter/starter_IngestArticleZip.py
```diff
@@ -21,24 +21,25 @@ class starter_IngestArticleZip():
...
+        }
 
         workflow_id, \
         workflow_name, \
         workflow_version, \
         child_policy, \
         execution_start_to_close_timeout, \
         workflow_input = helper.set_workflow_information(self.const_name, "1", None, input,
-                                                         info.file_name.replace('/', '_'),
+                                                         "ingest-" + article_id,
```
gnott: Only comment here is in previous reviews we are preferring to use string interpolation for creating the workflow ID here, I will add this commit myself so I can merge this PR today.

gnott: I'll do this as a separate PR, will be easier.

Issue #588, opened 2018-01-06
SendDashboardProperties change to not return False


gnott:
Recent issues with the ``SendDashboardProperties`` activity causes a workflow to retry the activity for a very long time for what is probably an unrecoverable error.

The first example was a zip file was missing a valid article XML. For this, the activity returns False due to not finding an XML file.

The second example was an article with an incomplete pub date, which fails to match a required pattern, and returns False.

Instead of returning False this activity would be improved by returning

```
return activity.activity.ACTIVITY_PERMANENT_FAILURE
```

This will stop the workflow at this step and not continue to retry for hours.

Retrying at this stage causes the ``decider.log`` file to fill up.

PR #657, opened 2018-03-10
SendDashboardProperties to cause a permanent failure in XML parsing …

gnott:
…situations.

Includes some tests as described in the issue for two errors we've observed before.
1. When no XML file is available to parse - return a permanent failure
2. When the pub date in the XML cannot be parsed - return a permanent failure

If there is an error in emitting a message to the queue, I left it as returning ``False``, which can then retry the activity, if there is some temporary network problem.

Fixes https://github.com/elifesciences/elife-bot/issues/588


tests/activity/test_activity_send_dashboard_properties.py
```diff
@@ -51,5 +54,51 @@ def test_do_activity(self, fake_emit_monitor_property, fake_session, fake_s3_moc
...
+        directory = TempDirectory()
+        fake_session.return_value = FakeSession(test_data.session_example)
+        fake_s3_mock.return_value = FakeS3Connection()
+        fake_get_article_xml_key.return_value = None, None
+
+        result = self.send_dashboard_properties.do_activity(test_data.ConvertJATS_data)
+
+        self.assertEqual(result, self.send_dashboard_properties.ACTIVITY_PERMANENT_FAILURE)
+
+        directory.cleanup()
```
giorgiosironi: if you put `directory` on `self` this could become:
```
if self.directory:
    self.directory.cleanup()
```
in the `tearDown()` method, which will always be executed also in case of exceptions.

gnott: :+1:

PR #661, opened 2018-03-14
Uninstall elifepubmed before installing requirements.

gnott:
After reviewing the PubMed deposit generated from a retraction article (https://github.com/elifesciences/elife-pubmed-feed/issues/40) it did not include the ``<Object Type="Retraction">`` as expected.

I think even though we have the correct git commit for the ``elifepubmed`` library in ``requirements.txt`` here, it is not reinstalling it on the instance properly.

This is to uninstall it first (as we do with other libraries) when running ``install.sh`` so we have the desired library commit.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -17,5 +17,8 @@ fi
 if pip list | grep elifecrossref; then
     pip uninstall -y elifecrossref
 fi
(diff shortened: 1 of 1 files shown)


gnott:
:+1:

PR #664, opened 2018-03-15
ELPP-1265  Feature automate silent corrections

gnott:
Looks like silent corrections are working well, and this will connect the S3 notifications to start a new silent corrections workflow. 

I added a very basic test for ``queue_worker.py`` since it had no tests yet. It could probably be improved if I refactored it a bit where is it passed the ``settings``, instead if it being passed ``ENV``, then we could mock the settings in tests easier.


gnott:
I think the settings may need to be imported like it is in ``lax_response_adapter.py`` in order for the travis-ci tests to pass as written. Please indicate a thumbs up if you like my suggestion to refactor how settings is instantiated and passed in.

newFileWorkflows.yaml
```diff
@@ -1,4 +1,9 @@
 ArticleZip:
   bucket_name_pattern: '.*elife-production-final$'
   file_name_pattern: '.*\.zip'
-  starter_name: 'InitialArticleZip'
\ No newline at end of file
+  starter_name: 'InitialArticleZip'
+
+SilentCorrectionsArticleZip:
```
giorgiosironi: are the S3 notifications for this bucket already set up in production? They need to be in builder to be replicated across environments like `end2end`.

giorgiosironi:
`LaxResponseAdapter` is injected with `settings` and `logger`. Though if you look at it what it really needs instead of `settings` is `input_queue` and `output_queue` so I'd inject those.

There is no `QueueWorker` class (only a `work` function) but if the module was refactored to create one, its dependencies would similarly be `queue`, `out_queue` and `logger`. So I'd inject these three objects inside the constructor. Then the tests can mock those not having to worry about `boto.sqs` or its connection.

newFileWorkflows.yaml
gnott: Checking, and I do not think notification events are configured on the bucket. Should these be defined around here? https://github.com/elifesciences/builder/blob/master/projects/elife.yaml#L632

If so, I think the bucket names, queues, notification settings, are not complete. Understandable since many of these were probably configured manually and were not the result of a builder action.

I would appreciate some help here to confirm what and how much detail should be in builder, or if I'm pointing to the wrong place!

giorgiosironi: Essentially since builder uses CloudFormation underneath, it has the same limitations of only managing buckets that were not already existing when it was introduced. Therefore `elife-production-final` and similar older buckets cannot be put there.

However, `{instance}-elife-silent-corrections` is there so we can setup notifications there. It will look like:
```
                end2end-elife-silent-corrections:
                    sqs-notifications:
                        end2end-incoming-queue: {}
```

The order of deployment should be:
- this PR (which won't however trigger the workflow automatically)
- builder changes (which will setup the notification) in `end2end`. I can write a PR for that.
- Update of the end2end test to remove the explicit silent corrections call (and leave in only the file upload)
- builder changes in `continuumtest` and `prod`

giorgiosironi: https://github.com/elifesciences/builder/pull/257

gnott:
I refactored to create the ``QueueWorker`` object, in order to pass it settings, initially. I changed it slightly to allow testing of the ``work()`` method too, by substituting the ``green()`` process monitor, and overriding the 10 second timeout for faster testing.

The test data in the ``FakeS3Event`` object has some defaults, but if it's too messy please send further suggestions.

I could also potentially write some tests for when it fails to connect to the queue, or if the workflow name is not found in the YAML file, or we could just write those edge case tests if they ever come up in a real-life situation.

queue_worker.py
```diff
@@ -22,89 +21,109 @@
...
+    def get_starter_name(self, rules, info):
+        for rule_name in rules:
+            rule = rules[rule_name]
+            if re.match(rule['bucket_name_pattern'], info.bucket_name) and \
+                    re.match(rule['file_name_pattern'], info.file_name):
+                return rule['starter_name']
+            pass
 
 
 def reload_module(module_name):
```
giorgiosironi: do we need this anymore? Seems not called

tests/test_queue_worker.py
```diff
@@ -0,0 +1,97 @@
...
+import json
+from mock import Mock, patch
+import settings_mock
+from queue_worker import QueueWorker
+from S3utility.s3_notification_info import S3NotificationInfo
+import tests.test_data as test_data
+from tests.activity.classes_mock import FakeSQSConn, FakeSQSMessage, FakeSQSQueue
+from testfixtures import TempDirectory
+
+class FakeFlag():
```
giorgiosironi: `FakeFlag` and `FakeS3Event` seem useful in general for testing, so they could be extracted in some common test module later

queue_worker.py
```diff
@@ -22,89 +21,109 @@
...
+        self.sleep_seconds = 10
+
+    def create_log(self):
+        # Log
+        identity = "queue_worker_%s" % os.getpid()
+        log_file = "queue_worker.log"
+        # logFile = None
+        self.logger = log.logger(log_file, self.settings.setLevel, identity)
+
+    def connect(self):
```
giorgiosironi: this isolation makes `work()` much more agnostic wrt how it got a reference to a queue. If it was pushed into a `provider` module it could be even better as it could be reused across various processes?

queue_worker.py
gnott: I noticed it too before, and I wasn't sure. I couldn't find any usage of it here and I think it should be safe to remove. I'll remove it in this PR, and if anything is wrong we should notice it on end2end testing.

tests/test_queue_worker.py
gnott: Moving them to a central file is an improvement for sure. I will move them to ``tests/classes_mock.py`` for potential re-use.

queue_worker.py
gnott: Were you thinking of a provider which specifically holds the SQS queue connection and get queue logic? (a little like the ``storage_provider.py`` or the older ``swfmeta.py`` in how they manage the connection). Activities (I find four which connect to SQS) and the utilities could then use it.

Or, were you thinking more of the logic in ``work()`` about how it reads messages from the queue could also be in a provider? ``lax_response_adapter.py`` has similar logic around the ``while flag.green()`` block.

The provider could perhaps be included in this block of work, although its original purpose was just to add 3 lines to the ``newFileWorkflows.yaml`` file to enable silent corrections S3 notifications, and then it blossomed into writing tests and refactoring.

giorgiosironi: Not sure about where to draw the line, I'd tend to go with "what can reasonably be reused by other activities". Not needed now, it's a possible future improvement.

> and then it blossomed into writing tests and refactoring.

Software development done right :)

PR #678, opened 2018-04-03
Clean the temp directory of files in PMCDeposit always.

gnott:
Issues after many silent correction workflows were run, the PMC deposits when successful leave their large files in the activity temporary directory. If too many are run at once, the disk fills to 100% and causes lots of problems.


gnott:
Please feel free to merge and deploy if it looks ok, or review as you please.

gnott:
The comments are not necessarily incorrect, also not very enlightening, and are probably just leftovers from building it up. Thanks for the review!

PR #681, opened 2018-04-06
Explicitly do not send the publication email for retraction articles.

gnott:
Production asked whether authors would be emailed upon the publication of a retraction article. 

Logs show for a retraction the ``PublicationEmail`` activity was unable to find any authors from its data sources, and wouldn't send it. As the code is written, I'm fairly certain the logic would be unable to find which email template to use and would not send emails for retractions anyway.

That said, here is a more explicit block to not email authors about retraction articles by adding it to the other types we do not send for (``editorial`` and ``correction``).

I added a test specifically for the ``approve_articles()`` function which processes some basic article objects to test the logic.

I will be creating a new issue to describe a more elaborate refactoring for the types of articles and possibly the template types to be moved to a configuration file rather than having it in code, which should make it easier for other users to adapt it to their own preferences for email communication.

### DIFF
--- a/tests/activity/test_activity_publication_email.py
+++ b/tests/activity/test_activity_publication_email.py
@@ -18,6 +18,7 @@ from classes_mock import FakeKey
 
 from testfixtures import TempDirectory
 import tests.test_data as test_data
+from tests.activity.helpers import instantiate_article
 
 import os
 # Add parent directory for imports, so activity classes can use elife-poa-xml-generation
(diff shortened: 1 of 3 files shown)


tests/activity/test_activity_publication_email.py
```diff
@@ -372,6 +372,31 @@ def test_send_email_bad_authors(self, fake_queue_author_email):
...
+        article_object = article()
+        article_object.article_type = article_type
+        article_object.doi = doi
+        article_object.doi_id = article_object.get_doi_id(doi)
+        article_object.is_poa = is_poa
+        article_object.was_ever_poa = was_ever_poa
+        return article_object
+
+    @patch('provider.lax_provider.article_versions')
+    def test_approve_articles(self, mock_lax_provider_article_versions):
```
giorgiosironi: I think the docstring-like "test removing articles based on article type" should become the test name `test_removes_articles_based_on_article_type`

tests/activity/test_activity_publication_email.py
```diff
@@ -372,6 +372,31 @@ def test_send_email_bad_authors(self, fake_queue_author_email):
             result = self.activity.send_email(None, None, failed_author, None, None)
             self.assertEqual(result, False)
 
+    def instantiate_article(self, article_type, doi, is_poa=None, was_ever_poa=None):
```
giorgiosironi: not sure if reusable outside of this test class

tests/activity/test_activity_publication_email.py
gnott: Sure, I like that. It is probably close to exceeding some PEP recommended function name length, although their importance varies :)

tests/activity/test_activity_publication_email.py
gnott: It could be useful elsewhere, I'm not sure yet, but I'll move it to a more central importable place -good idea, thanks @giorgiosironi !

gnott: The most appropriate place for functions it seems will be in ``tests.activity.helpers`` and where I am moving it to.

PR #709, opened 2018-05-24
ELPP-3522 Fix article_id type in invalidation

giorgiosironi:
Overlooked that it is a string, not a number

### DIFF
--- a/provider/fastly_provider.py
+++ b/provider/fastly_provider.py
@@ -1,7 +1,7 @@
 import requests
 
 def purge(article_id, settings):
-    surrogate_key = 'articles/%05d' % article_id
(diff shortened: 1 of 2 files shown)


giorgiosironi:
This was just a quick fix, but if you see an obvious refactor that reuses this pattern from elsewhere in the bot, please suggest.

gnott:
``pad_msid()`` in https://github.com/elifesciences/elife-bot/blob/develop/provider/utils.py#L6 is the spot right now, but it is not used throughout. We have a variety of formatting logic still and one day maybe we will go through with an aim to refactor specific journal-specific logic.

PR #711, opened 2018-05-25
ELPP-3522 Remove CloudFront from InvalidateCdn activity

giorgiosironi:
These CDNs are not used anymore in favor of the Fastly ones, so invalidation can be faster.

Incidentally I also found some settings that are unused (as far as I can tell from looking at the code) and that relate to CloudFront - better to remove them altogether to avoid confusion.


activity/activity_InvalidateCdn.py
```diff
@@ -30,38 +29,10 @@ def do_activity(self, data):
             return activity.activity.ACTIVITY_PERMANENT_FAILURE
 
         try:
-
-            self.emit_monitor_event(self.settings, article_id, version, run,
```
gnott: At the start of the activity, it should still have a ``self.emit_monitor_event()`` with a ``"start"`` message to the dashboard, which is later closed with the ``"end"`` message you have on line 36.

settings-example.py
```diff
@@ -71,15 +71,8 @@ class exp():
...
     ses_admin_email = "admin@example.com"
 
-    # CDN bucket settings
-    cdn_bucket = 'elife-cdn-dev'
-    cdn_distribution_id = u'E1HPZ2QWOYE9NX'
-    cdn_domain_name = 'dhkzd83nokruv.cloudfront.net'
-
     # Lens bucket settings
     lens_bucket = 'elife-lens-dev'
-    lens_distribution_id = u'E30WWCB2DNEOKI'
```
gnott: It's probably ok to drop the Lens CDN settings, which is a separate CDN, but I don't think we implemented invalidating the content. We do invalidate the XML file Lens loads (which is stored in the primary CDN) as well as the figures, which I think results in an updated Lens page. Lens invalidations would only matter if we changed the Lens template, or the Lens source files (``.js`` and ``.css`` files).

giorgiosironi: Invalidation is only an optimization as the caching of the files will still expire after the default TTL. The invalidation implemented here was for the production team wanting to get fast feedback from the preview site about a new ingestion.

activity/activity_InvalidateCdn.py
giorgiosironi: Good catch, I think the whole `try` block got removed in a single commit

giorgiosironi: It also looks like the job of a context manager, if I get time I'll do a follow up PR

PR #712, opened 2018-05-30
ELPP-3665 Refine surrogate keys for cache purging

giorgiosironi:
Will involves `iiif.elifesciences.org` and `cdn.elifesciences.org`.

In both cases the two cache keys correspond to versioned files and to the unversioned videos (and their stills), see https://github.com/elifesciences/builder/pull/328 for details.

### DIFF
--- a/provider/fastly_provider.py
+++ b/provider/fastly_provider.py
@@ -1,10 +1,10 @@
 import requests
 
-class Purge:
+class FastlyApi:
     def __init__(self, fastly_api_key):
         self._fastly_api_key = fastly_api_key
 
-    def of_key(self, surrogate_key, service_id):
+    def purge(self, surrogate_key, service_id):
         url = "https://api.fastly.com/service/%s/purge/%s" % (service_id, surrogate_key)
 
         response = requests.post(url, headers={
@@ -17,8 +17,11 @@ class Purge:
 KEYS = ['articles/{article_id}v{version}', 'articles/{article_id}/videos']
 
 def purge(article_id, version, settings):
-    p = Purge(settings.fastly_api_key)
+    responses = []
+    api = FastlyApi(settings.fastly_api_key)
     for service_id in settings.fastly_service_ids:
         for key in KEYS:
             surrogate_key = key.format(article_id=article_id.zfill(5), version=version)
-            p.of_key(surrogate_key, service_id)
+            responses.append(api.purge(surrogate_key, service_id))
+
+    return responses
--- a/provider/fastly_provider.py
+++ b/provider/fastly_provider.py
@@ -1,12 +1,24 @@
 import requests
 
-def purge(article_id, settings):
-    surrogate_key = 'articles/%s' % article_id.zfill(5)
-    url = "https://api.fastly.com/service/%s/purge/%s" % (settings.fastly_service_id, surrogate_key)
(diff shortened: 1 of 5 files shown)


giorgiosironi:
Configuration added in https://github.com/elifesciences/builder-private/commit/8f0b99385444002204f30ad260743a07e8b1a182

tests/provider/test_fastly_provider.py
```diff
@@ -8,14 +8,35 @@
 class TestFastlyProvider(unittest.TestCase):
     @patch('requests.post')
     def test_purge(self, post_mock):
-        response = Response()
-        response.status_code = 200
-        post_mock.return_value = response
-        fastly_provider.purge('10627', settings_mock)
+        self._allow_call(post_mock)
+        fastly_provider.purge('10627', '1', settings_mock)
```
gnott: Is there any assertion possible here?

tests/provider/test_fastly_provider.py
```diff
@@ -8,14 +8,35 @@
...
+
+    @patch('requests.post')
+    def test_purge_various_keys(self, post_mock):
+        self._allow_call(post_mock)
+        fastly_provider.purge('10627', '1', settings_mock)
```
gnott: Does this return a list of responses that could be looked at, rather than going through the mock object? Probably the same effect although it would follow the provider's return type (maybe).

provider/fastly_provider.py
```diff
@@ -1,12 +1,27 @@
...
+        })
+        response.raise_for_status()
+        return response
+
+KEYS = ['articles/{article_id}v{version}', 'articles/{article_id}/videos']
+
+def purge(article_id, version, settings):
+    responses = []
+    api = FastlyApi(settings.fastly_api_key)
+    for service_id in settings.fastly_service_ids:
```
gnott: I'm not sure I completely understand this part, my question is would we potentially have multiple ``service_ids``, and the surrogate keys are only applicable to one service_id, then here we might be trying to purge each surrogate key on every service_id?

tests/provider/test_fastly_provider.py
giorgiosironi: I dedicated the two following tests, `test_purge_various_keys` and `test_purge_return_responses`, to check the HTTP calls and the return value respectively. This first one may be redundant at this point, but it documents usage (call it, fire-and-forget).

tests/provider/test_fastly_provider.py
giorgiosironi: It does return responses, see `test_purge_return_responses`; the mock is checking the requests instead

provider/fastly_provider.py
giorgiosironi: Yes, we have two service ids now (`[prod--cdn-]iiif.elifesciences.org` and `cdn.elifesciences.org`), and both implement the two key patterns used here:
https://github.com/elifesciences/builder/blob/master/projects/elife.yaml#L760-L784
https://github.com/elifesciences/builder/blob/master/projects/elife.yaml#L1552-L1567
Each of the 4 calls should be in the order of 100-200 milliseconds. If we were to add even more CDNs with different surrogate keys, we could then optimize this to only make certain calls depending on what the service is.

PR #717, opened 2018-06-21
Continue rather than return when processing unknown S3 notifications

gnott:
Continue in ``queue_worker.py`` if an S3 notification cannot be processed rather than returning ``False``. It already writes to the log if the S3 notification cannot be processed. Issuing ``continue`` will send it back to ``while flag.green():`` and continue reading messages from the queue.

### DIFF
--- a/queue_worker.py
+++ b/queue_worker.py
@@ -81,19 +81,18 @@ class QueueWorker:
                             self.logger.info("S3NotificationInfo: %s", info.to_dict())
                             workflow_name = self.get_starter_name(rules, info)
                             if workflow_name is None:
-                                self.logger.info("Could not handle file %s in bucket %s" % (info.file_name, info.bucket_name))
-                                continue
-
-                            # build message
-                            message = {
-                                'workflow_name': workflow_name,
-                                'workflow_data': info.to_dict()
-                            }
-
-                            # send workflow initiation message
-                            m = Message()
-                            m.set_body(json.dumps(message))
-                            out_queue.write(m)
+                                self.logger.error("Could not handle file %s in bucket %s" % (info.file_name, info.bucket_name))
+                            else:
+                                # build message
+                                message = {
+                                    'workflow_name': workflow_name,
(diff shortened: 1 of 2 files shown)


queue_worker.py
```diff
@@ -82,7 +82,7 @@ def work(self, flag):
                             workflow_name = self.get_starter_name(rules, info)
                             if workflow_name is None:
                                 self.logger.info("Could not handle file %s in bucket %s" % (info.file_name, info.bucket_name))
```
giorgiosironi: this look more like an `error` than `info` now?

giorgiosironi: `queue.delete_message(queue_message)` is missing for this use case I think, which would see the message eventually time out and return to the queue. I think the existing `delete_message` call can be moved to a `finally` block placed inside the `while flag.green()` loop (as in `try...finally`) so that it it always executed?

gnott: Very perceptive suggestion, I only attempted to avoid the utility crashing when it encountered an unexpected message. My next commit we can avoid the ``continue`` and just delete each message as it is processed.

If the queue gets a non-``S3Event`` message I think it may also loop forever, although it is less likely at the time being.

It could be a little more resilient if we changed it to reading only a single message from the queue with ``queue.read()`` to instead ``queue.get_messages()`` that would return more than one message. We could then leave bad messages in the queue in flight until someone could troubleshoot. If everything is configured properly we shouldn't get messages that cannot be processed, and we can check the log for some evidence of poor queue reading.

queue_worker.py
```diff
@@ -81,19 +81,18 @@ def work(self, flag):
                             self.logger.info("S3NotificationInfo: %s", info.to_dict())
```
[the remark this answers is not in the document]
gnott: Spot on, the log should be suitable for troubleshooting for now.

gnott:
Merging now, expecting all automated tests will have no errors.

PR #719, opened 2018-06-27
Add publishing_buckets_prefix to the bucket names in PoA activities

gnott:
Add publishing_buckets_prefix to the bucket names in PoA activities for easier end2end testing.

Should have no effect on the ``prod`` environment.

### DIFF
--- a/activity/activity_PackagePOA.py
+++ b/activity/activity_PackagePOA.py
@@ -54,7 +54,7 @@ class activity_PackagePOA(activity.activity):
         self.db = dblib.SimpleDB(settings)
 
         # Bucket for outgoing files
-        self.publish_bucket = settings.poa_packaging_bucket
+        self.publish_bucket = settings.publishing_buckets_prefix + settings.poa_packaging_bucket
         self.outbox_folder = "outbox/"
 
         # Some values to set later
--- a/activity/activity_PublishFinalPOA.py
+++ b/activity/activity_PublishFinalPOA.py
@@ -50,7 +50,7 @@ class activity_PublishFinalPOA(activity.activity):
         self.DONE_DIR = self.get_tmp_dir() + os.sep + "done_dir"
 
         # Bucket for outgoing files
-        self.input_bucket = settings.poa_packaging_bucket
+        self.input_bucket = settings.publishing_buckets_prefix + settings.poa_packaging_bucket
         self.outbox_folder = "outbox/"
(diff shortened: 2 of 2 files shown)


giorgiosironi:
Have to review `elife-spectrum` and mostly `builder-private`, can get to this tomorrow.

activity/activity_PackagePOA.py
```diff
@@ -54,7 +54,7 @@ def __init__(self, settings, logger, conn=None, token=None, activity_task=None):
         self.db = dblib.SimpleDB(settings)
 
         # Bucket for outgoing files
-        self.publish_bucket = settings.poa_packaging_bucket
+        self.publish_bucket = settings.publishing_buckets_prefix + settings.poa_packaging_bucket
```
giorgiosironi: Only `end2end` uses `publishing_buckets_prefix`, but `continuumtest` and `prod` don't. I'd prefer to set it to empty in `end2end` too; this should result in using the `elife-poa-packaging-end2end` everywhere. Eventually we could remove this prefix if it's not used anymore? The name of buckets are varied and flexible so being explicit about their configuration makes it less difficult to predict where files will end up.

gnott: If I may describe what I understand - in the settings, the ``poa_packaging_bucket`` value would always indicate the precise bucket name for each environment, and the ``publishing_buckets_prefix`` would not be required. We would close this PR because it is not needed.

There is only one place I can see that uses the ``publishing_buckets_prefix`` for the ``poa_packaging_bucket``, and that is in https://github.com/elifesciences/elife-bot/blob/develop/activity/activity_PubmedArticleDeposit.py#L55   If we remove the usage of ``publishing_buckets_prefix`` there the I think all activities will use the correct bucket on all environments.

The builder configuration can follow the same naming logic, so there will be only one poa packaging bucket for each environment, where I think for end2end there are two buckets - one for Pubmed and one for all other activities (such as Crossref).

giorgiosironi: So I'll close this and open a PR with the suggested change to `PubmedArticleDeposit`?

gnott: I can do the PRs I opened this one to start with - coming up!

gnott:
Closing in favour of https://github.com/elifesciences/elife-bot/pull/722 is a better option.

PR #721, opened 2018-07-03
Expand logs of PackagePOA starter

giorgiosironi:
Helps with elifesciences/jira-import#1196 

Seems I have run into SimpleDB finally. Some logs will help understand what it is returning (probably nothing).


giorgiosironi:
> supply document as the PoA zip file you want to process.

Ok, that looks like a good trade-off.

PR #722, opened 2018-07-03
Remove the bucket prefix from the bucket name in PubmedArticleDeposit.

gnott:
In the ``PubmedArticleDeposit`` activity, removing the ``publishing_buckets_prefix`` from the bucket name will have no effect on the ``prod`` environment. All other environments will use the exact bucket name explicitly listed in ``settings.py`` for where it finds input and stores output.

### DIFF
--- a/activity/activity_PubmedArticleDeposit.py
+++ b/activity/activity_PubmedArticleDeposit.py
@@ -52,7 +52,7 @@ class activity_PubmedArticleDeposit(activity.activity):
         self.article = articlelib.article(self.settings, self.get_tmp_dir())
 
         # Bucket for outgoing files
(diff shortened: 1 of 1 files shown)


giorgiosironi:
I think I need to change https://github.com/elifesciences/builder-private/blob/master/pillar/elife.sls#L412 before merging so I'll do that and merge this right after to check they agree.

giorgiosironi:
Had to create the `elife-poa-packaging-end2end` bucket, which wasn't there. Retrying in https://alfred.elifesciences.org/job/test-elife-bot/485/ this should only affect the bot and not other projects.

giorgiosironi:
Also https://alfred.elifesciences.org/job/test-elife-bot/485/consoleFull failed, will try to reproduce manually.

PR #723, opened 2018-07-04
Write to un-prefixed pubmed outbox

giorgiosironi:
Follow-up to #722. Only affects `end2end` as an environment, but we have
to write the file where it is read by the other activity.

I've seen also `ScheduleCrossref` and we should discuss in general the
removal of the prefix, but this seems the minimal change to make the
test suite green.


giorgiosironi:
https://alfred.elifesciences.org/job/test-elife-bot/486/ is green now.

PR #725, opened 2018-07-09
Fix Fastly surrogate keys

thewilkybarkid:
To match https://github.com/elifesciences/builder/pull/362.

### DIFF
--- a/provider/fastly_provider.py
+++ b/provider/fastly_provider.py
@@ -10,11 +10,12 @@ class FastlyApi:
         response = requests.post(url, headers={
             'Accept': 'application/json',
             'Fastly-Key': self._fastly_api_key,
+            'Fastly-Soft-Purge': '1',
         })
         response.raise_for_status()
         return response
 
-KEYS = ['articles/{article_id}v{version}', 'articles/{article_id}/videos']
+KEYS = ['article/{article_id}v{version}', 'article/{article_id}/videos']
 
 def purge(article_id, version, settings):
     responses = []


provider/fastly_provider.py
```diff
@@ -10,11 +10,12 @@ def purge(self, surrogate_key, service_id):
         response = requests.post(url, headers={
             'Accept': 'application/json',
             'Fastly-Key': self._fastly_api_key,
+            'Fastly-Soft-Purge': '1',
```
thewilkybarkid: Unrelated, but better to use soft purges (causes revalidation rather than immediate expiry).

thewilkybarkid:
We could do this two stage, include both variants then a follow up removes `articles/`.

provider/fastly_provider.py
gnott: Seems ok, if what I understand is the origin file has an Etag (from S3), and somehow the user's browser will notice the stale content and request new files? The immediate purpose I think is for when Production previews an updated article then the correct figures appear.

thewilkybarkid: This isn't going to affect users' browsers, they'll still have to do a hard refresh to see the change (only way we could do that is to change the identifier if the image changes, or don't serve preview images through a cache).

This will just mean if Fastly can't revalidate (eg fails to connect to S3) it will serve the stale image rather than breaking (once https://github.com/elifesciences/jira-import/issues/4305 is done). Since it affects live content, it's a bit safer.

PR #734, opened 2018-07-17
Add logs of PackagePOA activity

giorgiosironi:
Helps with https://github.com/elifesciences/elife-spectrum/pull/119


giorgiosironi:
Makes sure that the activity completed correctly, but unfortunately doesn't go into the detail of bad data causing the XML generation to fail.

PR #735, opened 2018-07-17
Remove Crossref activity reference to bucket prefix

giorgiosironi:
Yet another place in which the prefix can go. The corresponding `DepositCrossref` activity seems already not to use it.

Spawned by seeing:

```
$ aws s3 ls end2end-elife-poa-packaging-end2end/
PRE crossref/
```

I should be able to delete that bucket after this change, in favor of
`elife-poa-packaging-end2end`.


gnott:
Looks good, small, should be all compatible with the prod environment. I'll merge it now, thanks @giorgiosironi!

PR #738, opened 2018-07-18
Switch back to python 2 compatible only StringIO.

gnott:
The Alfred build is not completing, I don't think this is the issue but I'll give it a try. This is basically reversing the Python 3 compatible attempt where ``StringIO`` was used, back to the original. If it builds we can run with this until there is time to debug the build.

### DIFF
--- a/provider/imageresize.py
+++ b/provider/imageresize.py
@@ -1,4 +1,4 @@
-from io import BytesIO
+from StringIO import StringIO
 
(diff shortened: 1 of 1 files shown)


de-code:
Is there an issue with what failed?

gnott:
I think it was a build infrastructure issue and not a ``BytesIO`` one. I will try to reintroduce this soon.

PR #742, opened 2018-07-28
Python 3 support for glencoe bits

gnott:
I did this a while ago, I've merged in the latest develop branch and test are still passing. Hopefully looks good. These are minor changes again just to get Python 3 compatibility.

### DIFF
--- a/activity/activity_CopyGlencoeStillImages.py
+++ b/activity/activity_CopyGlencoeStillImages.py
@@ -1,4 +1,4 @@
-import activity
+from .activity import Activity
 import json
 from provider.execution_context import get_session
 from provider.storage_provider import storage_context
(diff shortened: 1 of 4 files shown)


activity/activity_CopyGlencoeStillImages.py
```diff
@@ -1,4 +1,4 @@
-import activity
+from .activity import Activity
```
giorgiosironi: Do we have a policy on relative vs. absolute imports @seanwiseman?

seanwiseman: Not that I am aware of. It's a split argument, as long as it is consistent across the project itself I don't normally comment.

My personal preference is to be as absolute as possible up to the containing package.

PR #744, opened 2018-07-30
Make resize memory usage measurement reproducible

giorgiosironi:
Have to commit a image here for ease of use. Closes https://github.com/elifesciences/jira-import/issues/4353

Sample output:
```
/home/giorgio/code/elife-bot/venv/local/lib/python2.7/site-packages/wand/image.py:2548: CoderWarning: Incompatible type for "RichTIFFIPTC"; tag ignored. `TIFFFetchNormalTag' @ warning/tiff.c/TIFFWarnings/881
  self.raise_exception()
Before: 17477632
After: 27660288
Memory delta: 10182656
```

For some reason have to run it as `PYTHONPATH=. python scripts/resize_memory.py` otherwise I get 
```
  File "scripts/resize_memory.py", line 2, in <module>
    from provider import imageresize, memory
ImportError: No module named provider
```
possibly because the `scripts/` folder is picked up as the `PYTHONPATH`?


seanwiseman:
> possibly because the scripts/ folder is picked up as the PYTHONPATH?

When executing `python scripts/resize_memory.py` yes the python path is all relative to the execution location of `resize_memory.py` so the `provider` package does not exist as far as its concerned. If the root project was itself a package then you could have got around this with this for example:

`from bot.provider import imageresize, memory`

but obviously, it's not 😜 , so `PYTHONPATH=. python scripts/resize_memory.py` is an acceptable way to get around executing this script.

gnott:
Would it fix the imports if there was a ``__init__.py`` file added to the root folder of this project?

seanwiseman:
> Would it fix the imports if there was a __init__.py file added to the root folder of this project?

Yes, but not sure this one case is worth it? meaning I would think some cleaning up / moving of non-python files from the newly created package would be necessary. 

Though in theory, yes you could throw one in there and it would work.

giorgiosironi:
Can we merge to close https://github.com/elifesciences/jira-import/issues/4353?

gnott:
Sure! I will merge now, thanks @giorgiosironi

PR #745, opened 2018-07-30
Load the digest.cfg configuration in EmailDigest activity

gnott:
Load the digest.cfg configuration in EmailDigest activity, and tests for the output_file_name.

Should not be merged or deployed before the settings changes of PR https://github.com/elifesciences/builder-private/pull/210 are merged.

Makes it so the ``.docx`` file name is set by the configuration file, in case someone does specify it there. The config file will be more important in other activities than in ``EmailDigest`` particularly.

### DIFF
--- a/activity/activity_EmailDigest.py
+++ b/activity/activity_EmailDigest.py
@@ -4,6 +4,7 @@ import time
 import traceback
 import boto.swf
 from digestparser import build, output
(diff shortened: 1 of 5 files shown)


gnott:
It looks like ``ci`` tests will not run without the new settings too, ... I think I'll just merge them without review and run the tests again.

PR #746, opened 2018-08-01
Upload digest image to the S3 bucket

gnott:
Fixes https://github.com/elifesciences/jira-import/issues/4299

Primary addition here is the new ``DepositDigestIngestAssets`` activity.

To avoid code duplication between it and the ``EmailDigest`` activity, I created a ``digest_parser.py`` provider to hold some of the common downloading, parsing and building functions.

The final thing I noticed was the bucket name in the settings has ``/article`` on it which is unsuitable for uploading digest files to ``/digests``, so I added a new setting value for it in PR https://github.com/elifesciences/builder-private/pull/214, which would need to be merged before this PR.

For your review and eventual approval, then we should have digest images from the IIIF server. Also maybe room for enhanced end2end testing for this new image?


giorgiosironi:
> room for enhanced end2end testing for this new image?

Yes, definitely, just wanted to get a clear picture of the input, going through the code

activity/activity_DepositDigestIngestAssets.py
```diff
@@ -0,0 +1,104 @@
...
+        if not self.build_status:
+            self.logger.info("Failed to build the Digest in Deposit Digest Ingest Assets for %s",
+                             real_filename)
+            return self.ACTIVITY_PERMANENT_FAILURE
+
+        # check if there is an image and if not return True
+        if not digest_provider.has_image(self.digest):
+            self.logger.info("Digest for file %s has no images to deposit",
+                             real_filename)
+            return self.ACTIVITY_SUCCESS
```
giorgiosironi: I missed part of this discussion; [`image` seems required in the API](https://github.com/elifesciences/api-raml/blob/develop/src/snippets/digest.v1.yaml#L41) so where it should come from if it's not present in the input?

provider/digest_provider.py
```diff
@@ -0,0 +1,80 @@
...
+    real_filename = utils.unquote_plus(filename)
+    return real_filename, bucket_name, bucket_folder
+
+
+def build_digest(input_file, temp_dir, logger=None):
+    "Parse input and build a Digest object"
+    if not input_file:
+        return False, None
+    try:
+        digest = build.build_digest(input_file, temp_dir)
```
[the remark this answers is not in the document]
gnott: For now, the image will be mandatory so every digest will be supplied as a ``.zip``. There's sample data here https://github.com/elifesciences/digest-parser/blob/develop/tests/test_data/DIGEST%2099999.zip (just a fake image). Any image should do. There doesn't seem to be any stipulation on the image name itself that I'm aware of.

provider/digest_provider.py
```diff
@@ -0,0 +1,80 @@
...
+import traceback
+from docx.opc.exceptions import PackageNotFoundError
+from digestparser import build
+from S3utility.s3_notification_info import S3NotificationInfo
+from provider.storage_provider import storage_context
+import provider.utils as utils
+
+
+def parse_data(data):
+    "parse activity data from an S3 notification into useful bucket and file values"
```
giorgiosironi: this function is different as ti doesn't go through `storage_context` which should hide S3; it could sit well in there? It doesn't seem specific to digests

settings-example.py
```diff
@@ -41,6 +41,7 @@ class exp():
     production_bucket = 'elife-production-final'
     expanded_bucket = 'elife-publishing-expanded'
     ppp_cdn_bucket = 'elife-publishing-cdn'
```
giorgiosironi: these bucket names (just examples) can be updated to be more similar to the real ones e.g. `test-elife-published/digests`

tests/activity/test_activity_deposit_digest_ingest_assets.py
```diff
@@ -0,0 +1,102 @@
...
+        # check digest image values
+        if (
+                self.activity.digest and self.activity.digest.image and
+                test_data.get("expected_digest_image_file")):
+            file_name = self.activity.digest.image.file.split(os.sep)[-1]
+            self.assertEqual(file_name, test_data.get("expected_digest_image_file"),
+                             'failed in {comment}'.format(comment=test_data.get("comment")))
+
+        # Check the S3 object destination resource
+        if 'expected_dest_resource' in test_data:
```
giorgiosironi: `if` is suspicious in tests, although it is to support more combinations of inputs. Maybe splitting the failure results from the activity results would lead to two better focused tests that don't need as many conditionals?

provider/digest_provider.py
gnott: I agree, this function is more general purpose for parsing S3 notifications, and it can be moved to somewhere more logical. I think a good place for it is in https://github.com/elifesciences/elife-bot/blob/develop/S3utility/s3_notification_info.py, since that is where ``S3NotificationInfo.from_dict()`` is.

settings-example.py
gnott: I will compare these and also look for more example settings file changes to match the current settings file.

tests/activity/test_activity_deposit_digest_ingest_assets.py
gnott: The ``if`` statements made more sense in the tests for ``EmailDigest`` I think, for testing email contents. I can probably clarify the test scenario data to not requires ``if``s here.

activity/activity_DepositDigestIngestAssets.py
giorgiosironi: Talked with Stuart, support for a .docx will be used in the future for digests that may not require an image, but there are no designs that allow that for the moment.

activity/activity_DepositDigestIngestAssets.py
```diff
@@ -0,0 +1,105 @@
...
+
+        # Track the success of some steps
+        self.build_status = None
+
+    def do_activity(self, data=None):
+        "do the work"
+        if self.logger:
+            self.logger.info('data: %s' % json.dumps(data, sort_keys=True, indent=4))
+        # parse the data with the digest_provider
+        real_filename, bucket_name, bucket_folder = parse_activity_data(data)
```
giorgiosironi: a smell to think about for the future: `parse_activity_data` makes the activity depend on an S3 notification rather than just on S3 as storage. Means that if you want to re-trigger the workflow, you have to fake the S3 notification building more data than needed. Possibly that could be dealt with in the machinery that triggers the worfklow (`starter_*` modules?) rather than inside it.

Not a blocker.

workflow/workflow_IngestDigest.py
```diff
@@ -55,7 +55,18 @@ def __init__(self, settings, logger, conn=None, token=None, decision=None,
                         "schedule_to_close_timeout": 300,
                         "schedule_to_start_timeout": 300,
                         "start_to_close_timeout": 300
-                    }
+                    },
+                    {
+                        "activity_type": "DepositDigestIngestAssets",
```
giorgiosironi: I like the consistency with other `Deposit` activities

activity/activity_DepositDigestIngestAssets.py
gnott: There are two or three workflow starters that require an S3 notificatoin to start them. For those, we use the ``econ-tools`` feeder to create them without dropping a file again. There is room for some refactoring and standardising of how the starters prepare activity data so they're always consistent. When you view it, it is a fairly simple set of bucket and file attributes, that can be de-S3-i-fied.

gnott:
Thanks for the reviewing, with all the refactoring and building pretty fast, it is a little rough - but runs (I think). I will merge and then maybe do a test when ready.

Issue #683, opened 2018-04-11
Track whether is a silent correction or not in the workflow session


gnott:
The ``PublicationEmail`` activity sends emails to authors when an article is published. It has duplicate email checking logic, so a recipient would never receive the same email twice. 

However, as a side effect of some silent corrections, occasionally an email is sent to one author. I think this happens if an author adds or changes their email address since the last time the emails were sent. Since this new email address is not in the duplicate email checking log, then the activity sends an email to the author.

One suggestion for an improvement could be in the ``ScheduleDownstream`` activity - which is part of the ``PostPerfectPublication`` workflow - to not queue the article for ``PublicationEmail`` if the article supplied was a silent correction. Currently the ``PostPerfectPublication`` workflow is not aware whether it was part of a silent correction or a normal publication event. If we know it was a result of a silent correction then we can modify its logic accordingly. It could get the value from the workflow session data.

There are a couple potential markers for a silent correction in the data supplied at the start of a ``SilentCorrectionIngest`` workflow (https://github.com/elifesciences/elife-bot/blob/develop/starter/starter_SilentCorrectionsIngest.py#L32)

```
        input['version_lookup_function'] = "article_highest_version"
        input['force'] = True
```

The values of both ``version_lookup_function`` and ``force`` indicate a silent correction, although we do not have a specific value to represent a silent correction, for example if we had ``silent_correction = True`` it would be explicit. The two values above are also not stored in the session yet anyway, although they are passed between Lax and the publication approval (I think) we could store them in the session from the beginning.

Should we track the silent correction status in the workflow session somewhere? And/or, should we track the ``version_lookup_function`` and ``force`` values and then use those as proxies for whether it was part of a silent correction? Some input from @giorgiosironi, @seanwiseman, @Jenniferstrej or @jhroot  would be much appreciated - thanks!

giorgiosironi:
>  if we had silent_correction = True it would be explicit

:+1: for being explicit - the `version_lookup_function` values or even that parameter's name may change, but a ~~diamond~~ specific configuration is forever

> The two values above are also not stored in the session yet anyway, although they are passed between Lax and the publication approval (I think) we could store them in the session from the beginning.

It looks like you need to store in the session this new `silent_correction` so that it is persisted along the whole run even for more than one workflow; if that's the case, no problem with that.

To make this a bit more future-proof, rather than a boolean parameter a `run_type=silent_correction` could be used. That leaves space for other possible `run_type`s to be introduced.

[...]

PR #748, opened 2018-08-03
Session run_type

gnott:
Fixes https://github.com/elifesciences/elife-bot/issues/683

In ``starter_SilentCorrectionsIngest.py``, add a ``run_type`` to the data. Then, in ``activity_VersionLookup.py`` (the first activity run in the workflow) save it into the session.

Tests check the session for the value. It should be compatible if data is missing the value too, where ``run_type`` will be ``None``.

I went ahead and did some code linting and also made the activity Python 3 compatible while I'm at it.

We'll want this for digest processing for cases when it is part of a silent correction.


starter/starter_SilentCorrectionsIngest.py
```diff
@@ -30,6 +30,7 @@ def start(self, settings, info=None, run=None):
         input = S3NotificationInfo.to_dict(info)
         input['run'] = run
         input['version_lookup_function'] = "article_highest_version"
+        input['run_type'] = "silent_correction"
```
giorgiosironi: format of this: `silent_correction` or `silent-correction`? Any other value this should be consistent with? For example the run id is formatted with `-`, e.g. `8655c526-66fa-4e49-9ca1-b7569aeb32dc`

activity/activity_VersionLookup.py
```diff
@@ -31,46 +34,56 @@ def do_activity(self, data=None):
             run = data['run']
             session = get_session(self.settings, data, run)
             session.store_value('filename_last_element', filename)
+            session.store_value('run_type', data.get('run_type'))
```
giorgiosironi: I guess the intention is to also populate other `run_type`s later like `first-publication` (or some other non-silent-correction name)

starter/starter_SilentCorrectionsIngest.py
gnott: Run id values are a ``uuid`` which uses hyphens, evidently. We could go with ``SilentCorrections`` if we matched the workflow class names. I cannot find any standard to say what can or cannot be a value in a session attribute.

There are not many underscores though, so I will change it to a hyphen.

activity/activity_VersionLookup.py
gnott: ``first-publication`` might not fit in the ``run_type``, it would more likely be something like ``ingest-article`` for non-``silent-correction`` workflows. 

``run_type`` is just so when multiple workflows are linked together, we can know which "ingest" workflow started it all (since the same ``PostPerfectPublication`` workflow is run as a result of a silent or non-silent correction, we cannot choose what steps to do when we get there if we do not know what was done earlier in the ``run``).

The version status of the article can be determined separately at the time. So if we only want to do something for the latest version, we can query Lax what is the highest published version and then compare to the version number we are processing, and if they're equal then we are processing the most recent version.

giorgiosironi: Yes `first-publication` isn't a good name. Very useful to see the overarching story of the run.

PR #750, opened 2018-08-06
Copy digest to outbox

gnott:
Fixes https://github.com/elifesciences/issues/issues/4300
Fixes https://github.com/elifesciences/issues/issues/4302

A new ``CopyDigestToOutbox`` activity, complets the ``IngestDigest`` workflow.

Hopefully the time spent trying to read from the bucket, delete files, and upload files, all mocked using the test objects, will be useful because it was not pleasant to figure it all out. Not tested for real yet either, so hopefully the tests represent the real state of the buckets.

The ``.docx`` file is renamed as suggested in issue https://github.com/elifesciences/issues/issues/4300. The particular issue can probably be closed, since all the unknowns about file locations are filled in, though it is providing some details for later activities to be built that will read from the bucket the files kept in stasis.

Feedback and edits are welcome - thanks!

### DIFF
--- a/activity/activity_CopyDigestToOutbox.py
+++ b/activity/activity_CopyDigestToOutbox.py
@@ -28,10 +28,6 @@ class activity_CopyDigestToOutbox(Activity):
         self.description = ("Copies the Digest files to a bucket folder for later use" +
                             " in article ingestion")
 
-        # Track some values
-        self.input_file = None
-        self.digest = None
-
         # Local directory settings
         self.temp_dir = os.path.join(self.get_tmp_dir(), "tmp_dir")
         self.input_dir = os.path.join(self.get_tmp_dir(), "input_dir")
@@ -39,22 +35,19 @@ class activity_CopyDigestToOutbox(Activity):
         # Create output directories
         self.create_activity_directories()
 
-        # Track the success of some steps
-        self.build_status = None
-
     def do_activity(self, data=None):
         if self.logger:
             self.logger.info('data: %s' % json.dumps(data, sort_keys=True, indent=4))
         # parse the data with the digest_provider
         real_filename, bucket_name, bucket_folder = parse_activity_data(data)
         # Download from S3
-        self.input_file = digest_provider.download_digest_from_s3(
+        input_file = digest_provider.download_digest_from_s3(
             self.settings, real_filename, bucket_name, bucket_folder, self.input_dir)
         # Parse input and build digest
-        self.build_status, self.digest = digest_provider.build_digest(
-            self.input_file, self.temp_dir, self.logger)
+        build_status, digest = digest_provider.build_digest(
+            input_file, self.temp_dir, self.logger)
 
-        if not self.build_status:
+        if not build_status:
             self.logger.info("Failed to build the Digest in Copy Digest To Outbox for %s",
                              real_filename)
             return self.ACTIVITY_PERMANENT_FAILURE
@@ -65,14 +58,14 @@ class activity_CopyDigestToOutbox(Activity):
         bucket_name = self.settings.bot_bucket
 
         # clean out the outbox if not empty
-        self.clean_outbox(self.digest, bucket_name)
+        self.clean_outbox(digest, bucket_name)
 
         # copy the files to S3
         # if it is zip file take the files from the temp_dir, otherwise from the input_dir
         from_dir = self.input_dir
-        if self.input_file.endswith('.zip'):
+        if input_file.endswith('.zip'):
             from_dir = self.temp_dir
-        self.copy_files_to_outbox(self.digest, bucket_name, from_dir)
+        self.copy_files_to_outbox(digest, bucket_name, from_dir)
 
         return self.ACTIVITY_SUCCESS
 
--- a/activity/activity_CopyDigestToOutbox.py
+++ b/activity/activity_CopyDigestToOutbox.py
@@ -54,9 +54,27 @@ class activity_CopyDigestToOutbox(Activity):
                              real_filename)
             return self.ACTIVITY_PERMANENT_FAILURE
 
-        # todo!!  the rest of the logic
+        # continue with copying files now
+
+        # bucket name
+        bucket_name = self.settings.bot_bucket
+
 
         return self.ACTIVITY_SUCCESS
+
+    def dest_resource_path(self, digest, bucket_name):
+        "the bucket folder where files will be saved"
+        msid = utils.msid_from_doi(digest.doi)
+        article_id = utils.pad_msid(msid)
+        storage_provider = self.settings.storage_provider + "://"
+        return storage_provider + bucket_name + "/digests/outbox/" + article_id + "/"
+
+    def file_dest_resource(self, digest, bucket_name, file_path):
+        "concatenate the S3 bucket object path we copy the file to"
+        resource_path = self.dest_resource_path(digest, bucket_name)
+        file_name = file_path.split(os.sep)[-1]
+        dest_resource = resource_path + file_name
+        return dest_resource
 
     def create_activity_directories(self):
         """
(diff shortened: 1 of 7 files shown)


tests/activity/test_activity_copy_digest_to_outbox.py
```diff
@@ -0,0 +1,125 @@
...
+    def tearDown(self):
+        # clean the temporary directory
+        self.activity.clean_tmp_dir()
+        # clean out the bucket destination folder
+        helpers.delete_files_in_folder(testdata.ExpandArticle_files_dest_folder,
+                                       filter_out=['.gitkeep'])
+
+    @patch.object(digest_provider, 'storage_context')
+    @patch('activity.activity_CopyDigestToOutbox.storage_context')
+    @data(
```
giorgiosironi: these multiple instances of the same test seem to explain well what are the possible scenarios :+1:

provider/digest_provider.py
```diff
@@ -21,6 +21,13 @@ def build_digest(input_file, temp_dir, logger=None):
     return True, digest
 
 
+def new_file_name(file_name, msid):
+    "new name for the files stored for internal use"
+    if file_name and file_name.endswith('.docx'):
+        return 'digest-{msid:0>5}.docx'.format(msid=msid)
+    return file_name
```
giorgiosironi: given the image is going out on the CDN, I suppose it would be consistent to adopt the similar naming style `digest-12345.jpg`? (or whatever other file extension is used)

activity/activity_CopyDigestToOutbox.py
```diff
@@ -0,0 +1,121 @@
...
+            settings, logger, conn, token, activity_task)
+
+        self.name = "CopyDigestToOutbox"
+        self.pretty_name = "Copy Digest files to an outbox"
+        self.version = "1"
+        self.default_task_heartbeat_timeout = 30
+        self.default_task_schedule_to_close_timeout = 60 * 5
+        self.default_task_schedule_to_start_timeout = 30
+        self.default_task_start_to_close_timeout = 60 * 5
+        self.description = "Copies the Digest files to a bucket folder for later use"
```
giorgiosironi: `for later use in article ingestion`?

activity/activity_CopyDigestToOutbox.py
```diff
@@ -0,0 +1,121 @@
...
+        # Track some values
+        self.input_file = None
+        self.digest = None
+
+        # Local directory settings
+        self.temp_dir = os.path.join(self.get_tmp_dir(), "tmp_dir")
+        self.input_dir = os.path.join(self.get_tmp_dir(), "input_dir")
+
+        # Create output directories
+        self.create_activity_directories()
```
giorgiosironi: Most of these initializations of fields like `self.temp_dir`, `self.default_*` seem boilerplate, could it be done in the base `Activity` class?

activity/activity_CopyDigestToOutbox.py
```diff
@@ -0,0 +1,121 @@
...
+    def do_activity(self, data=None):
+        if self.logger:
+            self.logger.info('data: %s' % json.dumps(data, sort_keys=True, indent=4))
+        # parse the data with the digest_provider
+        real_filename, bucket_name, bucket_folder = parse_activity_data(data)
+        # Download from S3
+        self.input_file = digest_provider.download_digest_from_s3(
+            self.settings, real_filename, bucket_name, bucket_folder, self.input_dir)
+        # Parse input and build digest
+        self.build_status, self.digest = digest_provider.build_digest(
```
giorgiosironi: `self.build_status` doesn't seem to be used outside of this method so could simply be `build_status`

giorgiosironi: also `self.digest`, same reasoning applies unless there's something hidden that accesses these fields?

activity/activity_CopyDigestToOutbox.py
```diff
@@ -0,0 +1,121 @@
...
+        bucket_name = self.settings.bot_bucket
+
+        # clean out the outbox if not empty
+        self.clean_outbox(self.digest, bucket_name)
+
+        # copy the files to S3
+        # if it is zip file take the files from the temp_dir, otherwise from the input_dir
+        from_dir = self.input_dir
+        if self.input_file.endswith('.zip'):
+            from_dir = self.temp_dir
```
giorgiosironi: should the library always target `self.temp_dir` (whatever the input) so that this switch between `self.input_dir` and `self.temp_dir` can be simplified?

activity/activity_CopyDigestToOutbox.py
```diff
@@ -0,0 +1,121 @@
...
+        dest_resource = resource_path + new_file_name
+        return dest_resource
+
+    def clean_outbox(self, digest, bucket_name):
+        "remove files from the outbox folder"
+        resource_path = self.dest_resource_path(digest, bucket_name)
+        storage = storage_context(self.settings)
+        files_in_bucket = storage.list_resources(resource_path)
+        for resource in files_in_bucket:
+            self.logger.info("Deleting %s from the outbox", resource)
```
giorgiosironi: some logs are written before the action they take (`Deleting`) some after (`Copied`); usually the safest thing for integrations is to log before uniformely, unless there is a try/catch system in place; would change the other to `Copying ...`

activity/activity_CopyDigestToOutbox.py
gnott: In article ingestion as well as Medium article creation. ``for later`` I was just being short. Should I expand it do you think?

gnott: Oh, cancel that, we'll have a different system for Medium article creation. I'll add ``in article ingestion`` here :+1:

activity/activity_CopyDigestToOutbox.py
gnott: I think the activities that use the subdirectories the most are in the PoA packaging, where it needed lots of places to unzip files, rename them, repackage again, FTP them to an endpoint. That said, the digest related activities use these two ``input_dir`` and ``temp_dir`` because of how the ``digestparser`` library works.

There is certainly some opportunities to establish some commonly used and created directories in every activity and we create them every time, or when invoked. Is it something we could describe in a new issue for general activity class clean-up? I like the idea for sure!

activity/activity_CopyDigestToOutbox.py
gnott: Nothing checks for ``build_status`` on this activities tests, correct. It looks like I was going to test for the digest object DOI, but I forgot to, so instead I'll change it as you've suggested :+1:

activity/activity_CopyDigestToOutbox.py
gnott: I like the ``Copying`` edit, applying it now. When adding the ``Deleting`` logging line, I thought it would be better to log before potentially throwing an exception, even though the worker wraps this and we would get the exception caught, we might lose the logging message when it happens.

provider/digest_provider.py
gnott: The logic for that can go in this provider, and it is structured to allow it. I think a standard, web-friendly naming format is a good idea for images and I wouldn't expect much resistance to changing the name of it. We would probably want to rename it before copying to the CDN bucket and the IIIF image will get that name too, yes?

activity/activity_CopyDigestToOutbox.py
giorgiosironi: No clear specification from me on what to extract since I don't know the requirements of other activities, but yes it can be a follow-up

activity/activity_CopyDigestToOutbox.py
giorgiosironi: Yes, preemptive logs give context as all lines can be traced by referring to the `worker_1234` identity

PR #773, opened 2018-09-18
Ingest digest to endpoint

gnott:
Fixes https://github.com/elifesciences/issues/issues/4372

Very basic summary, add new activity `IngestDigestToEndpoint`, added to the end of the `ProcessArticleZip` workflow, to `PUT` digest JSON to an API endpoint.

The `digest_provider.py` provider was refactored (where `CopyDigestToOutbox` is changed to rely on it more), and expanded with JSON generation logic and API endpoint logic.

Introduces the ability to get article JSON from Lax, and also authenticate with Lax to get unpublished content, in order to set `relatedContent` in the digest JSON.

Many test scenarios for the activity and the digest provider to cover expected possible error situations when gathering data from the many external sources.

This is relying on an unmerged commit of the `digest-parser` library, used here to be able to verify whether it was working. I think the library has some feedback yet to finish about handling IIIF server errors and logging in https://github.com/elifesciences/digest-parser/pull/43.

Code for review, and I hope it may show how some end2end tests might be possible for this new activity.

I think holding off more refactoring to make the new activity `PublishDigest` to reuse some of of the digest logic is a possible course of action. We could get digests to ingest in the test environments to allow front-end development to continue. Publishing the ingests is a smaller later step. The coordination of how to deploy this activity is open for feedback.


requirements.txt
```diff
@@ -35,5 +35,5 @@ six==1.10.0
 pyFunctional==1.0.0
 func_timeout==4.3.0
 python-docx==0.8.6
-git+https://github.com/elifesciences/digest-parser.git@bd6a52744c213a4dfb7e4f6677643f2d0a8c20c0#egg=digestparser
+git+https://github.com/elifesciences/digest-parser.git@76255df400354e0ec9464bfdf64a2e1ed0e681bc#egg=digestparser
```
giorgiosironi: blocker for merging: this must be in the `master` of `digestparser` (if I understood correctly)

provider/digest_provider.py
```diff
@@ -78,3 +113,65 @@ def has_image(digest_content):
     if not digest_content.image.file:
         return False
     return True
+
+
+def digest_get_request(url, verify_ssl, digest_id):
+    "common get request logic to digests API"
+    response = requests.get(url, verify=verify_ssl)
```
giorgiosironi: is this `verify_ssl` still needed? (also in the `lax_provider`) All requests should go through an internal gateway that only exposes HTTP (`*--gateway.elife.internal`) and where we provide SSL in general the certificates are correctly signed and can be verified.

activity/activity_IngestDigestToEndpoint.py
```diff
@@ -0,0 +1,377 @@
+import os
+import json
+from digestparser import json_output
+from provider.storage_provider import storage_context
+from provider.execution_context import get_session
+import provider.digest_provider as digest_provider
+import provider.lax_provider as lax_provider
+from .activity import Activity
```
giorgiosironi: the set of dependencies (providers and activities) makes sense

activity/activity_IngestDigestToEndpoint.py
```diff
@@ -0,0 +1,377 @@
...
+                            " to be run when a research article is ingested")
+
+        # Local directory settings
+        self.temp_dir = os.path.join(self.get_tmp_dir(), "tmp_dir")
+        self.input_dir = os.path.join(self.get_tmp_dir(), "input_dir")
+
+        # Create output directories
+        self.create_activity_directories()
+
+        # Track the success of some steps
```
giorgiosironi: If these are meant to be steps, they may benefit from a common data structure like
```
self.statuses = [
    'approve': None,
    ...
]
```
which is easy to dump in case you want to log it.

activity/activity_IngestDigestToEndpoint.py
```diff
@@ -0,0 +1,377 @@
...
+        try:
+            run = data["run"]
+            session = get_session(self.settings, data, run)
+            version = session.get_value("version")
+            article_id = session.get_value("article_id")
+            success = True
+        except (TypeError, KeyError) as exception:
+            self.logger.exception("Exception when getting the session for Starting ingest digest " +
+                                  " to endpoint. Details: %s" % str(exception))
+            success = False
```
giorgiosironi: favor throwing an exception to a boolean return value as part of a tuple. That lets you deal with most failure with a single top-level `except` that declares the activity as a permanent failure (unless the exception is special and can be retried). Will simplify the tuple by making it shorter.

activity/activity_IngestDigestToEndpoint.py
```diff
@@ -0,0 +1,377 @@
...
+        "emit message to the queue"
+        success = None
+        try:
+            self.emit_monitor_event(self.settings, article_id, version, run,
+                                    self.pretty_name, status, message)
+            success = True
+        except Exception as exception:
+            self.logger.exception("Exception emitting %s message. Details: %s" %
+                                  (str(status), str(exception)))
+            success = False
```
giorgiosironi: also here, can throw rather than return `False`

activity/activity_IngestDigestToEndpoint.py
```diff
@@ -0,0 +1,377 @@
...
+            if status_code == 204:
+                put_status = True
+            else:
+                self.logger.error("PUT to digest endpoint status code  %s. Response: %s" %
+                                  str(status_code), str(response))
+        except Exception as exception:
+            self.logger.exception(
+                "Exception issuing PUT to the digest endpoint for digest_id %s. Details: %s" %
+                (str(digest_id), str(exception)))
+            put_status = False
```
giorgiosironi: same as before, once logging has been dealt with, can throw rather than returning a boolean

activity/activity_IngestDigestToEndpoint.py
```diff
@@ -0,0 +1,377 @@
...
+        for dir_name in [self.temp_dir, self.input_dir]:
+            try:
+                os.mkdir(dir_name)
+            except OSError:
+                pass
+
+
+def approve_by_status(logger, article_id, status):
+    "determine approval status by article status value"
+    approve_status = None
```
giorgiosironi: makes sense to use `None` as the neutral value, as `True` would mean "definitely ingest this one"

activity/activity_IngestDigestToEndpoint.py
```diff
@@ -0,0 +1,377 @@
...
+    related = None
+    try:
+        status_code, related_json = lax_provider.article_json(article_id, version, settings, auth)
+    except Exception as exception:
+        self.logger.exception(
+            "Exception in getting article_json from Lax for article_id %s, version %s. Details: %s" %
+            (str(article_id), str(version), str(exception)))
+    if related_json:
+        related = [related_json]
+    return status_code, related
```
giorgiosironi: `status_code` can be dropped here too

activity/activity_IngestDigestToEndpoint.py
```diff
@@ -0,0 +1,377 @@
...
+        storage_resource_origin = orig_resource + "/" + article_xml_filename
+        storage.get_resource_to_file(storage_resource_origin, open_file)
+        return filename_plus_path
+
+
+def related_from_lax(article_id, version, settings, auth=True):
+    "get article json from Lax and return as a list of related data"
+    related = None
+    try:
+        status_code, related_json = lax_provider.article_json(article_id, version, settings, auth)
```
giorgiosironi: inside `relatedContent` we don't need the full article, but only its snippet. If `lax_provider.article_json()` is returning all the fields, how can we strip it down to the snippet or request only the part that we need?

The hardcoded version that comes to mind to me is to only keep a fixed set of fields, perhaps the transformation to snippet is already dealt with in a provider or in the `digest-parser`.

activity/activity_IngestDigestToEndpoint.py
```diff
@@ -0,0 +1,377 @@
...
+            return self.ACTIVITY_SUCCESS
+
+        # get existing digest data
+        digest_id = self.digest_content.get("id")
+        digest_status_code, digest_json = digest_provider.get_digest(digest_id, self.settings)
+        if not digest_json:
+            self.logger.info(
+                "Did not get existing digest json from the endpoint for digest_id %s" %
+                str(digest_id))
+        self.digest_content = sync_json(self.digest_content, digest_json)
```
giorgiosironi: unclear in this call what is the existing digest and what is the new ingestion. Can rename `digest_json` to `existing_digest_json`?

giorgiosironi: (also in `sync_json`'s function definition)

activity/activity_IngestDigestToEndpoint.py
```diff
@@ -0,0 +1,377 @@
...
+
+def sync_json(json_content, digest_json):
+    "update values in json_content with some from digest_json if present"
+    if not digest_json:
+        return json_content
+    for attr in ["published", "stage"]:
+        if digest_json.get(attr):
+            json_content[attr] = digest_json.get(attr)
+    return json_content
+
```
giorgiosironi: Functions in this module, but outside of the activity class, are an intermediate level of isolation with respect to pushing them inside a `provider`. I find it useful at least to keep them independent of that class.

activity/activity_IngestDigestToEndpoint.py
```diff
@@ -0,0 +1,377 @@
...
+        # Digest JSON content
+        self.digest_content = None
+
+        # Load the config
+        self.digest_config = digest_provider.digest_config(
+            self.settings.digest_config_section,
+            self.settings.digest_config_file)
+
+    def do_activity(self, data=None):
+        self.logger.info("data: %s" % json.dumps(data, sort_keys=True, indent=4))
```
giorgiosironi: requirement for rollout: a big `try`/`except` that engulfs the whole activity and logs any exception as `ERROR`. We can then tackle errors without stopping the article ingestions.

activity/activity_IngestDigestToEndpoint.py
```diff
@@ -0,0 +1,377 @@
...
+        # check by status
+        return_status = approve_by_status(self.logger, article_id, status)
+        if return_status is False:
+            approve_status = return_status
+
+        # check silent corrections
+        return_status = approve_by_run_type(
+            self.settings, self.logger, article_id, run_type, version)
+        if return_status is False:
+            approve_status = return_status
```
giorgiosironi: I think there is another case to exclude:

- v1 poa has been published
- v2 vor has been published
- v3 vor is being ingested

In this last case, you don't want to ingest the digest again because (since digests do not have versions), it will be immediately visible as the only digest version. The safer option to keep it unpublished would be to only deal with that v3 during `PublishDigest`, but it takes away the possibility to preview. 

Opinion of the day is that we should simplify this and say a digest is _only_ ingested and published once, then it has to be silently corrected if there is a problem. In this case the digest will remain "tied" to v2 forever. But this will remove lots of headaches or the necessity to introduce versions in the `digests` service as well.

activity/activity_IngestDigestToEndpoint.py
lsh-0: > ... how can we strip it down to the snippet or request only the part that we need?
> The hardcoded version that comes to mind to me is to only keep a fixed set of fields, ...

yup: https://github.com/elifesciences/lax/blob/develop/src/publisher/fragment_logic.py#L109-L126

despite my plea in the comment to define the structure of a snippet in the schema, this section of the code has been very stable and only changed once iirc, to add a new field.

gnott: I was just looking for this in Lax right now. The snippets are only returned in the `versions` endpoint though as a list, are they? I would only want the most recent version (including unpublished snippets). I may be able to adapt it to read the versions data, but it seemed simpler to just copy the values from a single article version's JSON.

lsh-0: @thewilkybarkid might have an idea about a parameter that could be passed to the `article` endpoint that would return the snippet instead of the full article json. I have no problems implementing it, the alternatives are we slurp down all the snippets for the versions endpoint and pick the largest or copy the logic in lax to create a snippet from the full article json.

The path of least resistance would be hitting the `versions` endpoint I think.

thewilkybarkid: That's what Journal CMS does IIRC (though we have that logic separately implemented in PHP).

If we did allow access to snippets, it'd be through content negotiation. But lets keep it as simple as we can.

giorgiosironi: :+1: for filtering a whitelist of snippet fields from the full article
:+1: also for reading the most recent item in `/articles/:id/versions` if that's easier

giorgiosironi:
Pylint is sniffing some problems:
```
[18.212.227.99] out: ************* Module activity.activity_IngestDigestToEndpoint
[18.212.227.99] out: E:254,21: Undefined variable 'docx_file' (undefined-variable)
[18.212.227.99] out: E:356, 8: Undefined variable 'self' (undefined-variable)
```

activity/activity_IngestDigestToEndpoint.py
giorgiosironi: TL;DR: to start with it, I suggest to only ingest and publish a digest on the first version that has one (meaning the first VOR), don't do anything on further version.

requirements.txt
gnott: The sha for `digestparser` will be updated for the change to the `subject_slug()` logic, but I think we are normally ok with deploying commits on `develop` branch for most libraries. It can be merged to `master` if you think so.

provider/digest_provider.py
gnott: Technically we do not need it now since we're not connecting to `https://`. It depends on choosing between clean code versus flexible code. To me, it looks like leaving the option open to connect to a non-VPC endpoint could be useful to retain.

activity/activity_IngestDigestToEndpoint.py
gnott: I like it, and done it. I was looking for ways to reduce the class properties, according to the linter too many are bad.

activity/activity_IngestDigestToEndpoint.py
gnott: Wrapped them just below this first line that logs the input data, which I hope will be safe.

activity/activity_IngestDigestToEndpoint.py
gnott: If we want to only ingest a digest once, we could possibly check whether a published digest already exists, and if so, we do not ingest one again, if it is not in the case of a silent correction. If there's a digest with the stage of `'preview'` then we can ingest it again, which would presumably be an additional run for the v2 vor in the above steps. But, ... looking at it again we can probably just work from the Lax versions and only approve it for the first vor version.

I've run through some other options of how updating a digest to have the paragraphs from the most recent JATS file might be good, but it doesn't outweigh the simplicity of only publishing a digest once. A silent correction would then be required to forcibly replace a published digest.

gnott: I think I got the only first vor logic, but when silent correction only allow the most recent version. In the end maybe it can be simpler, and my attempt may have left some extra code it didn't actually need. But, I think it works, there a bunch of test cases for `approve()` specifically now, and the test cases can be expanded, or it can be refactored to satisfy the existing test cases, as desired.

gnott:
Hope these revisions get us closer, if there's a comment I missed, please be welcome to point it out and I will take a look. Looking to reach a approval to see how this works (or doesn't) on continuumtest.

provider/lax_provider.py
```diff
@@ -15,25 +15,83 @@ class ErrorCallingLaxException(Exception):
...
+def lax_request(url, article_id, verify_ssl, request_type='version', auth_key=None):
+    "common request logic to Lax"
+    response = requests.get(url, verify=verify_ssl, headers=lax_auth_header(auth_key))
     logger.info("Request to lax: GET %s", url)
     logger.info("Response from lax: %s\n%s", response.status_code, response.content)
     status_code = response.status_code
     if status_code not in [200, 404]:
-        raise ErrorCallingLaxException("Error looking up article " + article_id + " version in Lax: %s\n%s" % (status_code, response.content))
+        raise ErrorCallingLaxException(
+            "Error looking up article " + article_id + " %s in Lax: %s\n%s" %
```
lsh-0: is `article_id` a string or an integer here?
```python
>>> article_id = 123
>>> "" + article_id
  File "<stdin>", line 1, in <module>
TypeError: can only concatenate str (not "int") to str
>>> "%s" % article_id
'123'
```

gnott: Generally in the `elife-bot` project, and specifically in the workflow session, the `article_id` is stored as a string. It is derived from S3 object names, and is not converted to an `int`. There were some issues with a mismatch of the values, and I think when the value comes back from Lax upon ingest, it is explicitly cast back to a string. Subsequently it is used to create folder names, so almost always it is cast as a string.

giorgiosironi: In general (also in Libero) we are using id as strings to be more general e.g. article `d41586-018-05525-y` (a real one from Nature)

lsh-0: well, if you have an ID like `d41586-018-05525-y` then you *obviously* can't use an int... elife uses an integer as an ID, which is brilliant, but for no good reason everybody insists on padding it, propagating a whimsical decision made by Andy of editorial down through the years and piles of code.

PR #775, opened 2018-09-20
Emit end activity event for digest ingestion

giorgiosironi:
Found in https://github.com/elifesciences/issues/issues/4326

Avoids this situation on activity failure (which we catch):

Moreover, the article should be in a `ready to publish` state only if its digest has been processed (whether it has been skipped because it's a PoA, successfully ingested, failed but caught and skipped, etc.). Practically this has little effect for now because we never mark the activity as failed, but long-term it will mean maintaining a clear boundary for when the `Publish` button appears.


gnott:
I will merge so I can do some more edits and not worry about a merge conflict.

giorgiosironi:
Partly worked:

giorgiosironi:
Meaning the order worked, not the icon, but perhaps it must be explicitly positive to become green. This seems good enough for now.

PR #780, opened 2018-09-21
PUT /digests/:id Content-Type request header

giorgiosironi:
Header seems mandatory for the time being. We test both `Authorization` and `Content-Type` are passed in.


giorgiosironi:
Let's try to integrate this.

PR #784, opened 2018-09-26
Feature publish digest

gnott:
Implementing https://github.com/elifesciences/issues/issues/4303.

Started off with the basics of `GET` a digest, if it exists, change to `published` (if applicable) and `PUT` it back.

I refactored by moving some of the functions from the `IngestDigestToEndpoint` activity over to the `digest_provider`.

`PublishDigest` only checks for whether it is a `poa` and does not approve continuing for those. The rest of the logic is based on the `stage` status of the digest JSON. There are a few test scenarios for different type of input.

Exceptions will be caught in `PublishDigest` if anything goes awry.

I added it to the `PostPerfectPublication` workflow before it tries the `GeneratePDFCovers` activity, because `GeneratePDFCovers` still tends to terminate some workflows on a weekly basis, and I don't want it to interfere with digest publishing.

### DIFF
--- a/activity/activity_PublishDigest.py
+++ b/activity/activity_PublishDigest.py
@@ -135,7 +135,7 @@ class activity_PublishDigest(Activity):
         approve_status = True
 
         # check by status
-        return_status = approve_by_status(self.logger, article_id, status)
+        return_status = digest_provider.approve_by_status(self.logger, article_id, status)
         if return_status is False:
             approve_status = return_status
 
@@ -152,19 +152,6 @@ class activity_PublishDigest(Activity):
                 "Exception issuing PUT to the digest endpoint for digest_id %s. Details: %s" %
                 (str(digest_id), str(exception)))
         return put_status
-
-
-def approve_by_status(logger, article_id, status):
-    "determine approval status by article status value"
-    approve_status = None
-    # PoA do not ingest digests
-    if status == "poa":
-        approve_status = False
-        message = ("\nNot ingesting digest for PoA article {article_id}".format(
-            article_id=article_id
-        ))
-        logger.info(message)
-    return approve_status
 
 
 def set_stage(json_content, stage="preview"):
--- a/activity/activity_PublishDigest.py
+++ b/activity/activity_PublishDigest.py
@@ -21,7 +21,6 @@ class activity_PublishDigest(Activity):
         # Track the success of some steps
         self.statuses = {
             "approve": None,
-            "stage": None,
             "put": None
         }
 
@@ -71,11 +70,10 @@ class activity_PublishDigest(Activity):
             if existing_digest_json.get("stage") != "published":
                 self.digest_content = digest_provider.set_stage(existing_digest_json, 'published')
                 self.logger.info("Set Digest stage value of %s to published" % article_id)
-                self.statuses["stage"] = True
-            if self.statuses.get("stage"):
-                self.statuses["put"] = digest_provider.put_digest_to_endpoint(
+                put_response = digest_provider.put_digest_to_endpoint(
                     self.logger, digest_id, self.digest_content, self.settings)
-                if self.statuses.get("put"):
+                if put_response:
+                    self.statuses["put"] = True
                     self.logger.info("Put Digest for %s to the endpoint" % article_id)
 
         except Exception as exception:
--- a/activity/activity_PublishDigest.py
+++ b/activity/activity_PublishDigest.py
@@ -69,7 +69,7 @@ class activity_PublishDigest(Activity):
 
             # set the stage attribute if is not published
             if existing_digest_json.get("stage") != "published":
-                self.digest_content = set_stage(existing_digest_json, 'published')
+                self.digest_content = digest_provider.set_stage(existing_digest_json, 'published')
                 self.logger.info("Set Digest stage value of %s to published" % article_id)
                 self.statuses["stage"] = True
             if self.statuses.get("stage"):
@@ -154,7 +154,3 @@ class activity_PublishDigest(Activity):
         return put_status
 
 
-def set_stage(json_content, stage="preview"):
-    "set the stage attribute if missing"
-    json_content["stage"] = stage
-    return json_content
--- a/activity/activity_PublishDigest.py
+++ b/activity/activity_PublishDigest.py
@@ -62,7 +62,7 @@ class activity_PublishDigest(Activity):
             existing_digest_json = digest_provider.get_digest(digest_id, self.settings)
             if not existing_digest_json:
                 self.logger.info(
-                    "Did not get existing digest json from the endpoint for digest_id %s" %
+                    "There is no existing digest for digest_id %s" %
                     str(digest_id))
                 self.emit_end_message(article_id, version, run)
                 return self.ACTIVITY_SUCCESS
--- a/activity/activity_PublishDigest.py
+++ b/activity/activity_PublishDigest.py
@@ -1,0 +1,173 @@
+import json
+import provider.digest_provider as digest_provider
+from .activity import Activity
+
+
+class activity_PublishDigest(Activity):
+    def __init__(self, settings, logger, conn=None, token=None, activity_task=None):
+        super(activity_PublishDigest, self).__init__(
+            settings, logger, conn, token, activity_task)
+
+        self.name = "PublishDigest"
+        self.pretty_name = "Publish the Digest via the API endpoint"
+        self.version = "1"
+        self.default_task_heartbeat_timeout = 30
+        self.default_task_schedule_to_close_timeout = 60 * 5
+        self.default_task_schedule_to_start_timeout = 30
+        self.default_task_start_to_close_timeout = 60 * 5
+        self.description = ("Set digest as published, related to the research article," +
+                            " if the digest not already published.")
+
+        # Track the success of some steps
+        self.statuses = {
+            "approve": None,
+            "stage": None,
+            "put": None
+        }
+
+        # Digest JSON content
+        self.digest_content = None
+
+        # Load the config
+        self.digest_config = digest_provider.digest_config(
+            self.settings.digest_config_section,
+            self.settings.digest_config_file)
+
+    def do_activity(self, data=None):
+        self.logger.info("data: %s" % json.dumps(data, sort_keys=True, indent=4))
+
+        # get data
+        (success, run, article_id, version, status) = self.parse_data(data)
+        if success is not True:
+            self.logger.error("Failed to parse data in %s" % self.pretty_name)
+            return self.ACTIVITY_PERMANENT_FAILURE
+        # emit start message
+        success = self.emit_start_message(article_id, version, run)
+        if success is not True:
+            self.logger.error("Failed to emit a start message in %s" % self.pretty_name)
+            return self.ACTIVITY_PERMANENT_FAILURE
+
+        # Wrap in an exception during testing phase
+        try:
+            # Approve for ingestion
+            self.statuses["approve"] = self.approve(article_id, status)
+            if self.statuses.get("approve") is not True:
+                self.logger.info(
+                    "Digest for article %s was not approved for publish" % article_id)
+                self.emit_end_message(article_id, version, run)
+                return self.ACTIVITY_SUCCESS
+
+            # get existing digest data
+            digest_id = article_id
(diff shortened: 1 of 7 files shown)


giorgiosironi:
> GeneratePDFCovers still tends to terminate some workflows on a weekly basis

:grrr:

provider/digest_provider.py
```diff
@@ -182,3 +183,73 @@ def put_digest(digest_id, data, settings, auth=True):
     url = settings.digest_endpoint.replace('{digest_id}', str(digest_id))
     return digest_put_request(url, settings.verify_ssl, digest_id, data,
                               digest_auth_key(settings, auth))
+
+
```
giorgiosironi: these are all functions moved from the first activity, good for me to reuse them

tests/activity/test_activity_publish_digest.py
```diff
@@ -0,0 +1,143 @@
...
+                         test_data.get("expected_approve_status"),
+                         'failed in {comment}'.format(comment=test_data.get("comment")))
+        self.assertEqual(self.activity.statuses.get("stage"),
+                         test_data.get("expected_stage_status"),
+                         'failed in {comment}'.format(comment=test_data.get("comment")))
+        self.assertEqual(self.activity.statuses.get("put"),
+                         test_data.get("expected_put_status"),
+                         'failed in {comment}'.format(comment=test_data.get("comment")))
+        # check stage value in json_content
+        if self.activity.digest_content:
```
giorgiosironi: can this `if` read something from `test_data` rather than the `self.activity` fields? As it stands runs the risk of skipping the assertion because of a change in the activity code, while it should always execute it for the cases that need it

activity/activity_PublishDigest.py
```diff
@@ -0,0 +1,140 @@
...
+        # Load the config
+        self.digest_config = digest_provider.digest_config(
+            self.settings.digest_config_section,
+            self.settings.digest_config_file)
+
+    def do_activity(self, data=None):
+        self.logger.info("data: %s" % json.dumps(data, sort_keys=True, indent=4))
+
+        # get data
+        (success, run, article_id, version, status) = self.parse_data(data)
```
giorgiosironi: a matter for later, but do we need this on every activity? It seems like a fallback to return `self.ACTIVITY_PERMANENT_FAILURE` that should be enacted by the base `activity` class in case of unknown errors (or of this error signaled by a `ParseSessionDataError` exception)

activity/activity_PublishDigest.py
```diff
@@ -0,0 +1,140 @@
...
+        # emit start message
+        success = self.emit_start_message(article_id, version, run)
+        if success is not True:
+            self.logger.error("Failed to emit a start message in %s" % self.pretty_name)
+            return self.ACTIVITY_PERMANENT_FAILURE
+
+        # Wrap in an exception during testing phase
+        try:
+            # Approve for ingestion
+            self.statuses["approve"] = self.approve(article_id, status)
```
giorgiosironi: I don't understand the usage of `approve` here; there is a human approving something from a dashboard or a preview? What I think was happening is that PoA and some later-version-VoRs do not have a digest to publish at all; so we should call it `digest_enabled`? `digest_attached`?

activity/activity_PublishDigest.py
```diff
@@ -0,0 +1,140 @@
...
+                    "Digest for article %s was not approved for publish" % article_id)
+                self.emit_end_message(article_id, version, run)
+                return self.ACTIVITY_SUCCESS
+
+            # get existing digest data
+            digest_id = article_id
+            existing_digest_json = digest_provider.get_digest(digest_id, self.settings)
+            if not existing_digest_json:
+                self.logger.info(
+                    "Did not get existing digest json from the endpoint for digest_id %s" %
```
giorgiosironi: it did try, but there was nothing there; `No existing digest json for digest_id %s`?

activity/activity_PublishDigest.py
```diff
@@ -0,0 +1,140 @@
...
+                    str(digest_id))
+                self.emit_end_message(article_id, version, run)
+                return self.ACTIVITY_SUCCESS
+
+            # set the stage attribute if is not published
+            if existing_digest_json.get("stage") != "published":
+                self.digest_content = digest_provider.set_stage(existing_digest_json, 'published')
+                self.logger.info("Set Digest stage value of %s to published" % article_id)
+                self.statuses["stage"] = True
+            if self.statuses.get("stage"):
```
giorgiosironi: can remove the condition and join with the previous `if`? All part of the happy path

activity/activity_PublishDigest.py
```diff
@@ -0,0 +1,140 @@
...
+                self.statuses["put"] = digest_provider.put_digest_to_endpoint(
+                    self.logger, digest_id, self.digest_content, self.settings)
+                if self.statuses.get("put"):
```
giorgiosironi: given `self.statuses['put']` has just been assigned in the previous line, this condition should always be `True`, hence not needed

giorgiosironi: Or is it because it can be `False`? I'd favor raising the exception from `digest_provider` and remove this log (`digest_provider` already logs the response, whether it is successful or not)

provider/digest_provider.py
```diff
@@ -182,3 +183,73 @@ def put_digest(digest_id, data, settings, auth=True):
...
+    "handle issuing the PUT to the digest endpoint"
+    put_status = None
+    try:
+        put_digest(digest_id, digest_content, settings)
+        put_status = True
+    except Exception as exception:
+        logger.exception(
+            "Exception issuing PUT to the digest endpoint for digest_id %s. Details: %s" %
+            (str(digest_id), str(exception)))
+    return put_status
```
giorgiosironi: We are covered in `except` blocks in the activities, so I'd say we can re-`raise` the exception rather than returning a boolean that could be go unchecked:

- happy path `None` is returned
- any error, an `Exception` is raised

activity/activity_PublishDigest.py
gnott: The need to parse the input data is very common, but not every activity has the same input data. 

I started `parse_data()` on a new activity in a different branch (https://github.com/elifesciences/elife-bot/blob/feature-email-video-article-published/activity/activity_EmailVideoArticlePublished.py). In that can we also want the `expanded_folder`, and here in `PublishDigest` we do not need that. The linter will say that that variable is not used. 

In the future, yes, maybe these could be refactored to specify which values to take from the input data on a per-activity basis. We could create a new issue to refactor all the activities as a separate PR maybe?

activity/activity_PublishDigest.py
gnott: `approve` is just a name for the logic to decide whether to actually perform the activity. In other places, it may be a list of approved articles, for sending to Crossref or sending an email. The rules for approval are all in code, it does not require human input. I guess more accurately here is means "approve for PublishDigest", and if an article is status "poa" it is not approved. Or the name could be "article_is_approved_for_publishing_a_digest". If you think there is a better name I'm open to suggestions.

Comparing it to the `IngestDigestToEndpoint` activity, there is more than one rule that could set `approve` to `False`. The function is a way to determine a `True` or `False` value for whether it should just return quietly before continuing. It saves having lots of nested conditional statements. Keeping track of the value in the `self.statuses{}` let's the tests for `do_activity()` check the value set at this point.

activity/activity_PublishDigest.py
gnott: The logger messages are probably not great, I didn't read them too carefully. At this point, it means there is no digest at all (`preview` or `published`), and I can improve the message here.

activity/activity_PublishDigest.py
gnott: I will do away with the `"stage"` status, we can just check for whether the `"put"` happened.

provider/digest_provider.py
gnott: A good observation here, thanks. 

I want to return something. 

Most of our requests are a `get`. There are a couple `post` in our existing providers, one being in the Fastly provider, where the return value if an exception is not raised is to return the `response`. 

I am also going to return the `response` if a digest `put` is successful, and this does away with the `put_status`. These affect both the `PublishDigest` and `IngestDigestToEndpoint` activities and their tests.

activity/activity_PublishDigest.py
giorgiosironi: Yes this is not specific to this feature so postpone

provider/digest_provider.py
giorgiosironi: I agree if the response is interesting, it should be returned in the positive case.

activity/activity_PublishDigest.py
giorgiosironi: On the naming, I would ask @elifesciences/production if they have a standard name for "features of an article that may be enabled only for a subset of articles depending on its poa/vor status and other categories e.g. whether a digest will be published together with it"

activity/activity_PublishDigest.py
```diff
@@ -0,0 +1,138 @@
...
+                    "There is no existing digest for digest_id %s" %
+                    str(digest_id))
+                self.emit_end_message(article_id, version, run)
+                return self.ACTIVITY_SUCCESS
+
+            # set the stage attribute if is not published
+            if existing_digest_json.get("stage") != "published":
+                self.digest_content = digest_provider.set_stage(existing_digest_json, 'published')
+                self.logger.info("Set Digest stage value of %s to published" % article_id)
+                put_response = digest_provider.put_digest_to_endpoint(
```
giorgiosironi: it seems that in the current implementation `digest_provider.put_digest_to_endpoint` will never fail. It is what we need for rollout, although we have to be careful to reintroduce hard error checks when the implementation is stable.

activity/activity_PublishDigest.py
```diff
@@ -0,0 +1,138 @@
...
+        "emit the start message to the queue"
+        return self.emit_message(
+            article_id, version, run, "start",
+            "Starting ingest digest to endpoint for " + str(article_id))
+
+    def emit_end_message(self, article_id, version, run):
+        "emit the end message to the queue"
+        return self.emit_message(
+            article_id, version, run, "end",
+            "Finished ingest digest to endpoint for " + str(article_id))
```
giorgiosironi: I think we could leverage this end message to expose `self.statuses`? It would appear in the dashboard giving immediate feedback without looking into the logs?

activity/activity_PublishDigest.py
gnott: I think the default retry option right now is if the activity returns `False`, then the `decider.py` will try the activity again after a standard 10 second wait after a failing activity. But, after the activity timeout, in this case 5 minutes, if the activity has not succeeded then the workflow will timeout, and the next activity in the series will not be attempted.

The only likely situation I would think here is just a short network or application issue where a second attempt would succeed, but if the endpoint is inaccessible for 5 minutes then we eventually get the workflow timeout.

I was thinking we could add a "number of retries" to activities as an improvement. To do that, we'd expand the decider's logic to count the failed attempts from the workflow execution history and if the retry count is reached, the activity could return `True` (after logging "max retries reached" or something). This workflow level context we do not look at right now.

activity/activity_PublishDigest.py
gnott: Alternate names from other teams are welcome. I don't think the `approve` word will show to anyone but those who are reading the code or log files at this point. It will quietly continue if a research article publication will not result in a digest publication.

activity/activity_PublishDigest.py
gnott: If you follow the execution through the functions in the `digest_provider` provider, if a non-success response comes from the endpoint, e.g. `404` or `500` (https://github.com/elifesciences/elife-bot/blob/31fe8bc8a4e607f6235a21e5723282c0a3441885/provider/digest_provider.py#L173) , then it will return None, and the `put` will be considered a failure. Also if it encounters an exception due to any other problem making the request it would also be considered a failure (https://github.com/elifesciences/elife-bot/blob/31fe8bc8a4e607f6235a21e5723282c0a3441885/provider/digest_provider.py#L192), from the context of this `PublishDigest` activity.

activity/activity_PublishDigest.py
giorgiosironi: :+1:

Issue #561, opened 2017-12-14
Activity to send email notification when an article with a video is published


gnott:
Basic plan is a new activity will be created, added to the ``PostPerfectPublication`` workflow. The activity will look in the article XML to see if there are any videos, and if so, it will send a basic email to eLife recipients notifying them of an article with videos was published.

- [x] Create new bot activity
- [x] Add email templates to the repo
- [x] Add email templates to the bucket
- [x] Switch to sending email directly by SMTP in the activity instead of using the email queue
- [x] Simplify the "send only once" logic to only send for only when the first VoR version is published
- [x] Add recipient emails to the bot settings
- [x] Add the bot activity to the `PostPerfectPublication` workflow

gnott:
Code is developed under this branch https://github.com/elifesciences/elife-bot/tree/feature-email-video-article-published

gnott:
Settings PR https://github.com/elifesciences/builder-private/pull/246
Bot PR https://github.com/elifesciences/elife-bot/pull/785  contains the activity and added to the workflow

gnott:
Defined features are all done, bot code is merged in, and I'm deploying `prod-elife-bot` now. I will call this done, and we can open new issues if there are bugs or change requests.

PR #785, opened 2018-09-27
Feature email video article published

gnott:
In reference to issue https://github.com/elifesciences/elife-bot/issues/561

Adds a new activity named `EmailVideoArticlePublished` that will email recipients once only (no duplicates allowed) if an article that contains a video is published. The recipients are specified in the bot settings, and they are internal recipients, not the authors of the article.

A little refactoring in the downloading of templates in the templates provider.

I also added some methods to the activity object in `activity.py` for emitting some messages which many of the activities do. These could be used later to refactor the message sending in other activities. I think they are general purpose enough to use the common methods.

There are additional bot settings to be added for this before it is deployed. I also need to add the new email templates before deployment.


activity/activity.py
```diff
@@ -198,6 +198,28 @@ def clean_tmp_dir(self):
...
+            return True
+        except Exception as exception:
+            self.logger.exception("Exception emitting %s message. Details: %s" %
+                                  (str(status), str(exception)))
+
+    def emit_activity_start_message(self, article_id, version, run):
+        "emit the start message to the queue"
+        return self.emit_activity_message(
+            article_id, version, run, "start",
+            "Starting ingest digest to endpoint for " + str(article_id))
```
giorgiosironi: these messages look too specific (digests-related)

activity/activity_EmailVideoArticlePublished.py
```diff
@@ -0,0 +1,315 @@
...
+        self.pretty_name = "Email Video Article Published"
+        self.version = "1"
+        self.default_task_heartbeat_timeout = 30
+        self.default_task_schedule_to_close_timeout = 60 * 5
+        self.default_task_schedule_to_start_timeout = 30
+        self.default_task_start_to_close_timeout = 60 * 5
+        self.description = "Send email when an article containing a video is published."
+
+        # Data provider
+        self.database = dblib.SimpleDB(settings)
```
giorgiosironi: adding dependencies on SimpleDB at this stage is a strategic problem, any other source of data and/or storage of state that can be used?

activity/activity_EmailVideoArticlePublished.py
```diff
@@ -0,0 +1,315 @@
...
+            version = data.get("version")
+            status = data.get("status")
+            expanded_folder = data.get("expanded_folder")
+            success = True
+        except (TypeError, KeyError) as exception:
+            self.logger.exception("Exception when getting the session for Starting ingest " +
+                                  " digest to endpoint. Details: %s" % str(exception))
+        return success, run, article_id, version, status, expanded_folder
+
+    def download_xml(self, expanded_bucket, expanded_folder, version, to_dir):
```
giorgiosironi: this function seems to achieve a very common operation; should be extracted into a provider (`xml_provider` using `storage_provider` underneath?)

activity/activity_EmailVideoArticlePublished.py
```diff
@@ -0,0 +1,315 @@
...
+            email_type=headers["email_type"],
+            format=headers["format"],
+            subject=headers["subject"],
+            body=body,
+            doi_id=doi_id,
+            date_scheduled_timestamp=date_scheduled_timestamp)
+
+    def is_duplicate_email(self, doi_id, email_type, recipient_email):
+        """
+        Use the SimpleDB provider to count the number of emails
```
giorgiosironi: where do these duplicates come from? Is the activity executed multiple times for a single publishing workflow?

activity/activity_EmailVideoArticlePublished.py
```diff
@@ -0,0 +1,315 @@
...
+        """
+        body = self.templates.get_email_body(
+            email_type=email_type,
+            author=recipient,
+            article=article,
+            authors=authors,
+            format=email_format)
+
+        # Add the email to the email queue
+        self.database.elife_add_email_to_email_queue(
```
giorgiosironi: no idea what happens once the emails get into a queue in the database. Do they get sent immediately by some other process? By a cron? Are they scheduled for later?

activity/activity_EmailVideoArticlePublished.py
giorgiosironi: This is even more interesting for the email-related code, what can be pushed into a `email_provider` rather than being rewritten in each activity? There will be bits dependent on settings or on particular session data, but the general machinery of rendering and sending an email should be reusable

activity/activity.py
gnott: Oops, I will fix the basics to try and make it suitable for general use.

activity/activity_EmailVideoArticlePublished.py
gnott: There's an opportunity for refactoring here. This particular activity is taking the input from the expanded folder of the article ingest workflow.

 The publication emails use an S3 outbox system for input, and is separate from the article ingest workflow, so it wouldn't apply there. 

The best place for a shared function might be in the `execution_context.py` or maybe `article_structure.py`, or a new provider.

There are a few duplicates of this action, some older and some slightly different. The activities `ApplyVersionNumber` and `ModifyArticleSubjects` download the XML from the expanded folder. There are also similar functions for finding the article XML in the expanded folder in `article_structure/get_article_xml_key()` (https://github.com/elifesciences/elife-bot/blob/develop/provider/article_structure.py#L189) and `article/get_xml_file_name()` (https://github.com/elifesciences/elife-bot/blob/develop/provider/article.py#L734), also used by `lax_provider.get_xml_file_name()` (https://github.com/elifesciences/elife-bot/blob/develop/provider/lax_provider.py#L217).

No one has attempted to find the commonality between all of these and replace them in all the activities yet. It has evolved over time with different developers.

Trying to refactor this might be better attempted on a separate branch and a different PR for the functions listed above. This new `EmailVideoArticlePublished` activity could be blocked by that refactoring, and then once merged it can be used here.

What do you think about a separate refactor for finding and downloading article XML from the expanded bucket?

activity/activity_EmailVideoArticlePublished.py
gnott: The SimpleDB records form the history of emails, and duplicates are found in that history. It checks duplicates by a combination of `email_template_name` + `recipient_email`.

activity/activity_EmailVideoArticlePublished.py
gnott: All queued emails (the queue is stored in SimpleDB) are sent in the `SendQueuedEmail` workflow, the cron starts every 5 minutes. They are immediate in this case (but the queue does support scheduling emails to be sent at a future date).

This new `EmailVideoArticlePublished` activity doesn't need to rely on the email queue, if it looks like a non-SimpleDB method of history and persistence is a better option here. We could track history of sending on a per-article basis by keeping track in S3 folders, and we could also send emails immediately using the SMTP method we know have when it was created for sending digest content.

activity/activity_EmailVideoArticlePublished.py
giorgiosironi: > It is not a complicated situation, but it requires effort and care.

Definitely, let's not add even more variations, as you say.

activity/activity_EmailVideoArticlePublished.py
giorgiosironi: > about a separate refactor for finding and downloading article XML from the expanded bucket

Definitely good to have an issue for that, it affects lots of key activities; whereas emails do not have the same impact on publication (I think).

activity/activity_EmailVideoArticlePublished.py
giorgiosironi: There are already a couple of queues (Postfix/SES) in place if you use SMTP; so if the queuing can be dropped without having to implement persistence with S3 and sending emails directly, I'd go for that. Assuming that removes code or executes less existing code, which is why it would be helpful as fewer things can break/need to be maintained.

activity/activity_EmailVideoArticlePublished.py
gnott: I'm having a rethink on the duplicates. We might be able to base it on sending emails

- only on first VoR published
- never after a silent correction

Then the duplicate check doesn't need to check the sending history.

activity/activity_EmailVideoArticlePublished.py
gnott: If the logic to send it based on VoR version and `run_type`, then there's no need to add it to the queue to check duplicates. We can just send directly by the SMTP method.

gnott:
It's been a while for this PR. I addressed earlier review comments.

Recent changes were to switch over to sending email direct by SMTP (instead of the older email queue), and to change how it decides when to send an email (no longer relying on checking for duplicate emails).

Also refactored how the XML is downloaded from the expanded folder, using `article_processing` provider now.

I expanded the `email_provider.py` to load email headers and body from templates with exception handling. It makes this activity code simpler, and is useful in further email refactoring I'm planning soon.

Including @GiancarloFusiello for review too if you are curious.

All the boxes I self-defined on issue https://github.com/elifesciences/elife-bot/issues/561 are completed here.

Thanks!

gnott:
Thanks for taking a look again, I guess if tests are passing it is time to merge and watch it go!

PR #786, opened 2018-09-27
Introduce get_digest_preview for PublishDigest

giorgiosironi:
The last `PublishDigest` error was

```
2018-09-27T08:37:56Z INFO worker_5920 Request to digest API: GET http://end2end--gateway.elife.internal/digests/6955600639083406847
2018-09-27T08:37:56Z INFO worker_5920 Response from digest API: 404
```

It likely has to check for preview digests too.

Note on multiple versions: I think `PublishDigest` will always re-publish the digest, even in a v3 VoR after a v2 VoR that already published it. Seems harmless but if we have a way to avoid it, one fewer thing that can go wrong.

Note on `IngestDigestToEndpoint`: I cannot guess the case for which we read back the digest. It currently can only read a published digest, so it should be for silent corrections. However the activity is not in silent corrections workflow, could we start out by not doing this at all? My comments are in the activity.

### DIFF
--- a/activity/activity_IngestDigestToEndpoint.py
+++ b/activity/activity_IngestDigestToEndpoint.py
@@ -115,6 +115,11 @@ class activity_IngestDigestToEndpoint(Activity):
 
             # get existing digest data
             digest_id = self.digest_content.get("id")
+            # TODO: what are we doing this for?
+            # assumption: we are ingesting the digest multiple times on the first VoR version
+            # assumption: we are not ingesting the digest on any other version
+            # only possible case is silent correction?
+            # but then, only do it on silent correction?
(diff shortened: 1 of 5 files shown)


giorgiosironi:
I'm going to merge this to fix the `PublishDigest` activity - by all means @gnott go through it to check if we can simplify anything.

gnott:
You may be correct we would only want to sync digest json content if it is due to a silent correction. If we can determine very clearly any situations we would want to take data from an existing digest and use it during ingestion, that would be good to be sure. It could also be we only sync the data if the existing digest has `stage` of `published`.

giorgiosironi:
> It could also be we only sync the data if the existing digest has stage of published.

Seems a good option to only select silent corrections; I'll take a look at the existing PR first.

PR #787, opened 2018-09-27
Isolate providers logs

giorgiosironi:
Fixes an issue where all providers end up using the same logger as the main process, meaning they duplicate into their own file the whole output of `worker.log` or whatever is using them:

```
elife@end2end--bot:/opt/elife-bot$ grep activityType: digest_provider.log | head
2018-09-20T13:15:45Z INFO process_4086 activityType: VerifyGlencoe
2018-09-20T13:15:48Z INFO process_4086 activityType: PingWorker
2018-09-20T13:15:48Z INFO process_4092 activityType: VersionLookup
2018-09-20T13:15:49Z INFO process_4086 activityType: PingWorker
2018-09-20T13:15:49Z INFO process_4092 activityType: VerifyPublishResponse
2018-09-20T13:15:49Z INFO process_4086 activityType: VersionReasonDecider
2018-09-20T13:15:50Z INFO process_4092 activityType: CopyGlencoeStillImages
2018-09-20T13:15:51Z INFO process_4086 activityType: ApplyVersionNumber
2018-09-20T13:15:52Z INFO process_4092 activityType: VersionReasonDecider
2018-09-20T13:15:54Z INFO process_4092 activityType: PingWorker
```

After this, only the calls inside the specific `provider.something._provider` should go to its own file, with no mixing.

### DIFF
--- a/activity/activity_ArchiveArticle.py
+++ b/activity/activity_ArchiveArticle.py
@@ -99,30 +99,3 @@ class activity_ArchiveArticle(activity.activity):
                                 "end", "Finished archiving article " + data['article_id'] +
                                 " for version " + version + " run " + data["run"])
         return activity.activity.ACTIVITY_SUCCESS
-
-def main(args):
-
-    """
-    This sets up dummy SWF activity data, creates an instance of this activity and runs it only for
-    testing and debugging. This activity would usually be executed by worker.py
-    """
-
-    data = {
-        'id': '00353',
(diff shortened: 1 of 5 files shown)


giorgiosironi:
> the logger can be passed in by the worker

It's idiomatic for Python to tag log lines with their module to quickly locate where they come from. Here we are going further and have separate files per module; having a single file would be easier for operations and also for correlating lines (A comes before B which comes before C), but the sheer amount of logs volume would make it difficult to navigate.

PR #788, opened 2018-09-27
PUT digest troubleshooting

giorgiosironi:
It's already working now, but in failure cases it will be helpful to have good logs and visibility on the dashboard.

### DIFF
--- a/activity/activity_PublishDigest.py
+++ b/activity/activity_PublishDigest.py
@@ -114,13 +114,15 @@ class activity_PublishDigest(Activity):
         "emit the start message to the queue"
         return self.emit_message(
             article_id, version, run, "start",
-            "Starting ingest digest to endpoint for " + str(article_id))
+            "Starting ingest digest to endpoint for %s" % article_id)
 
     def emit_end_message(self, article_id, version, run):
         "emit the end message to the queue"
         return self.emit_message(
             article_id, version, run, "end",
-            "Finished ingest digest to endpoint for " + str(article_id))
+            "Finished ingest digest to endpoint for %s. Statuses: %s" % \
+                (article_id, self.statuses)
+        )
 
     def emit_error_message(self, article_id, version, run, message):
         "emit an error message to the queue"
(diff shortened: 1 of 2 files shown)


activity/activity_PublishDigest.py
```diff
@@ -72,9 +72,8 @@ def do_activity(self, data=None):
                 self.logger.info("Set Digest stage value of %s to published" % article_id)
                 put_response = digest_provider.put_digest_to_endpoint(
                     self.logger, digest_id, self.digest_content, self.settings)
-                if put_response:
-                    self.statuses["put"] = True
-                    self.logger.info("Put Digest for %s to the endpoint" % article_id)
+                self.statuses["put"] = True
```
giorgiosironi: the exception will prevent this from being executed, but will still be caught by the outer `try/except` construct

gnott: I see, if we get an unacceptable status code from the endpoint it will raise an exception. There's nothing we need to check in the `put_response` at this time about it being a success -- no exception is a success.

provider/digest_provider.py
```diff
@@ -199,6 +199,7 @@ def put_digest_to_endpoint(logger, digest_id, digest_content, settings):
         logger.exception(
             "Exception issuing PUT to the digest endpoint for digest_id %s. Details: %s" %
             (str(digest_id), str(exception)))
+        raise
```
gnott: Does the exception need to be re-raised here, or would it not flow through? No harm I guess

gnott:
I see the tests were not passing, I did some changes to the `FakeResponse` and the return value of the put now it does not check the response returned, and relies on the exception instead.

provider/digest_provider.py
giorgiosironi: Yes, unless Python has some specific behavior I don't remember, re-raising must be explicit as otherwise `except` swallows the exception

giorgiosironi:
Thank you, not sure what was happening to those tests locally.

PR #789, opened 2018-09-28
Only set the digest stage to "preview" if stage is missing.

gnott:
Fix for bug in comment https://github.com/elifesciences/elife-bot/pull/786#issuecomment-425573237

When I moved `set_stage()` over to the digest_provider and simplified it for `PublishDigest` activity to also used, I removed the logic needed here on ingestion: it should only set `stage` to `preview` if the digest doesn't have a value for it, because if it is a silent correction, the digest will already have a `stage` of `published`, and we do not want to overwrite it with `preview` when the digest is ingested.

### DIFF
--- a/activity/activity_IngestDigestToEndpoint.py
+++ b/activity/activity_IngestDigestToEndpoint.py
@@ -127,7 +127,8 @@ class activity_IngestDigestToEndpoint(Activity):
                     str(digest_id))
             self.digest_content = sync_json(self.digest_content, existing_digest_json)
             # set the stage attribute if missing
-            digest_provider.set_stage(self.digest_content, "preview")
+            if "stage" not in self.digest_content:
+                digest_provider.set_stage(self.digest_content, "preview")
             self.logger.info("Digest stage value %s" % str(self.digest_content.get("stage")))
 
             put_response = digest_provider.put_digest_to_endpoint(
--- a/activity/activity_IngestDigestToEndpoint.py
+++ b/activity/activity_IngestDigestToEndpoint.py
@@ -127,7 +127,7 @@ class activity_IngestDigestToEndpoint(Activity):
                     str(digest_id))
             self.digest_content = sync_json(self.digest_content, existing_digest_json)
             # set the stage attribute if missing
-            if "stage" not in self.digest_content:
+            if self.digest_content.get("stage") != "published":
                 digest_provider.set_stage(self.digest_content, "preview")
             self.logger.info("Digest stage value %s" % str(self.digest_content.get("stage")))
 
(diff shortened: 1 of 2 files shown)


activity/activity_IngestDigestToEndpoint.py
```diff
@@ -127,7 +127,8 @@ def do_activity(self, data=None):
                     str(digest_id))
             self.digest_content = sync_json(self.digest_content, existing_digest_json)
             # set the stage attribute if missing
-            digest_provider.set_stage(self.digest_content, "preview")
+            if "stage" not in self.digest_content:
```
giorgiosironi: I think we should design this check with more than two stages in mind to understand if our model works. Suppose we had as stages:

- `sent-to-typesetter`
- `preview`
- `approved-by-production` (comes from Libero, hopefully self-explanatory)
- `published`
- `deleted`

Stage `sent-to-typesetter` would only work if in its case we overwrite the `stage` field here. Silent corrections however should not touch the `published` state or the `deleted` state; and possibly also not touch `approved-by-production` to be really "silent". So there is no clear indication on whether to check states, the presence of the field, and so on.

I suggest to model 4 cases here:
- `stage` is missing (first ingest): overwrite
- `stage` is `preview` (other ingests): overwrite
- `stage` is `published` (silent corrections): preserve. This also makes it clear it's a silent correction, it's a positive check
- any other `stage` value: fail with an exception. If an additional stage is introduced, we immediately know there's something to check here.

gnott: I'm not aware of the other `stage` values but I see your line of thinking and there could be others in the future.

Does the code in the line highlight above satisfy the situations for `IngestDigestToEndpoint`? It is only adding a value when it is not present, otherwise it would be preserved. Unless it should also do a sanity check on the value of `stage` it got from the `existing_digest_json` at the point of ingestion to check for invalid values. 

The other situations might be more applicable to where `stage` is set in `PublishDigest`. It is very simplistic there, if `stage` is not `published` then it sets it (https://github.com/elifesciences/elife-bot/blob/develop/activity/activity_PublishDigest.py#L70), ignoring any `deleted` or unknown value for `stage`.

I think `sent-to-typesetter` as a stage is very remote, since we cannot ingest a stub that is not valid. We send to typesetter very early on and we cannot record a digest missing the required data it needs to be validated. I just thought I'd mention it here.

giorgiosironi: Yes, all additional states are hypothetical, not meant for implementation but for design reflection. I'd say be explicit there with:
```
if "stage" != "published":
```
which makes it clear the special case of preserving is only for already published digests, hence silent corrections.

gnott: Covers all the cases of `None` and `published` :+1: I changed it here - thanks!

gnott:
In `PublishDigest`, it will not issue a `PUT` if the digest is already `stage` `published`. I added a message to the log in these cases so it is not a mystery.

gnott:
In terms of the `stage` value, is this ready to approve and merge now @giorgiosironi ?

PR #791, opened 2018-10-01
Emit and log activity statuses values in the digest activities.

gnott:
Additional logging of `statuses` as described in https://github.com/elifesciences/elife-bot/pull/784#discussion_r221384702.

### DIFF
--- a/activity/activity_IngestDigestToEndpoint.py
+++ b/activity/activity_IngestDigestToEndpoint.py
@@ -139,7 +139,7 @@ class activity_IngestDigestToEndpoint(Activity):
             self.logger.exception("Exception raised in do_activity. Details: %s" % str(exception))
 
         self.logger.info(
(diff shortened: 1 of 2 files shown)


activity/activity_IngestDigestToEndpoint.py
```diff
@@ -138,6 +138,9 @@ def do_activity(self, data=None):
         except Exception as exception:
             self.logger.exception("Exception raised in do_activity. Details: %s" % str(exception))
 
+        self.logger.info(
+            "%s for digest_id %s statuses: %s" % (self.name, str(digest_id), self.statuses))
```
giorgiosironi: as in the other activity, `digest_id` may not be present if an exception has been thrown, hence needs the initialization to `None`?

gnott:
Changed the log message to use `article_id` rather than `digest_id` - the values would be the same in a success situation, and more informative if a digest fails to be created, the log will include the `article_id` that initiated the workflow for a better clue to track.

Issue #710, opened 2018-05-25
Simplify the activity VerifyPublishResponse logic


gnott:
There is some older logic in https://github.com/elifesciences/elife-bot/blob/develop/activity/activity_VerifyPublishResponse.py  based on the setting ``publication_authority`` value, which was either ``journal`` or ``elife-website``. Now that all EIF file logic is removed, we also should not need to use the ``elife-website`` publication authority again. The activity can be simplified to no longer consider the ``elife-website``.

@giorgiosironi I hope this sounds like a good idea. Do you think we should keep the ``publication_authority``  in the settings for potential future use? Or can the value also be removed?

giorgiosironi:
> we also should not need to use the elife-website publication authority again. The activity can be simplified to no longer consider the elife-website.

Agreed, go for it and remove what's obsolete

> Do you think we should keep the publication_authority in the settings for potential future use? Or can the value also be removed?

It's uncertain enough that it should be removed, and if we need something similar in the future we'll add another setting.

PR #793, opened 2018-10-02
Pass run_type as workflow data

gnott:
The `PostPerfectPublication` workflow is the last one to run in a chain of workflows in the article publication process. The earlier workflows read data from the `session`. Once it gets to `PostPerfectPublication` it relies on the data passed to the workflow when the workflow is started.

The data used in the workflow start comes from a `token` - a base64 encoded dict of values - that is returned by Lax when a publication messages is received by the bot's `lax_response_adapter`.

In order for Lax to return the token containing the `run_type` value (added here) we also need to add it to the `token` the `PublishToLax` activity sends to the queue for Lax to read.

`IngestToLax` is also updated, because it can bypass issuing the publish message to Lax in the case of a silent correction.

The starters of some of the workflows also need to be aware of the `run_type` attribute so they can start their workflows and make the value available to the activities within them.

`PublishToLax` activity had no tests, so I spent some time creating some with good coverage.

### DIFF
--- a/activity/activity_IngestToLax.py
+++ b/activity/activity_IngestToLax.py
@@ -42,6 +42,7 @@ class activity_IngestToLax(activity.activity):
         data['status'] = session.get_value('status')
         data['expanded_folder'] = session.get_value('expanded_folder')
         data['update_date'] = session.get_value('update_date')
+        data['run_type'] = session.get_value('run_type')
 
         queue_connection_settings = {"sqs_region": self.settings.sqs_region,
                                      "aws_access_key_id":self.settings.aws_access_key_id,
@@ -76,6 +77,7 @@ class activity_IngestToLax(activity.activity):
 
         try:
             expanded_folder = data['expanded_folder']
+            run_type = data.get('run_type')
 
             ##########
             if not consider_elife_20:
@@ -95,7 +97,8 @@ class activity_IngestToLax(activity.activity):
                                 "expanded_folder": expanded_folder,
                                 "requested_action": "",
                                 "message": "",
-                                "update_date": data['update_date']
+                                "update_date": data['update_date'],
+                                "run_type": run_type
                             }
                         }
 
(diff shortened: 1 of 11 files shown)


giorgiosironi:
This change is surgical, and doesn't seem to block anything digest-related; so I'll review and keep it here for now until all the digest changes have been deployed.

activity/activity_IngestToLax.py
```diff
@@ -76,6 +77,7 @@ def get_message_queue(self, data=None, consider_elife_20=True):
 
         try:
             expanded_folder = data['expanded_folder']
+            run_type = data.get('run_type')
```
giorgiosironi: can we be stricter and use `data['run_type']` which won't produce a `None` value?

activity/activity_IngestToLax.py
```diff
@@ -76,6 +77,7 @@ def get_message_queue(self, data=None, consider_elife_20=True):
...
 
             ##########
             if not consider_elife_20:
```
giorgiosironi: [`consider_Lax_elife_2_0` is always `True` now](https://github.com/elifesciences/builder-private/blob/master/salt/elife-bot/config/opt-elife-bot-settings.py) so we should remove this whole block?

gnott: I think so, but it was slightly off-topic to adding the `run_type`. It is a little like this issue too in my todo list of additional clean-up https://github.com/elifesciences/elife-bot/issues/710

activity/activity_IngestToLax.py
gnott: Along the same lines as a similar comment here, `run_type` is not always set, only when it is `silent-correction`, and assuming the value is in the data may raise an exception here.

gnott:
`run_type` is only set right now when a silent correction starter adds it. You made me think of another option though, to also add a `run_type` of `article-ingest` or something to the starter for `InitialArticleZip`, and if we did that the every workflow would have a `run_type`, and no need to have it as `None`. I like that too - what do you think?

giorgiosironi:
> if we did that the every workflow would have a `run_type`

I like consistency (you don't have to think about whether a field is available when writing activity code), but I understand it may take a while to roll out that across all the various workflows so it can be reached incrementally.

gnott:
All workflows of this type will have a `run_type` value now, `silent-correction` for silent correction ones, and `initial-article` for "normal" article ingest runs. Any additional comments @giorgiosironi before I merge?

PR #794, opened 2018-10-03
Ingest digests in silent corrections

gnott:
Run the IngestDigestToEndpoint activity as part of the silent corrections workflow.

It should be as simple as adding the activity to the workflow definition to get things going.

### DIFF
--- a/workflow/workflow_SilentCorrectionsProcess.py
+++ b/workflow/workflow_SilentCorrectionsProcess.py
@@ -79,6 +79,17 @@ class workflow_SilentCorrectionsProcess(workflow.workflow):
                         "start_to_close_timeout": 60 * 5
                     },
                     {
+                        "activity_type": "IngestDigestToEndpoint",
+                        "activity_id": "IngestDigestToEndpoint",
+                        "version": "1",
+                        "input": data,
+                        "control": None,
+                        "heartbeat_timeout": 60 * 5,
+                        "schedule_to_close_timeout": 60 * 5,
+                        "schedule_to_start_timeout": 300,
+                        "start_to_close_timeout": 60 * 5
+                    },
+                    {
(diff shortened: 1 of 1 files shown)


gnott:
In reference to issue https://github.com/elifesciences/issues/issues/4462, this PR may complete it.

giorgiosironi:
Yes but it will need an expansion of the end2end test.

workflow/workflow_SilentCorrectionsProcess.py
```diff
@@ -78,6 +78,17 @@ def __init__(self, settings, logger, conn=None, token=None, decision=None,
...
+                        "activity_id": "IngestDigestToEndpoint",
+                        "version": "1",
+                        "input": data,
+                        "control": None,
+                        "heartbeat_timeout": 60 * 5,
+                        "schedule_to_close_timeout": 60 * 5,
+                        "schedule_to_start_timeout": 300,
+                        "start_to_close_timeout": 60 * 5
+                    },
                     {
```
giorgiosironi: why was `PublishToLax` needed? I thought the reingestion activity would suffice in a silent correction. 
The placement of `IngestDigestToEndpoint` looks correct.

gnott: My guess is we include `PublishToLax` to follow steps similar to the regular article publication, and it is not a requirement of Lax to issue a publish force after an ingest force (I could be wrong here). 

The default is a message back from Lax as a result of the publish action is what sets the next bot workflow into motion. https://github.com/elifesciences/elife-bot/blob/develop/lax_response_adapter.py#L114 is where the `PostPerfectPublication` workflow is started after a publish message comes back from Lax. If we didn't rely on the reply from Lax we would have to start the `PostPerfectPublication` workflow as part of a silent correction in some other way.

PR #797, opened 2018-10-03
Remove IngestArticleZip

giorgiosironi:
Has been superceded by InitialArticleZip, I think. Today I tried to figure out the difference and only one is in `newFileWorkflows.yaml`.

Any other that could be removed like `workflow_Sum`?


gnott:
`IngestArticleZip` is needed right now, the last activity of `InitialArticleZip` - https://github.com/elifesciences/elife-bot/blob/develop/activity/activity_VersionReasonDecider.py - triggers it.

Recently we discontinued developing the version history project, so we could potentially merge these two workflows back to one, if we wanted to.

gnott:
`workflow_Sum` is an old test workflow, it can be removed if we want.

giorgiosironi:
So `InitialArticleZip` and `IngestArticleZip` are actually chained together? I can modify this to do the merge as it would prevent someone triggering the wrong (second) workflow in place of the correct one.

gnott:
The two are chained together, correct. The link in the chain basically just returns `True` each time - it was to wait and prompt for human input on the dashboard. I'm fairly sure we will not be using that hook, and for clarity it can be removed. `InitialArticleZip` is the one that responds as a result of a bucket notification (as mentioned in https://github.com/elifesciences/econ-tools readme example), and `IngestArticleZip` was changed to run from session data I think.

giorgiosironi:
Closing in favor of [a quick proof-of-concept on how the unification of the workflows would look like](https://github.com/elifesciences/elife-bot/pull/802).

PR #798, opened 2018-10-04
Add preview link to the end message emitted by IngestDigestToEndpoint.

gnott:
Implements issue https://github.com/elifesciences/issues/issues/4471

Depends on new settings to be added, in a separate PR.


settings-example.py
```diff
@@ -605,6 +611,9 @@ class live():
     digest_endpoint = 'https://digests/{digest_id}'
     digest_auth_key = 'digest_auth_key'
 
+    # journal preview
+    journal_preview_base_url = 'https://preview--journal.example.org'
```
giorgiosironi: do we need all of this environments in the example? Could it be just one or two (the second one extending the first)?

tests/activity/test_activity_ingest_digest_to_endpoint.py
```diff
@@ -318,6 +318,30 @@ def test_emit_start_message_no_connection(self):
...
+        expected = "https://preview/digests/" + article_id
+        preview_link = self.activity.digest_preview_link(article_id)
+        self.assertEqual(preview_link, expected)
+
+    def test_activity_end_message_ingest(self):
+        "activity end message after a digest ingest"
+        article_id = "00353"
+        statuses = {"ingest": True}
+        expected = ("Finished ingest digest to endpoint for 00353. Preview link" +
+                    " https://preview/digests/00353. Statuses {'ingest': True}")
```
giorgiosironi: the `.` after 00353 is a bit ambiguous, I think we should swap `Statuses` and the link so that the link comes last and can't accidentally be pasted with additional text

activity/activity_IngestDigestToEndpoint.py
```diff
@@ -180,11 +180,25 @@ def emit_start_message(self, article_id, version, run):
             article_id, version, run, "start",
             "Starting ingest digest to endpoint for " + str(article_id))
 
+    def digest_preview_link(self, article_id):
+        "preview link for the digest using the preview base url"
+        return self.settings.journal_preview_base_url + "/digests/%s" % article_id
+
+    def activity_end_message(self, article_id, statuses):
+        "different end message to emit based on the ingest status"
+        message = ''
```
giorgiosironi: initialization may not be needed (unless required by the linter?) as `message` is overwritten in both `if` branches. Or, the two branches can use `+=`

activity/activity_IngestDigestToEndpoint.py
```diff
@@ -180,11 +180,25 @@ def emit_start_message(self, article_id, version, run):
             article_id, version, run, "start",
             "Starting ingest digest to endpoint for " + str(article_id))
 
+    def digest_preview_link(self, article_id):
+        "preview link for the digest using the preview base url"
+        return self.settings.journal_preview_base_url + "/digests/%s" % article_id
```
giorgiosironi: for type safety on `+` (which gets angry when types like `None` or anything that is not a `str` is used) this could be
```
return "%s/digests/%s" % (self.settings.journal_preview_base_url, article_id)
```

settings-example.py
gnott: The `settings-example.py` file is a little old and just a demonstration of multiple environments and most (hopefully all) of the possible settings. Extending from a base environment might be confusing to new users, where they may thing that all environments would have the same values (most would not be common between environments). I think we were going to check and clean up this file but we didn't do it, not recently anyway. 

Is there a builder project that demonstrates the format of the `settings.py` file? I guess we should do a better job keeping https://github.com/elifesciences/elife-bot-formula/blob/master/salt/elife-bot/config/opt-elife-bot-settings.py more current with all the available settings. If we did then we could delete the example from this project.

Test cases use their own abbreviated settings files now. In the past this `settings-example.py` was copied to `settings.py` when running travis-ci tests, but no longer.

tests/activity/test_activity_ingest_digest_to_endpoint.py
gnott: It didn't look great with that `.` there for sure. I will move it to the end.

activity/activity_IngestDigestToEndpoint.py
gnott: I like it, thanks!

activity/activity_IngestDigestToEndpoint.py
gnott: The linter doesn't seem to mind either, my habit is to initialise when I start on things, and I don't go back to check whether it is required or not. I can remove it here though.

PR #800, opened 2018-10-05
Fix for clean_outbox in CopyDigestToOutbox

gnott:
Fixes https://github.com/elifesciences/issues/issues/4490

When deleting files from S3, the path must be added back to the file name first to get the full path.

### DIFF
--- a/activity/activity_CopyDigestToOutbox.py
+++ b/activity/activity_CopyDigestToOutbox.py
@@ -78,8 +78,9 @@ class activity_CopyDigestToOutbox(Activity):
         storage = storage_context(self.settings)
         files_in_bucket = storage.list_resources(resource_path)
         for resource in files_in_bucket:
-            self.logger.info("Deleting %s from the outbox", resource)
-            storage.delete_resource(resource)
+            orig_resource = resource_path + "/" + resource
+            self.logger.info("Deleting %s from the outbox", orig_resource)
+            storage.delete_resource(orig_resource)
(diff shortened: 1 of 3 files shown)


gnott:
The procedure for retrying an activity for unhandled exceptions is the default. The only expansion to it was making it possible to fail a workflow completely if that happens. 

We are not able to set a context like "try this activity three times, and if it is still failing, then return True / fail the workflow". It wouldn't be too hard to add something like this, but I would prefer if we could describe the use cases in a new issue and then add those improvements in a separate PR.

The bug here in `CopyDigestToOutbox` is something the automated tests are not testing properly due to how the mocked buckets are not creating subfolders on the local system. That could also be looked at as an improvement, because I don't like the disconnect between local `"object"` paths and the live bucket paths - they are not the same when tests are run.

I wouldn't expect this particular activity, `CopyDigestToOutbox`, to encounter an exception when cleaning the outbox except in a major networking or cloud service outage, so in this particular case I don't think we need to be concerned about the activity retrying once this bug fix is merged.

This is a blocker to developing a digest silent correction workflow.

giorgiosironi:
> We are not able to set a context like "try this activity three times, and if it is still failing, then return True / fail the workflow". It wouldn't be too hard to add something like this, but I would prefer if we could describe the use cases in a new issue

I agree this is a general issue, extracted into https://github.com/elifesciences/issues/issues/4527

PR #802, opened 2018-10-08
[wip] Implement POC of Initial/Ingest workflows unification

giorgiosironi:


gnott:
If you look at `VersionReasonDecider`, it actually starts an `IngestArticleZip` workflow, so if you delete that workflow, it would work. A full removal is probably best!

giorgiosironi:
I'm going to close this as outdated, the intention was good but I am following other priorities so don't have the time to take it to completion.

PR #804, opened 2018-10-12
Feature: bumps vulnerable versions

lsh-0:
cc @giorgiosironi

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,7 +18,7 @@ git+https://github.com/elifesciences/jats-generator.git@ea0d4d90a28198eb2953cc98
 git+https://github.com/elifesciences/package-poa.git@2649777a020661e507de4087d447f50ca8c2fe57#egg=packagepoa
 PyYAML==3.11
 Wand==0.4.2
-paramiko==2.4.1
+paramiko==2.4.2
 mock==1.3.0
 redis==2.10.5
 testfixtures==4.9.1
@@ -27,7 +27,7 @@ ddt==1.1.0
 Pillow==3.2.0
 ndg-httpsclient==0.4.2
 pyasn1==0.1.9
-pyOpenSSL==16.2.0
+pyOpenSSL==18.0.0
 newrelic==2.72.0.52
 pylint==1.6.4
 pyGithub==1.27.1


requirements.txt
```diff
@@ -18,7 +18,7 @@ git+https://github.com/elifesciences/jats-generator.git@ea0d4d90a28198eb2953cc98
 git+https://github.com/elifesciences/package-poa.git@2649777a020661e507de4087d447f50ca8c2fe57#egg=packagepoa
 PyYAML==3.11
 Wand==0.4.2
-paramiko==2.4.1
+paramiko==2.4.2
```
giorgiosironi: not sure if this applies but it's worth checking as once upon a time we upgraded it and got performance problems:
https://github.com/elifesciences/issues/issues/3416#issuecomment-402115053
(Also if upstream Paramiko is good enough again it may unblock Python 3 in builder)

lsh-0: changelog here: http://www.paramiko.org/changelog.html

it doesn't mention any fix for that

gnott:
There may be some repeatable errors in the end2end tests running, possibly as a result of this change, or possibly not. It could be related to some other infrastructure. It could also be related to some certificate validation. @giorgiosironi or @lsh-0 if you could please take a look. I'll run the tests again, but this re-run now is not looking bright. https://alfred.elifesciences.org/job/test-elife-bot/598/

lsh-0:
this looks like the error log from that run:
https://alfred.elifesciences.org/blue/rest/organizations/jenkins/pipelines/test-elife-bot/runs/598/nodes/32/steps/197/log/?start=0

looks like elife-dashboard on the face of it:

> E           TimeoutError: Cannot find 'article version 1 in status ready to publish on dashboard (run filter None, run_after filter 2018-10-16 22:05:40.069004): https://end2end--ppp-dash.elifesciences.org/api/article/5635053430208904637

lsh-0:
feel free to drop paramiko down a notch in your current build if you feel it's the cause or an unknown variable. The dependency bump passed all tests and the changelog had nothing major in it

gnott:
I noticed errors like `bad handshake: Error([(\'SSL routines\', \'tls_process_server_certificate\', \'certificate verify failed\')]` and assumed that may have been it, but I don't see the connection with paramiko, unless the tests also depend on that library. It could be something different. Certainly not passing though.

lsh-0:
> I don't see the connection with paramiko, unless the tests also depend on that library

the log I linked has 27 errors failing to connect to elife-dashboard during end2end tests. It passes the elife-bot highstate and tests fine but talking to the dashboard fails.

I just took a closer look at that `events` struct, I thought it was an expected response that would be compared against the actual response, but I see the glencoe error now:

`{u'event-message': u'An error occurred when checking/copying Glencoe still images. Article 5934833990957500790; message: ("bad handshake: Error([(\'SSL routines\', \'tls_process_server_certificate\', \'certificate verify failed\')],)",)',`

I thought paramiko dealt with ssh connections and the `requests` library doesn't appear to have any direct dependency on paramiko: https://github.com/requests/requests/blob/master/setup.py#L50-L56

Maybe Glencoe's certificate has expired? (I'm guessing we're talking to a test sever of theirs). Will investigate a bit more

lsh-0:
can't replicate locally, will try remotely on the machine itself

lsh-0:
hooray, a stacktrace:
```
>>> glencoe_check.metadata(569, settings)
From cffi callback <function _verify_callback at 0x7f0d7e41e668>:
  File "/opt/elife-bot/venv/local/lib/python2.7/site-packages/OpenSSL/SSL.py", line 309, in wrapper
    _lib.X509_up_ref(x509)
AttributeError: 'module' object has no attribute 'X509_up_ref'
  File "<stdin>", line 1, in <module>
  File "provider/glencoe_check.py", line 58, in metadata
    resp = requests.get(url)
  File "/opt/elife-bot/venv/local/lib/python2.7/site-packages/requests/api.py", line 70, in get
    return request('get', url, params=params, **kwargs)
  File "/opt/elife-bot/venv/local/lib/python2.7/site-packages/requests/api.py", line 56, in request
    return session.request(method=method, url=url, **kwargs)
  File "/opt/elife-bot/venv/local/lib/python2.7/site-packages/requests/sessions.py", line 475, in request
    resp = self.send(prep, **send_kwargs)
  File "/opt/elife-bot/venv/local/lib/python2.7/site-packages/requests/sessions.py", line 596, in send
    r = adapter.send(request, **kwargs)
  File "/opt/elife-bot/venv/local/lib/python2.7/site-packages/requests/adapters.py", line 497, in send
    raise SSLError(e, request=request)
requests.exceptions.SSLError: ("bad handshake: Error([('SSL routines', 'tls_process_server_certificate', 'certificate verify failed')],)",)
>>> 
```

lsh-0:
problem still exists after downgrading paramiko to 2.4.1

to be honest, I've seen these problems before and it's a system issue and gut feeling is because it's not receiving the daily system updates as it's off most the time

lsh-0:
tried curl, and curl worked fine, so probably not system. did an apt-get upgrade anyway and that didn't fix anything.

since it was just affecting python, I recreated the virtualenv and it now works fine. I'm doing a diff now of the differences between dependencies

lsh-0:
it's interesting. `pycrypto` is not present in new venv, `cryptography` got two minor version bumps and `asn1crypto` got a single minor version bump.


[venv.diff.txt](https://github.com/elifesciences/elife-bot/files/2485449/venv.diff.txt)

lsh-0:
giving build another kick. it's taking about 35-40mins per build

lsh-0:
all good: https://alfred.elifesciences.org/job/test-elife-bot/600/console

gnott:
:green_apple: :tada: :pray: !

giorgiosironi:
Very good investigation. Will the other environments need similar fixes, e.g. `elife-bot--prod` or `elife-bot--continuumtest`?

giorgiosironi:
Run a spectrum test on `continuumtest`, same error there:
https://continuumtest--ppp-dash.elifesciences.org/article/616873991717015893/1/f3b012db-5bc9-4666-82b6-04ef62fb9e78
```
    Article 616873991717015893; message: ("bad handshake: Error([('SSL routines', 'SSL3_GET_SERVER_CERTIFICATE', 'certificate verify failed')],)",) 
```
Same manual cleaning applies. https://alfred.elifesciences.org/job/prod-elife-bot/ should not be deployed until the same can be applied there too.

giorgiosironi:
Manually fixed `continuumtest` and `prod` to allow deployments.

`continuumtest` however had a disk full because logs were not rotated, I manually wrote an empty string to `worker.log` and `lax_response_adapter.log` to free space.

PR #805, opened 2018-10-23
Deposit all assets and invalidate the CDN prior to ingesting to Lax.

gnott:
Simple re-ordering of some activities in the article ingest workflows, it should hopefully help with refreshing caches before ingest to Lax.

In reference to ticket https://github.com/elifesciences/issues/issues/4458

It may cause some merge conflicts with changes in PR https://github.com/elifesciences/elife-bot/pull/802 if this PR is merged first, since it edits a workflow (`IngestArticleZip`) that will no longer exist if the workflows are unified.

### DIFF
--- a/workflow/workflow_IngestArticleZip.py
+++ b/workflow/workflow_IngestArticleZip.py
@@ -112,6 +112,28 @@ class workflow_IngestArticleZip(workflow.workflow):
                         "start_to_close_timeout": 60 * 15
                     },
                     {
+                        "activity_type": "DepositAssets",
+                        "activity_id": "DepositAssets",
+                        "version": "1",
+                        "input": data,
+                        "control": None,
+                        "heartbeat_timeout": 60 * 5,
+                        "schedule_to_close_timeout": 60 * 5,
+                        "schedule_to_start_timeout": 300,
+                        "start_to_close_timeout": 60 * 5
+                    },
+                    {
+                        "activity_type": "InvalidateCdn",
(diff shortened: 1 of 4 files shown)


giorgiosironi:
No problem for #802, I will rewrite it after this.

workflow/workflow_IngestArticleZip.py
```diff
@@ -111,6 +111,28 @@ def __init__(self, settings, logger, conn=None, token=None, decision=None,
                         "schedule_to_start_timeout": 300,
                         "start_to_close_timeout": 60 * 15
                     },
+                    {
```
giorgiosironi: will we be able to remove `DepositIngestAssets` or unify it with this one later? By the way, not a change to do now

gnott: We could review them to be sure, I think one was for the image server assets and the other is all the non-image server assets.

gnott:
I'm curious to see how end2end does, let's see.

gnott:
Two end2end failures in a row looks to be consistent. I wonder if the timeout for each article ingestion needs to be increased, or it could be for some other reason. Could you please take a look when you get a moment @giorgiosironi ? https://alfred.elifesciences.org/job/test-elife-bot/604/consoleFull

gnott:
Cancel my request to look at the tests please - I see the issue, that the `InvalidateCdn` activity expects to find values in workflow start data, but earlier on it will be in the session only. I'll fix this with more code.

PR #807, opened 2018-10-24
Feature article structure improvement

gnott:
In reference to issue https://github.com/elifesciences/issues/issues/4539, a more precise detection of the ArticleXML file if a similar looking file is part of the article files. I added in lots of linting as well, and some tests to cover other edge cases.

### DIFF
--- a/tests/provider/test_article_structure.py
+++ b/tests/provider/test_article_structure.py
@@ -80,7 +80,6 @@ class TestArticleStructure(unittest.TestCase):
         {'input': 'elife-00013-vor-v1-20121015000000.zip', 'expected': 'ArticleZip'},
         {'input': 'elife-00666-v1.pdf', 'expected': 'Other'},
         {'input': 'elife-00666-v1.xml', 'expected': 'ArticleXML'},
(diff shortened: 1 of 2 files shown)


tests/provider/test_article_structure.py
```diff
@@ -79,6 +80,7 @@ def test_get_version_from_zip_filename(self, input, expected):
         {'input': 'elife-00013-vor-v1-20121015000000.zip', 'expected': 'ArticleZip'},
         {'input': 'elife-00666-v1.pdf', 'expected': 'Other'},
         {'input': 'elife-00666-v1.xml', 'expected': 'ArticleXML'},
+        {'input': 'elife-00666-supp99.xml', 'expected': 'Other'},
```
giorgiosironi: this doesn't have the `-v1` suffix like the other cases, intentional?

gnott: It was the name we observed in the bucket. I could also add `elife-00666-supp99-v1.xml` to the list, which should pass, and improve the test coverage.

PR #808, opened 2018-10-25
ValidateDigestInput activity

gnott:
Fixes https://github.com/elifesciences/issues/issues/4537

Some refactoring to move the validation logic to the `digest_provider`, otherwise the changes planned on the ticket went fairly smoothly.

### DIFF
--- a/activity/activity_ValidateDigestInput.py
+++ b/activity/activity_ValidateDigestInput.py
@@ -64,11 +64,11 @@ class activity_ValidateDigestInput(Activity):
             self.input_file, self.temp_dir, self.logger, self.digest_config)
 
         # Approve files for emailing
-        self.statuses["valid"], error_message = digest_provider.validate_digest(self.digest)
+        self.statuses["valid"], error_messages = digest_provider.validate_digest(self.digest)
 
         if self.statuses.get("build") is not True or self.statuses.get("valid") is not True:
             # Send error email
-            self.statuses["email"] = self.email_error_report(real_filename, error_message)
+            self.statuses["email"] = self.email_error_report(real_filename, error_messages)
             self.log_statuses(self.input_file)
             return self.ACTIVITY_PERMANENT_FAILURE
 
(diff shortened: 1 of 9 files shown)


provider/digest_provider.py
```diff
@@ -272,3 +272,22 @@ def set_stage(json_content, stage="preview"):
...
+        error_message += '\nDigest was empty'
+    if digest_content and not digest_content.author:
+        is_valid = False
+        error_message += '\nDigest author is missing'
+    if digest_content and not digest_content.doi:
+        is_valid = False
+        error_message += '\nDigest DOI is missing'
+    if digest_content and not digest_content.text:
+        is_valid = False
+        error_message += '\nDigest text is missing'
```
giorgiosironi: In general I would build a `error_message` list and implode it with `"\n".join()` when completed; `if error_messages` would also give you `is_valid` as `True` if empty or `False` if non-empty.

To elaborate on that, could probably return a list of error strings from the provider (which is abstract code) so that the activity can decide to format them with a `\n`, or with HTML for an email, or to send a JSON to a dashboard, or use whatever other format.

activity/activity_ValidateDigestInput.py
```diff
@@ -0,0 +1,130 @@
...
+            self.settings, real_filename, bucket_name, bucket_folder, self.input_dir)
+
+        # Parse input and build digest
+        self.statuses["build"], self.digest = digest_provider.build_digest(
+            self.input_file, self.temp_dir, self.logger, self.digest_config)
+
+        # Approve files for emailing
+        self.statuses["valid"], error_message = digest_provider.validate_digest(self.digest)
+
+        if self.statuses.get("build") is not True or self.statuses.get("valid") is not True:
```
giorgiosironi: just need `not self.statuses.get("build")` as `False`, `None` etc would all result in the if being triggered

provider/digest_provider.py
gnott: I was thinking this too as I was editing, and I'll convert these to a list :+1:

PR #809, opened 2018-10-29
Digest silent deposit

gnott:
In reference to issue https://github.com/elifesciences/issues/issues/4479

The simplest solution I could think of is expressed here: silent in this context is do not email the `.docx` output, but continue to do the validation and depositing the files to the bucket outbox folder, and replace the digest ingest assets.

`EmailDigest` activity has access to the original bucket object name, which ends up as the `real_filename`. In `digest_provider.py`, look for case insensitive string `"silent"` in the file name, and if present, do not send an email in `EmailDigest`.

An alternative to this would be to save the `silent` status in a session somewhere, except the `IngestDigest` workflow is not using an session to read or write data at the moment, and reading it from the original workflow starter data is possible here for the activities that need to be silent or not.


activity/activity_EmailDigest.py
```diff
@@ -76,9 +76,13 @@ def do_activity(self, data=None):
...
         self.approve_status, error_messages = digest_provider.validate_digest(self.digest)
 
-        if self.approve_status is True and self.generate_status is True:
+        silent = digest_provider.silent_digest(real_filename)
+
+        if not silent and self.approve_status is True and self.generate_status is True:
             # Email file
             self.email_status = self.email_digest(self.digest, output_file)
+        elif silent:
+            self.logger.info('EmailDigest silent deposit of real_filename: %s', real_filename)
```
giorgiosironi: I would still send an email to the features team in the silent scenario. There is no way for someone without access to the logs to know when this process has finished and that is has exited correctly, given the dashboard does not display it. So an internal email will give confirmation the silent correction can now be made, and it will use the latest digest.

gnott: @giorgiosironi some additional discussion on https://github.com/elifesciences/issues/issues/4479 we could possibly add an additional activity at the end of the workflow to always send an email as a receipt of what was deposited. If we do that, I think it would be good to add it via a separate branch and PR.

Do you think the changes in the PR are suitable to merge at this time?

Alternatively we could add the new email receipt activity before we merge this PR.

giorgiosironi: Yes, no need to do it here, it's a good idea in general.

provider/digest_provider.py
```diff
@@ -286,3 +286,8 @@ def validate_digest(digest_content):
     if digest_content and not digest_content.text:
         error_messages.append('Digest text is missing')
     return not bool(error_messages), error_messages
+
+
+def silent_digest(filename):
+    "check if file name supplied to the bucket is a silent deposit"
+    return bool('silent' in str(filename).lower())
```
giorgiosironi:
```suggest
return bool(re.match(".*-silent.zip", str(filename).lower()))
```

gnott: Being more strict here is welcome 👍

PR #820, opened 2018-11-21
Post digest JATS

gnott:
In reference to issue https://github.com/elifesciences/issues/issues/4538. 

Includes the new activity, converts digest `.docx` data to JATS using the `digest-parser` library, adds it and some other values to a dictionary, and send that as data in a POST request to an endpoint.

The new activity is not added to a workflow yet, which can be in a later PR, when we are ready to try it.

I hope the code here is complete enough for code review of this portion, and it may indicate how some end2end testing might be possible for it.


activity/activity_PostDigestJATS.py
```diff
@@ -0,0 +1,130 @@
...
+
+        # Local directory settings
+        self.temp_dir = os.path.join(self.get_tmp_dir(), "tmp_dir")
+        self.input_dir = os.path.join(self.get_tmp_dir(), "input_dir")
+
+        # Create output directories
+        self.create_activity_directories()
+
+        # Track the success of some steps
+        self.statuses = {
```
giorgiosironi: this variable seems local to `do_activity()` so it could be encapsulated there rather than as a field on the `activity` object

activity/activity_PostDigestJATS.py
```diff
@@ -0,0 +1,130 @@
...
+        self.jats_content = self.digest_jats(self.input_file)
+
+        if self.jats_content:
+            self.statuses["jats"] = True
+
+        # POST to API endpoint
+        try:
+            self.statuses["post"] = self.post_jats(self.digest, self.jats_content)
+        except Exception as exception:
+            self.logger.exception("Exception raised in do_activity. Details: %s" % str(exception))
```
giorgiosironi: safe rollout, failure of `post` doesn't affect the activity and its workflow

activity/activity_PostDigestJATS.py
```diff
@@ -0,0 +1,130 @@
...
+    payload["type"] = content_type
+    payload["content"] = jats_content
+    return payload
+
+
+def post_jats_to_endpoint(url, payload):
+    "issue the POST"
+    resp = requests.post(url, data=payload)
+    # Check for good HTTP status code
+    if resp.status_code != 200:
```
giorgiosironi: logging the response status code, and the body if there should be anything in there, can help debugging

activity/activity_PostDigestJATS.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def post_payload(digest, jats_content, api_key):
+    "compile the POST data payload"
+    account_key = 1
+    content_type = "digest"
+    payload = OrderedDict()
+    payload["apiKey"] = api_key
+    payload["accountKey"] = account_key
+    if digest:
```
giorgiosironi: arguably if there is no `digest` this should fail hard, not attempting to do the request at all
(the failure would still be caught in the activity for now)

activity/activity_PostDigestJATS.py
```diff
@@ -0,0 +1,130 @@
...
+
+        if self.statuses.get("jats"):
+            return self.ACTIVITY_SUCCESS
+
+        return self.ACTIVITY_PERMANENT_FAILURE
+
+    def digest_jats(self, docx_file):
+        "generate the digest JATS content from the docx file"
+        try:
+            return digest_provider.build_jats(docx_file, self.temp_dir, self.digest_config)
```
giorgiosironi: that's a strange contract, I would expect JATS to be able to be generated purely from `self.digest` rather than from `docx_file`. That way you decouple input (currently `.docx` but can also be an in-memory Python object generated for a test) from output (a JSON for the `digests` service, a JATS for the typesetter); all the combinations work.

activity/activity_PostDigestJATS.py
gnott: Normally it should be shielded from having no digest because the input is validated in a normal workflow before it reaches this point, but I do have some test cases with bad `.docx` input and this was a quick fix. I agree if there is no `digest` object the payload will be useless, and we can return `None` first if there is no `digest` here 👍

activity/activity_PostDigestJATS.py
gnott: The `digestparser` library it is using, you can generate a digest object from the `.docx` input file, or you can generate the JATS from the file (but it internally just generates a digest object and then converts it to JATS). 

I can simplify it as you describe, where it creates the digest object, as normal, and then can call `digestparser.jats.digest_jats(digest)` instead.

activity/activity_PostDigestJATS.py
gnott: Agreed, more logging.

activity/activity_PostDigestJATS.py
giorgiosironi: Understood, the return value is opaque. A story for another time: a Listener class that can collect information about what's going on and satisfy testing, logging, monitoring requirement.

activity/activity_PostDigestJATS.py
giorgiosironi: Would still log something about the successful request, with `INFO` level; you expect to see it as a positive confirmation that the activity has been executed (as opposed to being queued, interrupted, etc)

activity/activity_PostDigestJATS.py
```diff
@@ -0,0 +1,129 @@
...
+        if self.statuses.get("jats"):
+            return self.ACTIVITY_SUCCESS
+
+        return self.ACTIVITY_PERMANENT_FAILURE
+
+    def post_jats(self, digest, jats_content):
+        "prepare and POST jats to API endpoint"
+        url = self.settings.typesetter_digest_endpoint
+        payload = post_payload(digest, jats_content, self.settings.typesetter_digest_api_key)
+        if payload:
```
giorgiosironi: this also seems like something that should always be true (may need to be stubbed in tests)

activity/activity_PostDigestJATS.py
gnott: More logging is always helpful 👍

giorgiosironi:
(still approved)

gnott:
Thanks for confirming it is still approved for merging. Low risk of catastrophe, I will merge it for the next set of testing.

PR #821, opened 2018-11-22
Starters and python 3

gnott:
Continuation of improving Python 3 support, I looked at getting the tests for the starters to pass.

The imports for `starter_helper` are improved with a full path used.

`base64` uses `encodebytes` since `encodestring` is deprecated, which is handled here by a new `base64_encode_string()` function.

`bytes` has no `__dict__` so when json dumping to string, the default function would now return `None` instead of (presumably) a dict. It should have no effect how it is used normally, as long as the activity `data` is checked to be non-None before trying to get values from it.

### DIFF
--- a/starter/starter_SilentCorrectionsProcess.py
+++ b/starter/starter_SilentCorrectionsProcess.py
@@ -8,8 +8,8 @@ import log
 import json
 import random
 from optparse import OptionParser
-import starter_helper as helper
-from starter_helper import NullRequiredDataException
+import starter.starter_helper as helper
+from starter.starter_helper import NullRequiredDataException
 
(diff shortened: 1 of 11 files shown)


starter/starter_SilentCorrectionsProcess.py
```diff
@@ -8,8 +8,8 @@
 import json
 import random
 from optparse import OptionParser
-import starter_helper as helper
-from starter_helper import NullRequiredDataException
+import starter.starter_helper as helper
+from starter.starter_helper import NullRequiredDataException
```
giorgiosironi: all the imports sound good

PR #823, opened 2018-11-23
Python 3 constructors and fix test running errors

gnott:
I first wanted to fix all incompatible `print` statements for python 3, then I wanted to fix all the test case errors (not the failures, just the things causing an error so tests could not be run), and most of the errors were to switch to the newer object constructor / init syntax.

There are also some changes for bytes versus strings, and some library and object property differences.

The automated tests are passing in Python 2, and hopefully if it looks ok, we can try it on end2end testing later to be sure I didn't break anything.


activity/activity_AcceptVersionReason.py
```diff
@@ -1,14 +1,15 @@
-import activity
 from provider.execution_context import get_session
+from .activity import Activity
```
giorgiosironi: looks fishy, is it a modern format? Should it be replaced with `from activity.activity import Activity` which is an absolute import?

provider/article.py
```diff
@@ -206,7 +206,11 @@ def get_tweet_url(self, doi):
         """
         doi_url = self.get_doi_url(doi)
         f = {"text": doi_url + " @eLife"}
-        tweet_url = "http://twitter.com/intent/tweet?" + urllib.urlencode(f)
+        if hasattr(urllib, 'urlencode'):
+            tweet_url = "http://twitter.com/intent/tweet?" + urllib.urlencode(f)
+        else:
+            # python 3
+            tweet_url = "http://twitter.com/intent/tweet?" + urllib.parse.urlencode(f)
```
giorgiosironi: do you need to import `urllib.parse` in the specific rather than just `urllib`? It is a [separate module](https://docs.python.org/3/library/urllib.parse.html#module-urllib.parse) but not clear to me.

gnott: I think it is not required to explicitly import `parse`, since the base module `urllib` is imported. It seems to work and is covered when running the tests in Python 3. 

I guess another way would be to import different libraries based on the version of python, or check if the `parse` module exists before importing, or use exception catching. The method above I've used in a couple places and it seems to work.

giorgiosironi:
> I cheated by deleting the old /opt/elife-bot/activity/activity.pyc

:cold_sweat: You know what to do: SSH and rm. Rare case, a Docker container image would still solve it

gnott:
Docker container advantage is a good one.

Any additional comments or opinion whether the code in this PR is ready to merge?

giorgiosironi:
Feel free to merge and deploy, keeping an eye on `test-elife-bot` and `prod-elife-bot` for brekages related to the `.pyc` file.

gnott:
Thanks, I will attempt it now.

PR #824, opened 2018-11-27
Transmitting digest JATS to endpoint

gnott:
This method of sending digest JATS to the endpoint is tested and currently works.

Waiting to see if the method will change after more discussion.

### DIFF
--- a/activity/activity_PostDigestJATS.py
+++ b/activity/activity_PostDigestJATS.py
@@ -126,7 +126,7 @@ def post_payload(digest, jats_content, api_key):
 
 def post_jats_to_endpoint(url, payload, logger):
     "issue the POST"
-    resp = requests.post(url, data=payload)
+    resp = requests.post(url, params=payload)
     # Check for good HTTP status code
     if resp.status_code != 200:
         logger.error(
--- a/activity/activity_PostDigestJATS.py
+++ b/activity/activity_PostDigestJATS.py
@@ -126,7 +126,7 @@ def post_payload(digest, jats_content, api_key):
 
 def post_jats_to_endpoint(url, payload, logger):
     "issue the POST"
-    resp = requests.post(url, params=payload)
+    resp = post_as_params(url, payload)
     # Check for good HTTP status code
     if resp.status_code != 200:
         logger.error(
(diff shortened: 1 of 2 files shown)


activity/activity_PostDigestJATS.py
```diff
@@ -126,7 +126,7 @@ def post_payload(digest, jats_content, api_key):
 
 def post_jats_to_endpoint(url, payload, logger):
     "issue the POST"
-    resp = requests.post(url, data=payload)
+    resp = requests.post(url, params=payload)
```
giorgiosironi: This seems to be related to encoding of the parameters: https://stackoverflow.com/a/53320437/91590
If it's this tricky, a test should be added that fails before the change and passes after the change.

gnott: I didn't look carefully enough to see if the values are encoded differently. If they are, I would hope they decode just as effectively, such as if you had a web form with `method="get"` or `method="post"` you do eventually get the same values in the end when accessing them from the request.

I'll see what we might be able to do in automated tests except I think we have a very basic mock object that returns an HTTP status code and maybe a body. Maybe it can dip into the actual `requests` library and we can try to mock its methods if we want to compare passing `data` or `params`.

gnott: I arrived at one way to test it: refactor the requests call so it is easier to test, write tests that mock the actually HTTP connection, and then test the `response.request` object.

There are three example test scenarios if the payload is sent as `params`, `data`, or `json`.

Preference being we send it by `data` calling `post_as_data(url, payload)`, but right now the only method I have confirmed is working with the live endpoint is the `post_as_params(url, payload)`.

tests/activity/test_activity_post_digest_jats.py
```diff
@@ -192,5 +192,60 @@ def test_post_payload_no_digest(self):
...
+
+    @patch('requests.adapters.HTTPAdapter.get_connection')
+    def test_post_as_params(self, fake_connection):
+        "test posting data as params only"
+        url = 'http://localhost/'
+        payload = OrderedDict([
+            ("type", "digest"),
+            ("content", '<p>"98%"β</p>')
+            ])
+        expected_url = url + '?type=digest&content=%3Cp%3E%2298%25%22%CE%B2%3C%2Fp%3E'
```
giorgiosironi: this is very descriptive of what is happening, rather than relying on `requests` doing the right thing according to the arguments being passed to it

gnott: It is much clearer for sure. Hopefully we can switch to `data` from `params` in the future, but we can start sending in the format that works at this time.

Issue #827, opened 2018-11-29
Increase test coverage on some activity objects


gnott:
When refactoring some activity code, there were some activity objects with no automated test cases at all. Seven of these can use at least some basic tests to check for library imports or typos at a minimum:

- [x] Basic automated tests for `activity/activity_AdminEmailHistory.py`
- [x] Automated tests and refactor storage_context for `activity/activity_ArchiveArticle.py`
- [x] Basic automated tests for `activity/activity_PingWorker.py`
- [x] Basic automated tests for `activity/activity_S3Monitor.py`
- [x] Basic automated tests for `activity/activity_ScheduleDownstream.py`
- [x] Basic automated tests for `activity/activity_SendQueuedEmail.py`
- [x] Basic automated tests for `activity/activity_VersionReasonDecider.py`
- [x] Increase test coverage for `activity/activity_PubRouterDeposit.py` (currently 20% coverage)
- [x] `activity/activity_Sum.py` also has no tests, but I am suggesting to delete it entirely.

giorgiosironi:
FWIW the more you can extract into a `provider` (not tied to SWF) the easier it will be to test and incidentally to reuse across activities.

gnott:
I was hoping to add some quick tests to be more confident all the Python 3 compatibility can be checked. 

After trying this on `ArchiveArticle` it was not that easy, and some refactoring to use the `storage_context` will be an improvement there, instead of using S3 and bucket keys directly. So some refactoring might be another good output from the attempt for test coverage.

These will probably be best added as a separate PR for each activity, otherwise it will take too long to get all these done in one branch.

gnott:
I think all activities have at least some tests now, fairly good coverage on many of them. The motivation of this issue for me was to make sure the Python 3 compatibility effort had the best chance of finding all the potential problems when running automated tests, rather than having gaping holes in the coverage.

PR #829, opened 2018-11-30
Tests for AdminEmailHistory activity.

gnott:
In reference to https://github.com/elifesciences/elife-bot/issues/827, a simple test scenario for coverage on `AdminEmailHistory`.

### DIFF
--- a/tests/activity/test_activity_admin_email_history.py
+++ b/tests/activity/test_activity_admin_email_history.py
@@ -1,0 +1,25 @@
+import unittest
+from mock import patch
+from provider.simpleDB import SimpleDB
+import provider.swfmeta as swfmetalib
+from activity.activity_AdminEmailHistory import activity_AdminEmailHistory as activity_object
+import tests.activity.settings_mock as settings_mock
+from tests.activity.classes_mock import FakeLogger
+
+
+class TestAdminEmailHistory(unittest.TestCase):
+
+    def setUp(self):
+        fake_logger = FakeLogger()
+        self.activity = activity_object(settings_mock, fake_logger, None, None, None)
+
+    @patch.object(swfmetalib.SWFMeta, 'get_closed_workflow_execution_count')
+    @patch.object(SimpleDB, 'elife_add_email_to_email_queue')
+    def test_do_activity(self, fake_email, fake_workflow_count):
+        fake_workflow_count.return_value = 0
+        success = self.activity.do_activity()
+        self.assertEqual(success, True)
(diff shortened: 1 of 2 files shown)


tests/activity/test_activity_admin_email_history.py
```diff
@@ -0,0 +1,25 @@
...
+from tests.activity.classes_mock import FakeLogger
+
+
+class TestAdminEmailHistory(unittest.TestCase):
+
+    def setUp(self):
+        fake_logger = FakeLogger()
+        self.activity = activity_object(settings_mock, fake_logger, None, None, None)
+
+    @patch.object(swfmetalib.SWFMeta, 'get_closed_workflow_execution_count')
```
giorgiosironi: if you find yourself reusing `swfmetalib` elsewhere, try to wrap it into a `workflow_provider`; or more specifically a `statistics_provider` if it's to get counters out of it

tests/activity/test_activity_admin_email_history.py
```diff
@@ -0,0 +1,25 @@
...
+    @patch.object(SimpleDB, 'elife_add_email_to_email_queue')
```
giorgiosironi: unfortunate that there are multiple ways to send emails in place, but this test achieve the goal which is basic coverage of the activity to allow Python 3 smoke testing

gnott: Yes, my intent with these is just to check for typos and slip ups when changes are made for Python 3 support. Email usage, as well as many other potential improvements, are possible in future PRs.

tests/activity/test_activity_admin_email_history.py
gnott: The `swfmeta.py` module is worthy of looking at in more detail. The alternate naming is possible, maybe in a later update? I will open a separate issue for it. 

Outside of a small bug fix and writing test cases for `swfmeta.py`, it has largely not been changed since 2013. In a quick review of it, half the code is not used (anything looking at open workflow executions), and counting open workflow executions is only used by the `AdminEmailHistory` activity.

It also makes me think with the current situation of instance monitoring, this entire `AdminEmailHistory` activity could be removed and it would be fine. I'm the only one that gets it, and it has little value. I'll open another ticket for that!

gnott:
Thanks for looking and approval. I created some new issues to address provider renaming and possible removal of this activity. I will merge the PR now to not get caught with code errors due to poor coverage.

PR #833, opened 2018-12-05
Schedule downstream refactoring and tests

gnott:
In reference to issue https://github.com/elifesciences/elife-bot/issues/827 for more tests on bot activities.

Started off with refactoring all the native `S3Connection` usage to use the `storage_context`.

`storage.copy_resource()` is used in other activities to copy objects from one bucket to another, but the mocked `copy_resource()` method doesn't actually copy any data that can be checked in test assertions right now.

Continued to refactor some of the object properties, moved them out to separate functions.

I also included moving `download_article_xml()` to the `article_processing.py` provider, except I didn't need it here since this activity only gets the XML key name and doesn't need to download the XML to copy bucket to bucket.

### DIFF
--- a/tests/activity/test_activity_schedule_downstream.py
+++ b/tests/activity/test_activity_schedule_downstream.py
@@ -1,0 +1,31 @@
+import unittest
+from mock import mock, patch
+import activity.activity_ScheduleDownstream as activity_module
+from activity.activity_ScheduleDownstream import activity_ScheduleDownstream as activity_object
+from provider import article
+import tests.activity.settings_mock as settings_mock
+from tests.activity.classes_mock import FakeLogger, FakeStorageContext
+import tests.activity.test_activity_data as activity_test_data
+
+
+class TestScheduleDownstream(unittest.TestCase):
+
+    def setUp(self):
+        fake_logger = FakeLogger()
+        self.activity = activity_object(settings_mock, fake_logger, None, None, None)
+
+    @patch.object(article, 'storage_context')
+    @patch.object(activity_module, 'storage_context')
+    def test_do_activity(self, fake_activity_storage_context, fake_storage_context):
+        expected_result = True
+        fake_storage_context.return_value = FakeStorageContext()
+        fake_activity_storage_context.return_value = FakeStorageContext()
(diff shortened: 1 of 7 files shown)


gnott:
On this and the earlier PR, if you could please ignore the result of the travis-ci tests and only consider Alfred, it seems like travis-ci is temporarily broken.

giorgiosironi:
Workarounds for Travis CI listed in https://github.com/travis-ci/travis-ci/issues/7940 where tons of broken builds in these months have gathered for the blame conga.

tests/activity/test_activity_schedule_downstream.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def setUp(self):
+        fake_logger = FakeLogger()
+        self.activity = activity_object(settings_mock, fake_logger, None, None, None)
+
+    @patch.object(article, 'storage_context')
+    @patch.object(activity_module, 'storage_context')
+    def test_do_activity(self, fake_activity_storage_context, fake_storage_context):
+        expected_result = True
+        fake_storage_context.return_value = FakeStorageContext()
```
giorgiosironi: ideally these should be the same `FakeStorageContext` object? Not that it makes a difference

gnott:
I've had the travis-ci error before, you can see in https://github.com/elifesciences/elife-bot/blob/develop/travis-install.sh#L3 I already added `export BOTO_CONFIG=/dev/null` . I can try removing the `sudo: false` as a test, it can't be much worse.

gnott:
Moved `export BOTO_CONFIG=/dev/null` to the travis test script instead of the install script, also `sudo: false` is gone, the build seems to be working for now.

tests/activity/test_activity_schedule_downstream.py
gnott: I agree in the testing it doesn't matter too much. The two places where `storage_context` is mocked would I think be different S3 connections in a non-mocked environment. First the `article` provider would get a list of files in the expanded bucket, then the activity would connect to the destination bucket and copy a key from the expanded bucket there.

gnott:
I will merge it and hopefully we have end2end tests to cover it in case I made a typo somewhere.

PR #834, opened 2018-12-06
Bug fix, add expanded folder name to old key in ScheduleDownstream.

gnott:
end2end testing is failing on `ScheduleDownstream`, the origin key name is wrong. Adding back the expanded bucket folder name to the origin key so the copy operation will work.


gnott:
Bug fix to code in PR https://github.com/elifesciences/elife-bot/pull/833

Issue #696, opened 2018-05-03
Refactor or remove lettuce tests


gnott:
The lettuce tests in the project are old and no new tests are written. There could be some lettuce tests testing some good things, and if so they can be rewritten as unit tests. Then delete all the lettuce test files.

gnott:
Related internal issue https://github.com/elifesciences/issues/issues/1254

PR #841, opened 2019-01-09
Lettuce tests bye bye

gnott:
Fixes https://github.com/elifesciences/elife-bot/issues/696
Fixes https://github.com/elifesciences/issues/issues/1254

Many files and changes included here, though they are as straightforward as is possible (I think).

Lettuce tests removed - includes deleting all files in the `/tests/features` folder, removing `/tests/terrain.py`, `lettuce` no longer in requirements, not running lettuce tests as part of the project tests.

Existing unittest test cases were expanded (`S3Monitor` activity, `article` provider, `ejp` provider, `article_structure` provider). New tests for untested parts are added (for `SimpleDB` provider, `worker`, `decider`).

There are some functions deleted from the `article` provider when I determined they are no used anymore.

To make testing the `worker` and `decider` easier when importing the `settings` module, I moved instantiating settings to the `__main__()` method, as we have done in other long running processes. I think it is all correct, and if not I think the end2end tests will catch any errors anyway.

There was also a strange glitch in automated tests for the `worker`, when it reloads the `PingWorker` activity, it then later causes the tests for the `PingWorker` activity to fail. I changed it so reloading the module upon import is a flag, and when testing reload is `False`.


decider.py
```diff
@@ -220,4 +216,8 @@ def get_workflow_object(workflow_name, settings, logger, conn, token, decision,
     (options, args) = parser.parse_args()
     if options.env:
         ENV = options.env
-    process.monitor_interrupt(lambda flag: decide(ENV, flag))
+
+    settings_lib = __import__('settings')
```
giorgiosironi: avoiding a dependency on `settings`, but a dynamic import is also a smell; reason why it can't stay at the top?

project_tests.sh
```diff
@@ -1,9 +1,6 @@
 #!/bin/bash
 set -e
 source venv/bin/activate
-cd tests/
-for i in `ls features/*.feature`; do echo $i; lettuce $i; done
-cd -
```
giorgiosironi: Lettuce tests :wave:

provider/article.py
```diff
@@ -578,18 +419,6 @@ def get_s3_key_names_from_bucket(self, bucket_name, prefix, file_extensions):
...
-        For each article XML downloaded from S3, check if it is published
-        """
-
-        doi_id = self.get_doi_id(doi)
-
-        if int(doi_id) in self.get_was_poa_doi_ids():
-            return True
-        else:
-            return False
-
```
giorgiosironi: if it's just dead code :+1:

worker.py
```diff
@@ -164,7 +160,8 @@ def import_activity_class(activity_name):
         module_name = "activity." + activity_name
         importlib.import_module(module_name)
         # Reload the module, in case it was imported before
-        reload_module(module_name)
+        if reload:
```
giorgiosironi: seems related to re-importing settings? Why is this needed? In theory the activity should receive settings dynamically (in the constructor?) and its code does not change so reloading also smells as patching some underlying problem

gnott: I think this is reloading the particular activity module. It was useful when developing locally and running workflows, so a code change is noticed each time the code is edited. It probably adds little in the production environment, where the worker restarts will reload the activity code. Possibly I could delete the reloading and it might be better, do you agree?

decider.py
gnott: Tests failed when `settings.py` is missing from the root directory of the project, which is not included in the repo. The worker is not encapsulated in a class, and I could not `import worker` without satisfying the `import settings ...` in `worker.py`.

For some reason the lettuce tests didn't have this problem, I'm not sure why.

Previously in travis-ci tests, I copied `settings-example.py` to `settings.py` before the tests were run, but did away with that a while ago when all the new tests used the `settings_mock.py`.

We could rework how the settings work (long overdue probably) or we could try to ignore the smell for this PR.

We pass `settings` to just about every other class or function, so I guess the only difference is importing the settings module at specific times, when it actually exists.

Possibly this could be fixed too by creating a `Worker` class, that seems to avoid the settings import for activity classes, etc.

provider/article.py
gnott: I search for any current usage and there was none I could find. Anything that wants this looks at Lax data now, instead of data that is in bucket folders.

decider.py
giorgiosironi: This I would postpone

worker.py
giorgiosironi: Yes, if we want to have a development workflow with code reloading we should target a more general solution like [reloadr](https://github.com/hoh/reloadr) (just the first one that came up)

gnott: I removed `reload_module()` and its usage from `worker.py` and `decider.py`.

If all tests pass, do you suppose this is worthy to merge now?

giorgiosironi: Yes!

PR #842, opened 2019-01-10
Python 3 updates for no errors

gnott:
In reference to issue https://github.com/elifesciences/issues/issues/4062, Python 3 compatibility.

The changes here allow me to run almost all the test cases with no errors in Python 3.5. Tests are also compatible in Python 2.7.

Mostly fixed by opening files in non-binary mode, and converting bytes to string when required.

I updated the EJP csv file for the list of authors data to reflect what the csv files look like now, which is where non-latin-1 characters are entities, and no longer in some strange encoding. I updated the `ejp.py` provider parsing to convert the entities to unicode, and then the values flow into `PublicationEmail`.

Other assorted code changes are different in Python 3, no `.iterkeys()`, no `unittest.TestCase. assertItemsEqual()`, no `.next()`, no `long`.

A few fixed test cases to not delete a `.gitkeep` file when the tests are run.

Also removed `Fom` from requirements, we do not need.


gnott:
Resolved conflict with the lettuce test cleanup, and fixed running simpledb provider tests in Python 3. I corrected the original description - not all tests pass in Python 3, there is still an issue with extending `workflow.workflow` objects, which can be addressed in a new PR. The renaming of the base `workflow` file for easier imports is part of the overall Python 3 compatibility ticket.

PR #848, opened 2019-01-16
More test coverage for starters

gnott:
Working from a list of files that have no test cases, the starters and cron starters here had no tests at all. The additional motivation for them is to have more confidence in automated testing for Python 3 support.

The parentheses around the `print` statements are duplicated here as is in another PR under review, but it was the easiest way to run the tests in Python 3 without having to wait for the other code to be merged. I expect the two should merge easily enough with no conflicts (if not I'll resolve the conflicts after the first PR is merged).

The other non-test code change is to `str(limit)` in the `SendQueuedEmail` starter, although the activity doesn't do anything with this limit value anyway right now.

### DIFF
--- a/starter/starter_SendQueuedEmail.py
+++ b/starter/starter_SendQueuedEmail.py
@@ -32,7 +32,7 @@ class starter_SendQueuedEmail():
         execution_start_to_close_timeout = None
 
         if limit:
-            input = '{"data": {"limit": "' + limit + '"}}'
+            input = '{"data": {"limit": "' + str(limit) + '"}}'
         else:
             input = None
 
@@ -48,7 +48,7 @@ class starter_SendQueuedEmail():
             # There is already a running workflow with that ID, cannot start another
             message = ('SWFWorkflowExecutionAlreadyStartedError: There is already ' +
                        'a running workflow with ID %s' % workflow_id)
-            print message
+            print(message)
             logger.info(message)
 
 if __name__ == "__main__":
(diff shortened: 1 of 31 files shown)


starter/starter_SendQueuedEmail.py
```diff
@@ -32,7 +32,7 @@ def start(self, settings, limit=None):
         execution_start_to_close_timeout = None
 
         if limit:
-            input = '{"data": {"limit": "' + limit + '"}}'
+            input = '{"data": {"limit": "' + str(limit) + '"}}'
```
giorgiosironi: this can only improves safety, as concatenation would fail with `limit` not being a string

tests/starter/test_starter_send_queued_email.py
```diff
@@ -0,0 +1,34 @@
...
+from tests.classes_mock import FakeLayer1
+import tests.settings_mock as settings_mock
+from mock import patch
+
+
+class TestStarterSendQueuedEmail(unittest.TestCase):
+    def setUp(self):
+        self.starter = starter_SendQueuedEmail()
+
+    @patch('boto.swf.layer1.Layer1')
```
giorgiosironi: tons of mocking, but I don't see a way to build a generic starter (rather than many) to save on the amount of code that is being written and maintained

gnott: Yes, at first I thought I might be able to test in a general way until I found some starters that are different, it is hopefully clearer to have one test module per starter and keep them simple for now.

PR #849, opened 2019-01-17
Queue workflow starter tests

gnott:
One of the last remaining parts with no test in the `elife-bot` project is `queue_workflow_starter.py`.

I moved the importing and loading of the `settings` module for easier testing, as is done on many of the parts that are run like console apps.

`json.loads()` didn't work too well in Python 3 until the bytes string was decoded.

### DIFF
--- a/queue_workflow_starter.py
+++ b/queue_workflow_starter.py
@@ -73,11 +73,6 @@ def start_workflow(settings, workflow_name, workflow_data):
         workflow_data = data_processor(workflow_name, workflow_data)
     module_name = "starter." + workflow_name
     module = importlib.import_module(module_name)
-    # TODO: deprecate reloading a module, this should not be needed as every deploy restarts all long-running processes?
-    try:
-        reload(eval(module))
-    except:
-        pass
     full_path = "starter." + workflow_name + "." + workflow_name + "()"
     s = eval(full_path)
     s.start(settings=settings, **workflow_data)
--- a/queue_workflow_starter.py
+++ b/queue_workflow_starter.py
@@ -27,23 +27,13 @@ Example message:
 }
 """
 
-settings = None
-logger = None
 
-
-def main(settings_input, env_input, flag):
-    global settings
-    global env
-    settings = settings_input
-    env = env_input
-
+def main(settings, flag):
     log_file = "queue_workflow_starter.log"
-    global logger
     identity = "queue_workflow_starter_%s" % os.getpid()
     logger = log.logger(log_file, settings.setLevel, identity=identity)
-
     # Simple connect
-    queue = get_queue()
+    queue = get_queue(settings)
 
     while flag.green():
         messages = queue.get_messages(1, visibility_timeout=60,
@@ -51,13 +41,13 @@ def main(settings_input, env_input, flag):
         if messages:
             logger.info(str(len(messages)) + " message received")
             logger.info('message contents: %s', messages[0].get_body())
-            process_message(messages[0])
+            process_message(settings, logger, messages[0])
         else:
             logger.debug("No messages received")
 
     logger.info("graceful shutdown")
 
-def get_queue():
+def get_queue(settings):
     conn = boto.sqs.connect_to_region(settings.sqs_region,
                                       aws_access_key_id=settings.aws_access_key_id,
                                       aws_secret_access_key=settings.aws_secret_access_key)
@@ -65,18 +55,18 @@ def get_queue():
     return queue
 
 
-def process_message(message):
+def process_message(settings, logger, message):
     try:
         message_payload = json.loads(str(unicode_decode(message.get_body())))
         name = message_payload.get('workflow_name')
         data = message_payload.get('workflow_data')
-        start_workflow(name, data)
+        start_workflow(settings, name, data)
     except Exception:
         logger.exception("Exception while processing %s", message.get_body())
     message.delete()
 
 @newrelic.agent.background_task(group='queue_workflow_starter.py')
-def start_workflow(workflow_name, workflow_data):
+def start_workflow(settings, workflow_name, workflow_data):
     data_processor = workflow_data_processors.get(workflow_name)
     workflow_name = 'starter_' + workflow_name
     if data_processor is not None:
@@ -128,14 +118,13 @@ workflow_data_processors = {
 }
 
 
-def load_settings(env):
-    "get settings for the environment"
+def get_settings(env):
     import settings as settings_lib
     return settings_lib.get_settings(env)
 
 
 def console_start():
-    "capture options when running standalone"
+    """capture options when running standalone"""
     parser = OptionParser()
     parser.add_option("-e", "--env", default="dev", action="store", type="string", dest="env",
                       help="set the environment to run, either dev or live")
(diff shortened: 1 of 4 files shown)


queue_workflow_starter.py
```diff
@@ -134,5 +127,25 @@ def process_data_ingestdigest(workflow_name, workflow_data):
...
+                      help="set the environment to run, either dev or live")
+    (options, args) = parser.parse_args()
+    if options.env:
+        env = options.env
+        return env
+
+
 if __name__ == "__main__":
-    process.monitor_interrupt(main)
+    ENV_INPUT = console_start()
```
giorgiosironi: you don't need to make these uppercase as they are local variables now, not module-level constants

queue_workflow_starter.py
```diff
@@ -31,18 +31,11 @@
 logger = None
 
 
-def main(flag):
+def main(settings_input, env_input, flag):
     global settings
     global env
```
giorgiosironi: if `settings` and `env` are being passed correctly where they are used, they shouldn't need to be `global`s anymore

queue_workflow_starter.py
```diff
@@ -134,5 +127,25 @@ def process_data_ingestdigest(workflow_name, workflow_data):
...
+    (options, args) = parser.parse_args()
+    if options.env:
+        env = options.env
+        return env
+
+
 if __name__ == "__main__":
-    process.monitor_interrupt(main)
+    ENV_INPUT = console_start()
+    SETTINGS_INPUT = load_settings(env)
```
giorgiosironi: needs to be passed the result of `console_start` rather than `env` to avoid the global variable

queue_workflow_starter.py
GiancarloFusiello: I realise this wasn't part of your PR but I agree with Giorgio, these do not need to be globals. Would you mind changing them to module level variables/constants?

queue_workflow_starter.py
```diff
@@ -31,18 +31,11 @@
...
-    if options.env:
-        env = options.env
-
-    settings = settings_lib.get_settings(env)
-    env = env
+    settings = settings_input
+    env = env_input
 
     log_file = "queue_workflow_starter.log"
     global logger
```
GiancarloFusiello: Also doesn't need to be a global.

queue_workflow_starter.py
```diff
@@ -134,5 +127,25 @@ def process_data_ingestdigest(workflow_name, workflow_data):
     'IngestDigest': process_data_ingestdigest
 }
 
+
+def load_settings(env):
+    "get settings for the environment"
+    import settings as settings_lib
```
GiancarloFusiello: I think if you resolve the globals above you shouldn't need an inline import.

queue_workflow_starter.py
```diff
@@ -134,5 +127,25 @@ def process_data_ingestdigest(workflow_name, workflow_data):
     'IngestDigest': process_data_ingestdigest
 }
 
+
+def load_settings(env):
+    "get settings for the environment"
```
GiancarloFusiello: doc strings should use three double quotes: https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

comments should use the # symbol

Also, you want to consider the naming of this function. Is it loading settings or getting settings? That way you would not need the doc string/description.

tests/test_queue_workflow_starter.py
```diff
@@ -0,0 +1,139 @@
...
+            }
+        )
+        mock_boto_connection = FakeBotoConnection()
+        fake_boto_conn.return_value = mock_boto_connection
+        mock_queue.return_value = FakeSQSQueue(directory)
+        fake_message = FakeSQSMessage(directory)
+        fake_message.set_body(message_body)
+        mock_queue_read.return_value = [fake_message]
+        queue_workflow_starter.main(settings_mock, 'dev', FakeFlag())
+        self.assertEqual(mock_boto_connection.start_called, True)
```
GiancarloFusiello: I'm not sure if this is really testing anything. I think this puts into words my line of thinking: https://stackoverflow.com/questions/42323543/should-i-unit-test-a-main-method

tests/test_queue_workflow_starter.py
```diff
@@ -0,0 +1,139 @@
...
+                'workflow_name': 'Ping',
+                'workflow_data': {}
+            }
+        )
+        mock_boto_connection = FakeBotoConnection()
+        fake_boto_conn.return_value = mock_boto_connection
+        fake_message = FakeSQSMessage(directory)
+        fake_message.set_body(message_body)
+        queue_workflow_starter.process_message(fake_message)
+        self.assertEqual(mock_boto_connection.start_called, True)
```
GiancarloFusiello: I'm not sure what this is testing. Is it testing that a connection to AWS using boto? or is it testing if a message can be processed? Or given a message a new process/workflow is started?

queue_workflow_starter.py
gnott: Good catch, thanks! I will rework these regarding globals too.

queue_workflow_starter.py
gnott: `proofreader` suggested to me these should be uppercase, so I followed the recommendation.

tests/test_queue_workflow_starter.py
gnott: The starters have no return value and I couldn't test for whether a workflow was started or not, as well as the connection is mocked. In the future, the starter's `start()` method could return something and it will make it clearer and easier to test. I'm just trying to get some tests written for much of the code to have confidence in Python 3 compatibility.

tests/test_queue_workflow_starter.py
gnott: I started out trying to write tests here, later I notice that I didn't need to start at `main()` to get good test coverage, so I changed most of the tests to call `process_message()` instead. I left this in because it was passing and the coverage report included these lines as being tested. The benefit of this test is very minimal though, I agree with you.

queue_workflow_starter.py
gnott: The `settings.py` file doesn't exist in the repo, so `import settings` is not invoked during automated tests in this way. The modules that have all the logic wrapped in a class allow testing without invoking the import, but for these files without a class wrapper this is a way to get around it.

queue_workflow_starter.py
gnott: I didn't know the triple quote was preferred, thanks!

gnott:
Thanks for all the comments, I applied some changes, passing around `settings` and `logger` with no globals, some linting, removing unused variables. I haven't tried running it, of course, but `end2end` tests should be able to tell us if it is wrong. Any additional feedback to reach merging?

queue_workflow_starter.py
GiancarloFusiello: Tests shouldn't dictate the structure of the code. Perhaps you can add a `settings.py` to the repo and have it contain an object/dictionary (what ever the current structure is) and have the keys populated by environment variables. For example:
```
settings = {'key' = os.getenv('MYKEY')}
```
If the environment variable is not supplied then the value of `key` will be `None` or it can be a fallback value if you specify one.

Another way of looking at all of this is, if you have related functionality that depend on the same values, it might be worth encapsulating all relating functions in a class and then simply instantiate the class with the values it requires to work.

Sorry if this is a little in-depth. Feel free to disagree :)

gnott:
Tests are passing, thanks for the feedback. If there are issues unaddressed could be in a new PR? One more test case to build to finish my issue todo list.

PR #850, opened 2019-01-22
Tests for cron.py

gnott:
In reference to issue https://github.com/elifesciences/issues/issues/4062, Python 3 support.

Test cases for the `cron.py` utility that starts workflows based on the clock time. Coverage for each time there are workflows run, and for the starters that are run.

It uncovered another `print` statement with no parentheses when running the tests in Python 3, now fixed.

### DIFF
--- a/cron.py
+++ b/cron.py
@@ -186,7 +186,6 @@ def workflow_conditional_start(settings, starter_name, start_seconds, data=None,
 
     if last_startTimestamp is not None:
         diff_seconds = current_timestamp - start_seconds - last_startTimestamp
-        print(diff_seconds)
 
     if diff_seconds >= 0 or last_startTimestamp is None:
         # Start a new workflow
--- a/cron.py
+++ b/cron.py
@@ -3,8 +3,6 @@ import time
 import importlib
 from optparse import OptionParser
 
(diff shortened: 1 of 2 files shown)


cron.py
```diff
@@ -188,7 +186,7 @@ def workflow_conditional_start(settings, starter_name, start_seconds, data=None,
 
     if last_startTimestamp is not None:
         diff_seconds = current_timestamp - start_seconds - last_startTimestamp
-        print diff_seconds
+        print(diff_seconds)
```
GiancarloFusiello: Why is there a `print` statement here?

giorgiosironi: I do think so, because output from a script executed by a cron goes into a black hole

gnott: It's probably there for old debugging purposes, if `diff_seconds` is negative then the workflow will not be started. It could be logged, but the value is very low, and it could also be removed.

gnott:
Outside the scope of this PR, but there is an internal ticket describing some potential improvements to the cron https://github.com/elifesciences/issues/issues/4221 - an FYI.

cron.py
giorgiosironi: Remove, remove

PR #852, opened 2019-01-25
Create digest medium post

gnott:
In reference to issue https://github.com/elifesciences/issues/issues/4371

`CreateDigestMediumPost` activity was created a while ago on this branch, develop branch has been merged in successively to avoid merge conflicts. This past week is when I built out the logic from the stub.

Functions in the `IngestDigestToEndpoint` activity were useful here too, and these common functions were moved over to `digest_provider.py` and `article_processing.py` so both activities can use them. Refactoring as required.

The `digestparser` in `requirements.txt` is the latest commit, where the temporary directory and image file name can be passed in. Also `medium-sdk-python` is needed to communicate with the Medium API.

An early look for comments. It should be silent if added to a workflow if any of the three required credentials are blank in the `digest.cfg` file.

Reused the email recipient that is already in the settings, but we might want an additional email address to keep the digest `.docx` recipients separate from the `"Medium post was created"` emails.


activity/activity_CreateDigestMediumPost.py
```diff
@@ -0,0 +1,258 @@
...
+                self.statuses["email"] = self.email_notification(article_id)
+
+        except Exception as exception:
+            self.logger.exception("Exception raised in do_activity. Details: %s" % str(exception))
+
+        self.emit_end_message(article_id, version, run)
+
+        return self.ACTIVITY_SUCCESS
+
+    def parse_data(self, data):
```
giorgiosironi: I understand `parse_data` tries to check everything is there, but it's fragile to return a tuple with 7 values that depends on correct ordering to be assigned in the calling code. A more general approach that just enforces the `data` dictionary has the right keys and values, and raises an exception if not?

activity/activity_CreateDigestMediumPost.py
```diff
@@ -0,0 +1,258 @@
...
+            'medium_access_token']
+        missing_credentials = [
+            cred_name for cred_name in required_credentials
+            if not self.digest_config.get(cred_name)]
+        if missing_credentials:
+            self.logger.info(
+                "Missing credentials, %s, in create digest Medium post, article %s" %
+                (missing_credentials, article_id))
+            self.emit_end_message(article_id, version, run)
+            return self.ACTIVITY_SUCCESS
```
giorgiosironi: that's fair, skip on non-configured environments as specified in the description

activity/activity_CreateDigestMediumPost.py
```diff
@@ -0,0 +1,258 @@
...
+            if self.medium_content:
+                self.statuses["post"] = post_medium_content(
+                    self.medium_content, self.digest_config, self.logger)
+
+            if self.statuses.get("post"):
+                # Email
+                self.statuses["email"] = self.email_notification(article_id)
+
+        except Exception as exception:
+            self.logger.exception("Exception raised in do_activity. Details: %s" % str(exception))
```
giorgiosironi: safe rollout :+1:

giorgiosironi: Can probably do more in the message, including article/digest id and version?

activity/activity_CreateDigestMediumPost.py
```diff
@@ -0,0 +1,258 @@
...
+            article_id, version, run, "end",
+            "Finished %s for %s. Statuses: %s" % (self.pretty_name, article_id, self.statuses))
+
+    def emit_error_message(self, article_id, version, run, message):
+        "emit an error message to the queue"
+        return self.emit_message(
+            article_id, version, run, "error", message)
+
+    def approve(self, article_id, status, version, run_type):
+        """should it create a Medium post based on some basic attributes"""
```
giorgiosironi: this smells like it is duplicate with the `digests` service creation activities?

activity/activity_CreateDigestMediumPost.py
```diff
@@ -0,0 +1,258 @@
...
+            approve_status = False
+        else:
+            first_vor_status = digest_provider.approve_by_first_vor(
+                self.settings, self.logger, article_id, version, status)
+            if first_vor_status is False:
+                approve_status = False
+
+        return approve_status
+
+    def email_notification(self, article_id):
```
giorgiosironi: also this sounds like a concern that is shared between activities rather than reimplemented for each one that needs it, `email` provider should house it?

provider/article_processing.py
```diff
@@ -153,5 +153,16 @@ def download_article_xml(settings, to_dir, bucket_folder, bucket_name, version=N
         return filename_plus_path
 
 
-if __name__ == '__main__':
-    main()
+def download_jats(settings, expanded_folder_name, to_dir, logger):
```
giorgiosironi: sounds like the right provider for this code

provider/digest_provider.py
```diff
@@ -87,6 +87,41 @@ def outbox_dest_resource_path(storage_provider, digest, bucket_name):
...
+
+
+def docx_exists_in_s3(settings, article_id, bucket_name, logger):
+    """check if a digest docx exists in the S3 outbox"""
+    resource_origin = docx_resource_origin(
+        settings.storage_provider, article_id, bucket_name)
+    storage = storage_context(settings)
+    try:
+        return storage.resource_exists(resource_origin)
+    except Exception as exception:
```
giorgiosironi: I know these functions were just moved rather than having been created now, but consider re-raising the exception rather than returning a falsy value silently. The activities can handle the exception (with their own `except`) and arguably once an error like this happens it makes no sense to continue executing the rest of the activity, so the interruption caused by the exception should be welcomed.

requirements.txt
```diff
@@ -33,5 +33,6 @@ six==1.10.0
 pyFunctional==1.0.0
 func_timeout==4.3.0
 python-docx==0.8.6
-git+https://github.com/elifesciences/digest-parser.git@edf1ab525dee5eb6196ffc057169e787c11da6d6#egg=digestparser
+git+https://github.com/elifesciences/digest-parser.git@62a5629fd582753f45f794501d3b845cb3680ffd#egg=digestparser
+git+https://github.com/Medium/medium-sdk-python.git@bdf34becf6dd24e3b1fc1048c36416bcba9fe0d6#egg=medium
```
giorgiosironi: `Warning: This code is no longer supported or maintained by Medium.`

It may be supported or maintained by someone else? I see `master` is 3 years newer than the latest tagged release, `0.3.0`.

gnott: I didn't look at that too closely. It still seems to be the `"official"` Python SDK for the Medium API. The commits since the `0.3.0` tag look to be limited to test cases and some readme updates (for example, the note about maintenance). I think this is probably the best Python library to interface with the Medium API for what we intend to do.

We could possibly fork the repo to elifesciences and load it from there in formulas, in case we are concerned about the future maintenance.

activity/activity_CreateDigestMediumPost.py
gnott: I think I moved these out to new function because previously these values are set in the `do_activity()` of an activity. 

Depending on what workflow uses the activity, these values may come from 1. the `session`, or 2. the `data`.

I like how these named variables are short and not complicated for use in emitting messages, writing to log files, and can also be used as function arguments.

I can also see how these could be refactored and cleaned up here. I would be interested to see if you have additional suggestions if you look at how these values are used in all the other activities, a whether we could benefit there too. For example, if instead of using `article_id` I used `data.get['article_id']`, would we ideally then refactor all of the `emit_monitor_event()` calls to use this pattern?

I could also just move these variable assignments back to `do_activity()`, if that is better.

I hope my description is not too confusing. I am open to change to a new method if it looks better.

activity/activity_CreateDigestMediumPost.py
gnott: It will have some similarities for sure. The difference I think is we will still possibly `PUT` an eLife Science Digest as part of a silent correction, but we only want to create a Medium post once only, and never during a silent correction. I think the logic inside `approve()` in each example is different in that way. 

I could possibly refactor with an additional argument `silent_allowed` as `True` or `False` and I might be able to get logic to satisfy both situations. It would cut down on potential code duplication. What do you think?

activity/activity_CreateDigestMediumPost.py
gnott: We started using the `SMTP` emails recently, and I intend on enhancing it and refactoring to use it elsewhere too. I can see how this could be refactored for better integration with activities. 

There are some differences with each usage: the sender, the recipient(s), an attachment (optional), the subject, the body - are all potentially different, even in each place digest activities send emails. I can break apart the composing of the message from the sending of it, and the appropriate values can be passed around.

I have an old issue that I'm also considering to roll in soon, which is to "install" email templates as a Python module, rather than downloading them from an S3 bucket (https://github.com/elifesciences/issues/issues/4172). I could avoid having to write the subject and body each activity needs by creating a pair of templates for each. How it is now is just a quick hack, fortunately these emails are very simple and do not really require templates right now. That is one way I was hoping to improve the email sending. It also touches on my three part plan to remove SimpleDB....

activity/activity_CreateDigestMediumPost.py
giorgiosironi: If the logic is diverging, adding boolean parameters may make it more difficult to understand; I'd hold changes for now

activity/activity_CreateDigestMediumPost.py
giorgiosironi: > I can break apart the composing of the message from the sending of it, and the appropriate values can be passed around.

It looks like you need a `Email` class whose [Value Objects](https://deviq.com/value-object/) you can create with these values, and then send to the `email` provider which has the rest of the logic.

requirements.txt
giorgiosironi: It's the best we can go for, then

activity/activity_CreateDigestMediumPost.py
gnott: There's a standard object that has much of this, `email.mime.multipart.MIMEMultipart`, and I think it satisfies all the requirements so far.

I refactored a little, adding to `email_provider.py`. I separated the message building from the sending. It generates a full list of messages which is hopefully not too memory intensive (this activity only sends a couple emails). Then that list of messages are sent via one SMTP connection.

In more heavy usage of sending a list of messages I wonder whether we will reach an SMTP connection timeout. If we are worried about that we could potentially make a new SMTP connection for every email sent.

I wanted to keep track of some return values, and when sending to a list I added some tracking of the `"success"` and `"error"` count of emails sent, so we do not just stop sending when the first error is encountered.

provider/digest_provider.py
gnott: I cannot say I'm looking in that much detail about the exception handling here. I think it just needs a `raise` in a couple places? I'm hazy on what that does, but maybe you can check (commit is incoming).

activity/activity_CreateDigestMediumPost.py
gnott: 👍

gnott:
Hopefully the tests will be passing again with the latest commit. I've tried to address all the comments, and thanks for the feedback. The new stuff for email sending might be worthy to take a look if you have time, because some of that will be used again when refactoring all the email sending.

activity/activity_CreateDigestMediumPost.py
```diff
@@ -0,0 +1,255 @@
...
+        current_time = time.gmtime()
+        body = success_email_body(current_time)
+        subject = success_email_subject(article_id)
+        sender_email = self.settings.digest_sender_email
+
+        recipient_email_list = email_provider.list_email_recipients(
+            self.settings.digest_recipient_email)
+
+        messages = email_provider.simple_messages(
+            sender_email, recipient_email_list, subject, body, None, self.logger)
```
giorgiosironi: if you say `logger=self.logger` you can then omit the `None`

provider/email_provider.py
```diff
@@ -100,6 +101,48 @@ def smtp_send(connection, sender, recipient, message, logger=None):
...
+def simple_message(sender, recipient, subject, body, attachment=None, logger=None):
+    """set values of a message"""
+    email_message = message(subject, sender, recipient)
+    add_text(email_message, body)
+    if attachment:
+        add_attachment(email_message, attachment)
+    return email_message
+
+
+def simple_messages(sender, recipient_list, subject, body, attachment=None, logger=None):
```
giorgiosironi: maybe it would be helpful to document the various types of the parameters in the docstring, they may be strings, list, files...

provider/email_provider.py
```diff
@@ -100,6 +101,48 @@ def smtp_send(connection, sender, recipient, message, logger=None):
...
+        result = smtp_send_message(connection, email_message, logger)
+        if result:
+            details["success"] += 1
+        else:
+            details["error"] += 1
+    return details
+
+
+def smtp_connect_send_messages(settings, messages, logger=None):
+    """connect then send a list of messages"""
```
giorgiosironi: does it matter to the calling code that it has to connect first? I would expect it. Could be named just `smtp_send_messages()`

activity/activity_CreateDigestMediumPost.py
gnott: Ah cool, I haven't mixed required and optional arguments this way before 👍

provider/email_provider.py
gnott: Giving it a go! 

I'm also changing `attachment` to `attachments` (a list of them) -- even though we only send emails with one attachment so far, supplying a list of them, even if it is a list of one, is probably a good idea. Attachments would possibly need some refactoring if they each needed specific media type definitions. We do not have a use case for that yet.

gnott:
Thanks for the feedback, I applied some changes, and if the tests are still good, I will merge.

There will be a follow-up PR to add it into a workflow, and specifying a new setting for the recipients of these emails.

PR #855, opened 2019-01-31
Create digest medium post enable

gnott:
Two parts to get ready for testing digest Medium posts.

Send to a new list of recipients `digest_medium_recipient_email` in settings so we can send the Medium email to people separately.

Added `CreateDigestMediumPost` to the post publication workflow so it gets invoked. It should be silent if there are no Medium credentials, and not cause failures if there is a runtime error.

Do not merge until builder settings are merged.


gnott:
Setting for email is merged, I'll merge this now. I don't know if there is a lag between settings updates and tests running, if so we can re-run `end2end` later when settings are available.

PR #857, opened 2019-02-05
Video email in html format

gnott:
There were some video emails coming from the workflows today, good! The bad is the HTML content was all in plain text. 

Here is a little tweak to support `text/plain` and `text/html` emails sent from the email provider. I also added in the `charset` option. If the `charset` is `utf-8` then the email body is base64 encoded. I don't know if we always want that, so there are some options to leave the charset as `None` for sending really simple emails.

Hopefully sending in multipart emails using this provider will work for all the current situations. If it can send simple plain messages, HTML emails, and emails with attachments, I think it will cover all the uses we have.

### DIFF
--- a/tests/provider/test_email_provider.py
+++ b/tests/provider/test_email_provider.py
@@ -1,5 +1,6 @@
 import unittest
 from ddt import ddt, data, unpack
+from provider.utils import base64_encode_string
 import provider.email_provider as email_provider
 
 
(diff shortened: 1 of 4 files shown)


tests/provider/test_email_provider.py
```diff
@@ -58,3 +59,32 @@ def test_valid_valid_recipient_object(self, e_mail, expected):
...
+        # note: boundary value is not constant so cannot compare with a fixture easily
+        expected_fragments.append('Content-Type: multipart/mixed; boundary=')
+        expected_fragments.append('Subject: %s' % subject)
+        expected_fragments.append('From: %s' % sender)
+        expected_fragments.append('To: %s' % recipient)
+        expected_fragments.append(expected_content_type)
+        expected_fragments.append('MIME-Version: 1.0')
+        expected_fragments.append('Content-Transfer-Encoding: base64')
+        # body is base64 encoded
+        expected_fragments.append(base64_encode_string(body))
```
giorgiosironi: not perfect for readability (of the source), but I very seldom open an email original source

gnott: It was funny when I viewed the email as a string to see it. I think it only happens when the charset is `utf-8` (and I don't know if it is controllable in the code libraries). I tried to make the test case as readable as possible.

gnott:
Thanks for the comments so far, with the current usage of the email provider as all internal recipients, it should be fairly low risk to failure, and I'm really curious to see the result of the new email headers, so I will merge now.

tests/provider/test_email_provider.py
gnott: As part of bug fixing emails, I have a function in `tests/activity/test_activity_validate_digest_input.py` that will extract the body from a multipart email and decode the body. It's a little rough but it could be expanded to other tests when required.

PR #863, opened 2019-02-08
Feature: uninstalls all source dependencies before installation

lsh-0:


install.sh
```diff
@@ -15,30 +15,8 @@ find starter/ -maxdepth 1 -name '*.pyc' -delete
...
-if pip list | grep jatsgenerator; then
-    pip uninstall -y jatsgenerator
-fi
-if pip list | grep packagepoa; then
-    pip uninstall -y packagepoa
-fi
-if pip list | grep digestparser; then
-    pip uninstall -y digestparser
-fi
+grep "git+" requirements.txt > source-requirements.txt
```
giorgiosironi: also includes `git+https://github.com/Medium/medium-sdk-python.git@bdf34becf6dd24e3b1fc1048c36416bcba9fe0d6#egg=medium`, not a problem

giorgiosironi: maybe `grep "^git+`

install.sh
```diff
@@ -15,30 +15,8 @@ find starter/ -maxdepth 1 -name '*.pyc' -delete
...
+pip uninstall -r source-requirements.txt -y
 pip install -r requirements.txt
-
+# pip install -r source-requirements.txt --no-cache-dir # only if old revisions are still 'sticking'
+rm source-requirements.txt
```
giorgiosironi: can probably be removed just after `pip uninstall`?

gnott:
Nothing further to add, looks well organised and a good idea.

install.sh
lsh-0: actually, that medium repository would suffer the same behaviour unless they're bumping the release version with every revision (which they probably aren't).

I'm thinking that `+` should be escaped actually...

install.sh
lsh-0: the commented line above it may still be necessary. pip will clone the repo and use the version number it finds there. I'm still uncertain whether it uses a cached version of that newly cloned repo with the same version number or installs it from scratch.

install.sh
lsh-0: I'll confirm it's precise behaviour today

gnott:
Any additional thoughts on this PR @lsh-0 before merging, or @giorgiosironi?

lsh-0:
I'll need to go back and look at the code as well as the comments. I'm pretty bogged under today, so it will be monday at the earliest

giorgiosironi:
Nothing to add to the discussion, leaving this for @lsh-0

gnott:
Any additional thoughts about this PR @lsh-0?

lsh-0:
uhm .. you can close it or merge it, I don't mind which. elife-bot has been ticking along merrily without it for a while. It might save on some boilerplate code in the future. I'll leave it up to you

gnott:
It avoids having to spell out which repos are from source, and works in the background producing the same result. All looks good to me - thanks! I'll merge it.

PR #864, opened 2019-02-09
Prune requirements.txt

gnott:
Prune the dependencies in requirements.txt to be direct ones.

Starting with a fresh virtual environment, these minimal requirements allow all of the automated tests to pass. It also installs the versions of `beautifulsoup4` and `lxml` specified in the `elifetools` library.

It has uncovered some missing requirements in other libraries, in the elife ones installed via git, where `GitPython` should be included in the library's requirements, such as in `elife-crossref-xml-generation`. Those could be improved separately. `GitPython` is installed here because it is stated as a requirement in https://github.com/elifesciences/ejp-csv-parser/blob/develop/setup.py

This pruning may be too radical, and we might want to include specific versions of the ones I removed, up for discussion?

The one pip warning I got installing fresh is:

```
pyfunctional 1.0.0 has requirement six==1.10.0, but you'll have six 1.12.0 which is incompatible.
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,12 +1,7 @@
 boto==2.48.0
-GitPython==2.1.11
 Jinja2==2.7.3
-PyPDF2==1.23
 arrow==0.4.4
-beautifulsoup4==4.6.0
 requests==2.20.0
-lxml==4.1.1
-xlrd==0.9.3
 git+https://github.com/elifesciences/elife-tools.git@6c4f822ab8d1bed99c20da057a7907a6ad9a78ce#egg=elifetools
 git+https://github.com/elifesciences/elife-article.git@bb61d4c0e3742403503efe10852114aeb1e07269#egg=elifearticle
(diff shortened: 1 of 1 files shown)


lsh-0:
ok, so what you're removing here are dependencies that were included from other dependencies? that seems fine, the `requirements` file can get messy over time. A project should only require what it directly needs and let the other projects sort themselves out.

once this is done we might look at doing a general upgrade of the dependency versions, make sure they're the latest. @GiancarloFusiello may have some thoughts for long term solutions to that

gnott:
Correct, it would depend more on dependencies twice removed from this project. We could be more specific for the preferred versions of some dependencies here if we are unhappy with the version selected at the time of install, so maybe it will trim down and then get filled back up again if we want to be more specific in the future.

gnott:
Tests are passing, I will go ahead and merge to see if end2end tests are also compatible. We can follow up with another PR to tweak dependencies based on further discussion.

PR #868, opened 2019-02-15
Digest unicode docx file name tests

gnott:
To start out, seeing if I can reproduce the digest docx unicode character file name in an automated test.


gnott:
This is just additional test cases to start with, which are passing, so I will merge now.

PR #870, opened 2019-02-19
Digest unicode fixes

gnott:
I continued troubleshooting the digest unicode character bug again today, continuing from last week, issue https://github.com/elifesciences/issues/issues/4664.

Trying to hunt down exactly why this is happening, and putting together an overall solution for it, I haven't been able to do yet. This is one part of a plan B, a working solution, which is what works when I tested on a bot instance.

The automated tests will not, on any OS we are testing on, exhibit the same error. As such, I think it has something to do with how `worker.py` calls the activity, or in how the `process` provider calls `worker.py`. We could possibly recreate this error with `DIGEST 35774.zip` in an end2end testing situation, because it cannot be reproduced otherwise (so far), an idea for coverage @giorgiosironi ?

Here, the fix is basically: if a digest zip contains a `.docx` file with a unicode character in it, after unzipping the file using the `digestparser` library, the activity `self.temp_dir` ends up being `unicode`, instead of normally a `str`, and this causes an exception when using it in `os.path.join`.

I think this bug is only on Python 2.7, and may not be on Python 3.5, but I'm not complete sure.

If this little fix is acceptable, it will manage to finish an `IngestDigest` workflow, except the digest email attachment file name will not be right. I can do that on a new PR. I'd like to get this part merged to test it again on `continuumtest` to confirm my troubleshooting did address the issue.

### DIFF
--- a/tests/provider/test_utils.py
+++ b/tests/provider/test_utils.py
@@ -1,3 +1,5 @@
+# coding=utf-8
+
 import unittest
 import time
 from ddt import ddt, data, unpack
@@ -88,6 +90,15 @@ class TestUtils(unittest.TestCase):
         )
     def test_unquote_plus(self, value, expected):
         self.assertEqual(utils.unquote_plus(value), expected)
+
+    @unpack
+    @data(
+        (None, None),
+        (u"tmp/foldér", u"tmp/foldér"),
+        (b"tmp/folde\xcc\x81r", u"tmp/foldér")
+        )
+    def test_unicode_encode(self, value, expected):
+        self.assertEqual(utils.unicode_encode(value), expected)
 
 
 if __name__ == '__main__':
--- a/activity/activity_CopyDigestToOutbox.py
+++ b/activity/activity_CopyDigestToOutbox.py
@@ -3,6 +3,7 @@ import json
 import glob
 from S3utility.s3_notification_info import parse_activity_data
 from provider.storage_provider import storage_context
+from provider.utils import unicode_encode
 import provider.digest_provider as digest_provider
 from activity.objects import Activity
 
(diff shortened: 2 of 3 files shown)


GiancarloFusiello:
@gnott The implementation of `str` in python 2 and 3 are differently after recognising the confusion developers faced dealing with unicode characters. This is a pretty good article on this: https://pythonhosted.org/kitchen/unicode-frustrations.html. I hope it helps.

tests/provider/test_utils.py
```diff
@@ -89,6 +91,15 @@ def test_volume_from_pub_date(self, pub_date_str, start_year, expected):
     def test_unquote_plus(self, value, expected):
         self.assertEqual(utils.unquote_plus(value), expected)
 
+    @unpack
+    @data(
+        (None, None),
+        (u"tmp/foldér", u"tmp/foldér"),
+        (b"tmp/folde\xcc\x81r", u"tmp/foldér")
+        )
```
giorgiosironi: the tests seem to do what we expect, I don't have full knowledge of the context here

gnott:
It is the best and simplest fix I have for this one scenario, and it doesn't seem to be an issue anywhere else yet. We can explore it in more detail when it requires it maybe? I will go ahead and merge to see if I can finish the digest bug fix.

gnott:
@GiancarloFusiello the page about unicode is a good one, I will read it more carefully again later. It is right when it says having the right string format matters at times, for example here when using `os.path.join` or `glob.glob`, they will not accept `unicode` in Python 2.7.

gnott:
As a follow-up, I tested this and did more troubleshooting. I didn't write `unicode_encode()` the way it should be. I will be changing it to something better. The differences between Python 2 and 3 messed me up yesterday, now I'm much closer.

PR #871, opened 2019-02-19
Digest unicode bug fix continued

gnott:
Related to issue https://github.com/elifesciences/issues/issues/4664 the digest zip unicode files.

Testing the latest code on `continuumtest` was still failing. First problem is `glob.glob()` internally also has an issue getting file names from a folder, so I am switching over to use `os.listdir()` with specific encoding of file name values.

The other is I messed up the return value for the new `unicode_encode()` function. It should always return a `str` if the input is `unicode` or `bytes`, in both Python 2 and 3.

Tests are passing, I will merge to test it out again on `continuumtest`, if it is failing there for regular files, I'll revert this.

### DIFF
--- a/activity/activity_CopyDigestToOutbox.py
+++ b/activity/activity_CopyDigestToOutbox.py
@@ -1,6 +1,5 @@
 import os
 import json
-import glob
 from S3utility.s3_notification_info import parse_activity_data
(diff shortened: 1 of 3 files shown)


gnott:
This was... successful!

PR #876, opened 2019-02-28
Digest medium check docx exists

gnott:
A small improvement listed on issue https://github.com/elifesciences/issues/issues/4371. 

In the logs there is an ugly exception if an article has no digest, so instead of trying to download the `.docx` file blindly, instead check the `.docx` file exists first. If no `.docx` file exists, log it and return a success for the activity.

The logic checking for if `.docx` files exist is borrowed from its usage in https://github.com/elifesciences/elife-bot/blob/develop/activity/activity_IngestDigestToEndpoint.py.

### DIFF
--- a/activity/activity_CreateDigestMediumPost.py
+++ b/activity/activity_CreateDigestMediumPost.py
@@ -71,6 +71,15 @@ class activity_CreateDigestMediumPost(Activity):
 
         # Wrap in an exception during testing phase
         try:
+            # check if there is a digest docx in the bucket for this article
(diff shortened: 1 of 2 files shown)


gnott:
It relies on the underlying generic call we have available, `return storage.resource_exists(resource_origin)`, at https://github.com/elifesciences/elife-bot/blob/develop/provider/digest_provider.py#L102 . The specific part is compiling the S3 resource location, and also wrapping it in exception catching.

PR #877, opened 2019-02-28
Pass run_type via the lax token

gnott:
As part of issue https://github.com/elifesciences/issues/issues/4777

I started off adding a test for silent correction ingest `workflow_data` in the tests for the `lax_response_adapter`, and linted it as well.

I had a difficult time figuring out where the `run_type` value is dropped as part of silent corrections, so I went to looking at why the `run_type` value of `"initial-article"` was getting lost in normal ingest runs once it reached the post publication workflow.

The `token` value sent back from Lax in an ingest versus a publish action were different. It looks like the token for publish, which is stored in the dashboard, was missing the `run_type` value. I added that here.

I will be continuing to refactor and simplify these in the future. For example, I will probably move the building of the "ready_to_publish" message to a more central place so it can also be used where the silent correction workflow token is made. That is one possibly solution, based on how I think I've figured out why the value is lost.

### DIFF
--- a/tests/test_lax_response_adapter.py
+++ b/tests/test_lax_response_adapter.py
@@ -1,60 +1,67 @@
 import unittest
+import json
 from lax_response_adapter import LaxResponseAdapter
 from mock import Mock
-import json
 from provider.utils import base64_encode_string
 
-fake_token = json.dumps({u'status': u'vor',
-                         u'expanded_folder': u'837411455.1/a8bb05df-2df9-4fce-8f9f-219aca0b0148',
-                         u'version': u'1',
-                         u'force': False,
-                         u'run': u'a8bb05df-2df9-4fce-8f9f-219aca0b0148'})
+FAKE_TOKEN = json.dumps({
+    u'status': u'vor',
+    u'expanded_folder': u'837411455.1/a8bb05df-2df9-4fce-8f9f-219aca0b0148',
+    u'version': u'1',
+    u'force': False,
+    u'run': u'a8bb05df-2df9-4fce-8f9f-219aca0b0148'})
 
-fake_lax_message = json.dumps({"status": "published",
-                               "requested-action": "publish",
-                               "datetime": "2013-03-26T00:00:00+00:00",
-                               "token": base64_encode_string(fake_token),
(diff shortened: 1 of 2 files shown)


tests/test_lax_response_adapter.py
```diff
@@ -1,60 +1,99 @@
...
 from mock import Mock
-import json
 from provider.utils import base64_encode_string
 
-fake_token = json.dumps({u'status': u'vor',
-                         u'expanded_folder': u'837411455.1/a8bb05df-2df9-4fce-8f9f-219aca0b0148',
-                         u'version': u'1',
-                         u'force': False,
-                         u'run': u'a8bb05df-2df9-4fce-8f9f-219aca0b0148'})
+FAKE_TOKEN = json.dumps({
```
giorgiosironi: This fake values act as a specification (well, an example at least) of the real values so they are very useful.

gnott: Agreed!

gnott:
`data` for an activity is the input, and in the most basic form for some activities it uses the `run` value to populate the `session` data. Without `data` in activities that relate to a publishing workflow, it has no idea which article it is considering to be `ReadyToPublish`, so if `data` is `None`, then something is seriously wrong, or the activity is being used in a workflow inappropriately.

There are a few activities that do not require `data` input, such as the `PubmedArticleDeposit`, which gets its input data from an S3 bucket.

gnott:
Confirmed testing on the `continuumtest` environment, the value was passed. In the `queue_workflow_starter.log` file it has `"run_type": "initial-article"`:

```
2019-02-28T19:09:25Z INFO queue_workflow_starter_25510 message contents: {"workflow_name": "PostPerfectPublication", "workflow_data": {"status": "vor", "run_type": "initial-article", "run": "a288251f-b676-421b-9c3d-45e55794290f", "expanded_folder": "39404.1/a288251f-b676-421b-9c3d-45e55794290f", "result": "published", "force": false, "message": null, "article_id": "39404", "update_date": "2019-02-28T00:00:00Z", "version": "1", "requested_action": "publish"}}
```

PR #878, opened 2019-03-01
Add email attachment file name as utf-8 encoded

gnott:
The remaining task on issue https://github.com/elifesciences/issues/issues/4664 to fix bugs with supplying a digest `.docx` file name with unicode characters.

I tested a few variations until after consulting the Python docs, it turns out there is a non-hacky way to add an encoded filename as an email header (my first attempts to set the header from scratch wasn't successful anyway).

The email on mailtrap as a result of this looked good, and much better than before.

I added a test using one of the digest `.docx` files and to check for the email message header string.

It seems between Python 2 and Python 3 the filename is wrapped in `"` characters in Python 2, and not wrapped in Python 3, so the string comparisons are broken into two parts to satisfy both situations.


gnott:
It was all going so well, the tests always pass for me on Mac OS, however the test runners will not open the `Baye\xcc\x81s_35774.docx` file from disk. I give up for today. It worked fine when I tested it on an instance as well. 

Maybe the file name in git has the wrong encoding? Or my test file has the wrong character(s)....

Suggestions are welcome, even better if you manage to fix it to open the fixture file.

gnott:
Forged forward with new ideas today, first is to get the `.docx` file name using the file system, the second is to normalise the file name based on the variant of unicode used on the OS, and always use the same file name encoding for the email attachment when setting headers.

provider/email_provider.py
```diff
@@ -47,18 +48,24 @@ def smtp_connect(settings, logger=None):
     return connection
 
 
+def encode_filename(file_name):
+    """consistently encode a file name in unicode"""
+    return unicode_encode(unicode_decode(
+        unicodedata.ucd_3_2_0.normalize('NFC', unicode_decode(file_name))))
```
giorgiosironi: this function needs a more extensive docstring to explain why are we doing all of this

gnott: I agree, I am just happy to have a test passing combination. I will explain better and add some specific test cases to illustrate it.

gnott: Some individual test scenarios added now, and more description about the new encoding function.

GiancarloFusiello:
This is a good example of why it's a good idea to containerise an application. The fix in this PR is to deal with varying versions of UTF-8 in different operating systems (namely Mac OS X). If you develop, run, test and deploy your application in the same operating system then you mitigate the need for fixes like these. It might be worth asking if other issues will appear due to the need to support multiple operating systems and whether it's worth taking the time to containerise the application?

gnott:
It is a good example of what a consistent container would probably avoid. Maybe we could compile a list on an issue for what a container would solve? In terms of priority I still have some improvements I'd like to do on this project before looking at full container usage (namely the SimpleDB refactoring).

gnott:
Thanks for the comments so far. I want to test this on `continuumtest` environment today to see if I can break it, so I will merge now.

PR #879, opened 2019-03-02
Run type fix part 1

gnott:
As described in https://github.com/elifesciences/issues/issues/4777, this is part 1 of 3 in the fix for passing the `run_type` to downstream workflows when a silent correction is run.

Move the logic of building a workflow starter message, which has workflow data in it, over to a new `token.py` provider. Refactor `ReadyToPublish` to create the message from that.

I'll also create some simple tests for the new provider now too.

### DIFF
--- a/activity/activity_ReadyToPublish.py
+++ b/activity/activity_ReadyToPublish.py
@@ -36,7 +36,7 @@ class activity_ReadyToPublish(Activity):
             update_date = session.get_value('update_date')
             run_type = session.get_value('run_type')
 
-            article_path = self.preview_path(self.settings.article_path_pattern, article_id, version)
+            article_path = preview_path(self.settings.article_path_pattern, article_id, version)
 
             publication_data_message = starter_message(
                 article_id=article_id,
@@ -65,9 +65,6 @@ class activity_ReadyToPublish(Activity):
 
         return self.ACTIVITY_SUCCESS
 
-    def preview_path(self, article_path_pattern, article_id, version):
-        return article_path_pattern.format(id=article_id, version=version)
-
     def prepare_ready_to_publish_message(self, article_id, version, article_path, publication_data_message):
 
         encoded_message = base64_encode_string(json.dumps(publication_data_message))
@@ -80,3 +77,7 @@ class activity_ReadyToPublish(Activity):
                                   encoded_message, "text", version=version)
         self.set_monitor_property(self.settings, article_id, "publication-status",
                                   "ready to publish", "text", version=version)
+
+
+def preview_path(article_path_pattern, article_id, version):
+    return article_path_pattern.format(id=article_id, version=version)
--- a/activity/activity_ReadyToPublish.py
+++ b/activity/activity_ReadyToPublish.py
@@ -4,6 +4,7 @@ from provider.utils import base64_encode_string
 from provider.token import starter_message
 from activity.objects import Activity
 
+
 class activity_ReadyToPublish(Activity):
     def __init__(self, settings, logger, conn=None, token=None, activity_task=None):
         super(activity_ReadyToPublish, self).__init__(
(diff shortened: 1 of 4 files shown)


tests/provider/test_token.py
```diff
@@ -0,0 +1,62 @@
...
+        """for reuse in tests some expected workflow data"""
+        return OrderedDict([
+            ('article_id', self.article_id),
+            ('version', self.version),
+            ('run', self.run),
+            ('expanded_folder', self.expanded_folder),
+            ('status', self.status),
+            ('update_date', self.update_date),
+            ('run_type', self.run_type),
+        ])
```
GiancarloFusiello: I'm not sure I see the purpose of this method. What's wrong with calling the class attributes directly?

tests/provider/test_token.py
```diff
@@ -0,0 +1,62 @@
...
+        workflow_data = OrderedDict([
+            ('article_id', self.article_id),
+            ('version', self.version),
+            ('run', self.run),
+            ('expanded_folder', self.expanded_folder),
+            ('status', self.status),
+            ('update_date', self.update_date),
+            ('run_type', self.run_type),
+        ])
+        self.assertEqual(workflow_data, expected)
```
GiancarloFusiello: If I'm reading this correctly this is not testing anything as it's comparing a hard-coded dictionary to another hard-coded dictionary. Am I missing something?

tests/provider/test_token.py
```diff
@@ -0,0 +1,62 @@
+# coding=utf-8
+
+import unittest
+from collections import OrderedDict
+import provider.token as token
```
GiancarloFusiello: `from provider import token`

activity/activity_ReadyToPublish.py
```diff
@@ -83,4 +77,8 @@ def prepare_ready_to_publish_message(self, article_id, version, run, expanded_fo
         self.set_monitor_property(self.settings, article_id, "_publication-data",
                                   encoded_message, "text", version=version)
         self.set_monitor_property(self.settings, article_id, "publication-status",
-                                  "ready to publish", "text", version=version)
\ No newline at end of file
+                                  "ready to publish", "text", version=version)
+
+
+def preview_path(article_path_pattern, article_id, version):
+    return article_path_pattern.format(id=article_id, version=version)
```
GiancarloFusiello: Is this used by other classes or just this `Activity` subclass?

provider/token.py
```diff
@@ -0,0 +1,19 @@
+from collections import OrderedDict
+
+
+def build_workflow_data(**kwargs):
+    """create a dict of workflow_data from the supplied keyword arguments"""
+    workflow_data = OrderedDict()
+    for data_property in ['article_id', 'version', 'run', 'expanded_folder', 'status',
+                          'update_date', 'run_type']:
+        workflow_data[data_property] = kwargs.get(data_property)
+    return workflow_data
```
GiancarloFusiello: Another way to do this is to specify arguments instead of keyword arguments and then return an Ordered dict with the supplied arguments. Other options are to use a Namedtuple or a class (see Builder pattern) or even `OrderedDict(**kwargs)` which would negate the need for this function.

gnott: I was trying something different hwere with ``**kwargs`` to avoid the PEP recommendation of having too many arguments. I think in this case, and considering your feedback, I will change it back to individual arguments. The flexibility of it handing any value doesn't illustrate what values are allowed and are in standard usage for workflow data.

tests/provider/test_token.py
gnott: My mistake, doing this last thing on a Friday. These tests will look different after I refactor back to function arguments.

tests/provider/test_token.py
gnott: 👍

activity/activity_ReadyToPublish.py
gnott: This is the only place it is used I am aware of in this project, which formats the `settings.article_path_pattern` value (searching for this used elsewhere also indicates this is the only place used).

gnott:
Changed over to a list of named arguments when building the workflow_data.

tests/provider/test_token.py
gnott: This was to avoid writing this structure again in the other test scenario that uses it `test_starter_message()`

gnott:
Also please note when the refactoring of `workflow_data` is finished, it will additionally include `result`, `requested_action`, and `message`, so we will need about 11 or 12 arguments for the token provider functions.

PR #883, opened 2019-03-07
Remove consider Lax two point oh

gnott:
As part of the workflow data clean up in issue https://github.com/elifesciences/issues/issues/4777, the `settings.consider_Lax_elife_2_0` is old and we do not need it anymore.

I started out by adding more tests for `IngestToLax` for better coverage before I refactored.

All the blocks that considered `consider_Lax_elife_2_0` are no longer considering.

### DIFF
--- a/activity/activity_IngestToLax.py
+++ b/activity/activity_IngestToLax.py
@@ -51,7 +51,7 @@ class activity_IngestToLax(Activity):
                                      "aws_secret_access_key": self.settings.aws_secret_access_key}
 
         (message, queue, start_event,
-         end_event, end_event_details, exception) = self.get_message_queue(data, self.settings.consider_Lax_elife_2_0)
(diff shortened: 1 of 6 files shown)


activity/activity_IngestToLax.py
```diff
@@ -65,7 +65,7 @@ def do_activity(self, data=None):
 
 
 
-    def get_message_queue(self, data=None, consider_elife_20=True):
+    def get_message_queue(self, data=None):
         """
         Do the work
```
GiancarloFusiello: I think docstrings like this are not useful. I realise it's not part of this PR but an issue should be created to either describe what a method function does/returns or remove them.

gnott: I'm improving the docstring. It is not normally something I look for when linting, I just run `proofreader` for some recommendations. I didn't write this code but I think I see what it is doing.

PR #884, opened 2019-03-07
Simplify VerifyPublishResponse, no more publication_authority

gnott:
Fixes https://github.com/elifesciences/elife-bot/issues/710

Lots of code deleted since we do not need to look at whether a publish message comes from `elife-website` or `journal`.

I did some linting too, which hopefully doesn't make the changes too hard to read, if you wish to review.

### DIFF
--- a/activity/activity_VerifyLaxResponse.py
+++ b/activity/activity_VerifyLaxResponse.py
@@ -1,5 +1,6 @@
 import json
 from provider.execution_context import get_session
+from provider.lax_provider import message_from_lax
 from uuid import UUID
 from activity.objects import Activity
 
@@ -59,7 +60,7 @@ class activity_VerifyLaxResponse(Activity):
 
                 return self.ACTIVITY_SUCCESS
 
-            message = data['message'] if data['message'] is not None else "(empty message)"
+            message = message_from_lax(data)
             self.emit_monitor_event(self.settings, article_id, version, run, self.pretty_name, "error",
                                     "Lax has not ingested article " + article_id +
                                     " result from lax:" + str(data['result']) + '; message from lax: ' + message)
--- a/activity/activity_VerifyPublishResponse.py
+++ b/activity/activity_VerifyPublishResponse.py
@@ -33,7 +33,9 @@ class activity_VerifyPublishResponse(Activity):
 
     def get_events(self, data):
         """
-        Do the work
+        Given the data for this activity from Lax and the activity starter,
+        return start and end messages to send to the dashboard queue,
+        also return the status return from Lax, and the return value for this workflow activity
         """
         self.logger.info('data: %s' % json.dumps(data, sort_keys=True, indent=4))
 
--- a/activity/activity_VerifyPublishResponse.py
+++ b/activity/activity_VerifyPublishResponse.py
@@ -74,8 +74,7 @@ class activity_VerifyPublishResponse(Activity):
                          self.pretty_name + ": journal", "error",
                          " Lax has not published article " + data['article_id'] +
                          " result from lax:" + str(data['result']) + '; message from lax: ' +
-                         data['message'] if ("message" in data) and
-                         (data['message'] is not None) else "(empty message)"]
+                         message_from_lax(data)]
 
             set_status_property = [
                 self.settings, data['article_id'], "publication-status",
@@ -85,3 +84,8 @@ class activity_VerifyPublishResponse(Activity):
 
         else:
             raise RuntimeError("The publication result isn't a valid one.")
+
+def message_from_lax(data):
+    """format a message from a Lax response"""
+    return data['message'] if (
+        "message" in data and data['message'] is not None) else "(empty message)"
--- a/activity/activity_VerifyPublishResponse.py
+++ b/activity/activity_VerifyPublishResponse.py
@@ -1,4 +1,5 @@
 import json
+from provider.lax_provider import message_from_lax
 from activity.objects import Activity
 
 """
@@ -88,8 +89,3 @@ class activity_VerifyPublishResponse(Activity):
 
         else:
             raise RuntimeError("The publication result isn't a valid one.")
-
-def message_from_lax(data):
-    """format a message from a Lax response"""
-    return data['message'] if (
-        "message" in data and data['message'] is not None) else "(empty message)"
--- a/activity/activity_VerifyPublishResponse.py
+++ b/activity/activity_VerifyPublishResponse.py
@@ -54,52 +54,7 @@ class activity_VerifyPublishResponse(Activity):
 
     def publication_verification_results(self, data, pub_authority, checking_result_from, success):
 
-        if pub_authority == 'elife-website' and checking_result_from == 'elife-website' and success:
-            start_event = [self.settings, data['article_id'], data['version'], data['run'],
-                           self.pretty_name + ": elife-website", "start",
-                           "Starting verification of Publish response " + data['article_id']]
-
-            end_event = [self.settings, data['article_id'], data['version'], data['run'],
-                         self.pretty_name + ": elife-website",
-                         "end", "Finished verification of Publish response " + data['article_id']]
-
-            set_status_event = None
-            success = self.ACTIVITY_SUCCESS
-            return start_event, end_event, set_status_event, success
-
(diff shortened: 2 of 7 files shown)


provider/lax_provider.py
```diff
@@ -233,3 +233,10 @@ def lax_token(run, version, expanded_folder, status, force=False, run_type=None)
     }
     return base64_encode_string(json.dumps(token))
 
+
+def message_from_lax(data):
+    """
+    format a message from a Lax response data
+    """
+    return data['message'] if (
+        "message" in data and data['message'] is not None) else "(empty message)"
```
giorgiosironi: wouldn't an exception be worth it here rather than passing back an unparseable string `(empty message)`?

activity/activity_VerifyLaxResponse.py
```diff
@@ -59,7 +60,7 @@ def do_activity(self, data=None):
 
                 return self.ACTIVITY_SUCCESS
 
-            message = data['message'] if data['message'] is not None else "(empty message)"
+            message = message_from_lax(data)
```
GiancarloFusiello: Here is an alternative to using the `message_from_lax` function:
```
message = data.get('message', '(empty message)')
```
This could also return None as a default value or raise an exception if you need to be notified if an expected message wasn't received.

provider/lax_provider.py
GiancarloFusiello: Please see my comment in `activity/activity_VerifyLaxResponse.py` on line 65.

activity/activity_VerifyPublishResponse.py
```diff
@@ -19,152 +20,72 @@ def __init__(self, settings, logger, conn=None, token=None, activity_task=None):
         self.default_task_schedule_to_close_timeout = 60 * 5
         self.default_task_schedule_to_start_timeout = 30
         self.default_task_start_to_close_timeout = 60 * 5
-        self.description = "Verifies data response from Lax in order to decide if we can carry on with workflow"
+        self.description = ("Verifies data response from Lax in order to decide if we can " +
```
GiancarloFusiello: Is this used anywhere? Normally this is specified in a doc string and then you can call `help(classname)` or `classname.__doc__`

activity/activity_VerifyPublishResponse.py
```diff
@@ -19,152 +20,72 @@ def __init__(self, settings, logger, conn=None, token=None, activity_task=None):
...
         self.emit_monitor_event(*start_msg)
         self.emit_monitor_event(*end_msg)
         if set_status is not None:
             self.set_monitor_property(*set_status, version=data['version'])
         return result
 
-    def get_events(self, data, pub_authority):
+    def get_events(self, data):
         """
         Do the work
```
GiancarloFusiello: I realise this is not part of this PR but this docstring needs updating or removing.

activity/activity_VerifyPublishResponse.py
```diff
@@ -19,152 +20,72 @@ def __init__(self, settings, logger, conn=None, token=None, activity_task=None):
...
         self.default_task_schedule_to_start_timeout = 30
         self.default_task_start_to_close_timeout = 60 * 5
-        self.description = "Verifies data response from Lax in order to decide if we can carry on with workflow"
+        self.description = ("Verifies data response from Lax in order to decide if we can " +
+                            "carry on with workflow")
         self.logger = logger
 
     def do_activity(self, data=None):
-        (start_msg, end_msg, set_status, result) = self.get_events(data, self.publication_authority(self.settings))
+        (start_msg, end_msg, set_status, result) = self.get_events(data)
```
GiancarloFusiello: I don't think you need the brackets here.

activity/activity_VerifyPublishResponse.py
```diff
@@ -19,152 +20,72 @@ def __init__(self, settings, logger, conn=None, token=None, activity_task=None):
...
         if set_status is not None:
             self.set_monitor_property(*set_status, version=data['version'])
         return result
 
-    def get_events(self, data, pub_authority):
+    def get_events(self, data):
         """
         Do the work
         """
         if self.logger:
```
GiancarloFusiello: Sorry again but this if statement is not needed as logger is required to instantiate the class.

activity/activity_VerifyPublishResponse.py
```diff
@@ -19,152 +20,72 @@ def __init__(self, settings, logger, conn=None, token=None, activity_task=None):
...
 
-            end_event = [self.settings, data['article_id'], data['version'], data['run'],
+            end_event = [self.settings, article_id, version, run,
                          self.pretty_name + ": journal", "end",
                          " Finished Verification. Lax has responded with result: published."
-                         " Article: " + data['article_id']]
+                         " Article: " + article_id]
 
-            set_status_property = [self.settings, data['article_id'], "publication-status", "published", "text"]
+            set_status_property = [
```
GiancarloFusiello: it's best practice not to give variables verbs as names. The name should describe what the variable represents.

tests/activity/test_activity_verify_publish_response.py
```diff
@@ -4,157 +4,97 @@
...
 @ddt
 class TestVerifyPublishResponse(unittest.TestCase):
     def setUp(self):
-        self.verifypublishresponse = activity_VerifyPublishResponse(settings_mock, None, None, None, None)
+        self.verifypublishresponse = activity_VerifyPublishResponse(
+            settings_mock, None, None, None, None)
 
-    @data(data_published_lax)
+    @data(DATA_PUBLISHED_LAX)
     def test_get_events_data_published_lax(self, data):
```
GiancarloFusiello: What is this testing?

activity/activity_VerifyPublishResponse.py
gnott: I had to look up where it is used, we do send this to Amazon SWF when the activity is registered  https://github.com/elifesciences/elife-bot/blob/develop/activity/objects.py#L86

tests/activity/test_activity_verify_publish_response.py
gnott: It looks to me like it takes in the data, calls `get_events()`, and then compares the returned values with assertions. I didn't write this activity or the tests, but that is my interpretation of the code.

activity/activity_VerifyPublishResponse.py
gnott: I agree, I saw them but I didn't get that deep into linting, just following what `proofreader` complains about. I can remove these extra parentheses though.

activity/activity_VerifyPublishResponse.py
gnott: 👍 I think since you could pass `None` in as the `logger` in the constructor, this guards against it. But I think that was when the test cases did not mock the logger, and many of the test cases pass in `FakeLogger()` now.

activity/activity_VerifyPublishResponse.py
gnott: I'm reluctant to rename variables in code I didn't write, so maybe we could let it be?

activity/activity_VerifyLaxResponse.py
gnott: Thanks, this is a good alternative.

provider/lax_provider.py
gnott: @giorgiosironi I pulled this out on its own to make the string concatenation in the activity just a little less complicated, and I think it makes up the message that is shown on the dashboard. 

The  `message` value itself is returned by Lax on the queue about Lax's message regarding the "ingest" or "publish" action that was requested, so all we get back is a string `message` that Lax has sent back to us. 

I don't think there is much additional exception handling outside of just catching any exceptions and recovering from them. 

It is also possible (I think) that the `message` value is blank or null, but I guess it is always present in the queue message back from Lax since it doesn't seem to fail due to `message` being absent from the `data` it returns to us.

This is my understanding of the activity while editing it to understand what in particular it does.

activity/activity_VerifyPublishResponse.py
gnott: I will try to expand on this docstring here too. You might want to open an issue regarding these, because you might not be pleased to know when I search for this string, there are about 26 more activities that have this docstring...

gnott:
I made some changes, hopefully for the better. end2end tests I hope will pick up on anything missed.

PR #887, opened 2019-03-13
Feature: adds python 3.6 to tox environments to test

lsh-0:
Ubuntu 16.04 LTS uses Python 3.5 (supported)
Ubuntu 18.04 LTS uses Python 3.6

cc @giorgiosironi @gnott

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,7 @@ git+https://github.com/elifesciences/ejp-csv-parser.git@9fdfa6a37c00b6134f8a2a46
 git+https://github.com/elifesciences/jats-generator.git@ea0d4d90a28198eb2953cc982da85392fbfee360#egg=jatsgenerator
 git+https://github.com/elifesciences/package-poa.git@2649777a020661e507de4087d447f50ca8c2fe57#egg=packagepoa
 PyYAML==4.2b2
-Wand==0.4.2
+Wand==0.5.1
 paramiko==2.4.2
 mock==1.3.0
 redis==2.10.5
@@ -23,7 +23,7 @@ pyOpenSSL==18.0.0
 newrelic==2.72.0.52
 pylint==1.6.4
 pyGithub==1.27.1
-pyFunctional==1.0.0
+pyFunctional==1.2.0
 func_timeout==4.3.0
 python-docx==0.8.6
 git+https://github.com/elifesciences/digest-parser.git@9a33eafcf12b0be4664a4225be9f4925dd75daec#egg=digestparser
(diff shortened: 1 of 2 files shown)


lsh-0:
I can see the elife-bot-formula has 16.04 support already, but I'm doing a quick pass on all 16.04 projects to see what the difficulty is in upgrading that to 18.04. It may save us some time and hassle in the medium term future

lsh-0:
Please merge if you're happy with the changes

giorgiosironi:
Only the virtualenv provided by `elife-bot-formula` is [tested by Jenkins](https://github.com/elifesciences/elife-bot/blob/develop/project_tests.sh). We could use `tox` in there but this would require having all these Python versions available on `elife-bot--ci`?

gnott:
Running the automated tests with `tox` was considered around the time this project was made to successfully run tests in Python 3, but I've been testing compatibility manually (actually, I'm always running tests in Python 3 as I develop, and the automated tests in Python 2 catches things that would break in prod). I'm ok with switching to `tox`, but I'll defer to @giorgiosironi for whether that complicates the testing environment.

Back to this PR edits - I'm running the tests in Python 3.6 locally, and it is uncovering some tests with delayed running. 

The first one I've diagnosed is the tests for the `PublicationEmail` activity, where it was not mocking a function that was trying to connect to S3. I guess in other tests it is ignoring the result or timing out more quickly. 

When running the tests in Python 3.6, or the new `tox` environment it created, it must be waiting for some S3 timeout for each test scenario, causing significant delays, estimated at 60 seconds per scenario and there are many in that test file.

I managed to track down this first one, and I can commit the change somewhere. I see another that is probably also causing a delay.

I might create a new branch and PR for the test case improvements, we can merge it, and then pull from the `develop` branch into here? It might be cleaner than me adding test case fixes to this branch, which would be off topic.

gnott:
I improved the test cases so they are better when I run `tox` locally. As I described above, the project tests are only running in `py27` right now and not using `tox`.

Whether this project can run `tox` in the automated tests maybe we can discuss separately @giorgiosironi? It might be a good idea, but it is probably not critical until we try running this project with Python 3 on real instances.

lsh-0:
a more gradual approach for elife-bot then, 16.04 then, eventually, 18.04.

gnott:
Looks like everyone is happy with this PR, I will merge it now so it is not forgotten.

giorgiosironi:
> Whether this project can run tox in the automated tests maybe we can discuss separately @giorgiosironi?

Yes, it's just a passing comment - I'd prefer containerization for example.

Issue #854, opened 2019-01-29
Replace elife_add_email_to_email_queue() with sending emails direct by SMTP


gnott:
Going ahead with the plan to refactor away SimpleDB usage in ticket https://github.com/elifesciences/elife-bot/issues/565.

Instead of adding emails to the SimpleDB email queue and sending the email later, all calls to `elife_add_email_to_email_queue()` in the `simpleDB.py` provider will be replaced with an almost-drop-in-replacement function in the `email_provider.py` provider.

- [x] sender, recipient, subject, body, as a minimum
- [x] optionally pass in an existing SMTP connection, in case reconnecting by SMTP for each email sent causes additional latency
- [x] Refactor email sending logic in `AdminEmailHistory` activity
- [x] Refactor email sending logic in `PackagePOA` activity
- [x] Refactor email sending logic in `PublishFinalPOA` activity
- [x] Refactor email sending logic in `PubmedArticleDeposit` activity
- [x] Refactor email sending logic in `PubRouterDeposit` activity
- [x] Remove `elife_add_email_to_email_queue()` and `elife_get_unique_email_queue_item_name()` from `SimpleDB` provider
- [x] Remove additional email queue logic from `SimpleDB` provider
- [x] Remove the logic from `cron_FiveMinute` that starts the `SendQueuedEmail` workflow
- [x] Remove the `SendQueuedEmail` workflow 
- [x] Remove the `SendQueuedEmail` activity
- [x] Remove the `starter_SendQueuedEmail` starter
- [x] ~~De-schedule `SendQueuedEmail` from the `cron.py` list of scheduled workflows~~ turns out this was in `cron_FiveMinute`

After merged and deployed

- [x] Can delete the email related tables (domains) of data stored in SimpleDB
- [x] Can delete email related data in S3 buckets (some email body are stored there due to size exceeded SimpleDB field lengths) for eLife stored in `s3://elife-bot/email/`

PR #889, opened 2019-03-14
Digest jats silent logic and email

gnott:
For some remaining tasks in issue https://github.com/elifesciences/issues/issues/4538, eventually instead of sending a digest `.docx` as an email attachment we will only be posting digest JATS to the endpoint.

First logical part is to not post to the endpoint if it is for a silent digest ingestion.

Second part is we want to send an email notification when a successful digest JATS post is completed. This goes to the same recipients as were receiving the email attachment. I added the JATS content to the body of the email which might be useful for debugging purposes.

### DIFF
--- a/activity/activity_PostDigestJATS.py
+++ b/activity/activity_PostDigestJATS.py
@@ -214,8 +214,7 @@ def success_email_body(current_time, digest_content, jats_content):
     """
     Format the body of the email
     """
-    body = "JATS content for article %s:\n\n%s\n\n" % (
-        str(digest_content.doi), str(jats_content))
+    body = "JATS content for article %s:\n\n%s\n\n" % (digest_content.doi, jats_content)
     date_format = '%Y-%m-%dT%H:%M:%S.000Z'
     datetime_string = time.strftime(date_format, current_time)
     body += "As at " + datetime_string + "\n"
--- a/activity/activity_PostDigestJATS.py
+++ b/activity/activity_PostDigestJATS.py
@@ -62,6 +62,11 @@ class activity_PostDigestJATS(Activity):
         # parse the data with the digest_provider
         real_filename, bucket_name, bucket_folder = parse_activity_data(data)
 
+        # check if it is a silent run
+        if digest_provider.silent_digest(real_filename):
+            self.logger.info('PostDigestJATS silent deposit of real_filename: %s', real_filename)
+            return self.ACTIVITY_SUCCESS
+
         # Download from S3
         self.input_file = digest_provider.download_digest_from_s3(
             self.settings, real_filename, bucket_name, bucket_folder, self.input_dir)
--- a/activity/activity_PostDigestJATS.py
+++ b/activity/activity_PostDigestJATS.py
@@ -101,7 +101,7 @@ class activity_PostDigestJATS(Activity):
         return self.ACTIVITY_PERMANENT_FAILURE
 
     def post_jats(self, digest, jats_content):
-        "prepare and POST jats to API endpoint"
+        """prepare and POST jats to API endpoint"""
         url = self.settings.typesetter_digest_endpoint
         payload = post_payload(digest, jats_content, self.settings.typesetter_digest_api_key)
         if payload:
@@ -146,7 +146,7 @@ class activity_PostDigestJATS(Activity):
 
 
 def post_payload(digest, jats_content, api_key):
-    "compile the POST data payload"
+    """compile the POST data payload"""
     if not digest:
         return None
     account_key = 1
@@ -161,7 +161,7 @@ def post_payload(digest, jats_content, api_key):
 
 
 def post_jats_to_endpoint(url, payload, logger):
-    "issue the POST"
+    """issue the POST"""
     resp = get_as_params(url, payload)
     # Check for good HTTP status code
     if resp.status_code != 200:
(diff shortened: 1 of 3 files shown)


activity/activity_PostDigestJATS.py
```diff
@@ -79,6 +86,9 @@ def do_activity(self, data=None):
         # POST to API endpoint
         try:
             self.statuses["post"] = self.post_jats(self.digest, self.jats_content)
+            # send email
+            if self.statuses.get("post"):
+                self.statuses["email"] = self.send_email(self.digest, self.jats_content)
```
giorgiosironi: shouldn't we sending an email also in case of failure (saying "this has failed")?

gnott: I agree! (I thought I said that somewhere already, maybe on a different issue). Added some error emails now.

gnott:
I adding sending one error email at first to get sending error emails started. I ran into a bug with `side_effect` tests raising an Exception (which I figured out today, it was due to settings manipulations).

This activity is part of the workflow started when a digest `.zip` file is copied into an S3 bucket, so it has some instances where it would return `ACTIVITY_PERMANENT_FAILURE` in some places when it encountered an error. 

After reviewing some of the steps of downloading input, building the digest, generating JATS, and posting to the endpoint, instead if it runs into an error, it will send an error email and then return `ACTIVITY_SUCCESS`. Failing the workflow doesn't have much benefit right now anyway, since this is the last activity to run in the workflow.

I hope these changes are favourable to reach merging this PR. It is the next step in switching over to complete JATS to endpoint deliveries of digests, so we can disable sending the digest email attachments (when all recipients of the data are happy).

giorgiosironi:
I initially missed this, but when you have added changes it may be helpful to hit the `Re-request review` button in the `Reviewers` section so that it's clear a new look is needed. In my case, it gets through email filters into the right folder which I look at more often.

activity/activity_PostDigestJATS.py
```diff
@@ -62,42 +64,114 @@ def do_activity(self, data=None):
         # parse the data with the digest_provider
         real_filename, bucket_name, bucket_folder = parse_activity_data(data)
 
+        # check if it is a silent run
+        if digest_provider.silent_digest(real_filename):
+            self.logger.info('PostDigestJATS silent deposit of real_filename: %s', real_filename)
```
giorgiosironi: Is it truly a "silent deposit"? "ignoring deposit due to silent run"?

activity/activity_PostDigestJATS.py
```diff
@@ -62,42 +64,114 @@ def do_activity(self, data=None):
...
         self.input_file = digest_provider.download_digest_from_s3(
             self.settings, real_filename, bucket_name, bucket_folder, self.input_dir)
 
         # Parse input and build digest
         self.statuses["build"], self.digest = digest_provider.build_digest(
             self.input_file, self.temp_dir, self.logger, self.digest_config)
+        if not self.statuses.get("build"):
+            error_message = "Failed to build digest from file %s" % self.input_file
+            self.logger.info(error_message)
+            self.statuses["error_email"] = self.email_error_report(
```
giorgiosironi: What's the point of populating `self.statuses` if we are going to return immediately after without using it? Ok if it's for future usage.

activity/activity_PostDigestJATS.py
```diff
@@ -62,42 +64,114 @@ def do_activity(self, data=None):
...
+
+        current_time = time.gmtime()
+        body = success_email_body(current_time, digest_content, jats_content)
+        subject = success_email_subject(digest_content)
+        sender_email = self.settings.digest_sender_email
+
+        recipient_email_list = email_provider.list_email_recipients(
+            self.settings.digest_recipient_email)
+
+        connection = email_provider.smtp_connect(self.settings, self.logger)
```
giorgiosironi: The `email_providers` is chatty and procedural (get a `connection`, then passing it back in separate functions like `smtp_send`); this pushes towards a class that can model that:
```
email_provider.new_email(self.settings, self.logger).add_text(...).send(recipient)
```
(an object being returned by `new_email`)
Wouldn't do this in this PR however.

activity/activity_PostDigestJATS.py
gnott: The file deposited would have `-SILENT` in its name, being the silent deposit, but I am changing the wording so this `"PostDigestJATS is not posting JATS because it is a silent workflow, real_filename: ...".

activity/activity_PostDigestJATS.py
gnott: `self.statuses` is used for some test cases to determine whether it reached a certain step in the `do_activity()` function, at least it is useful for that. We also log the values, but only if it gets all the way to the end. It's possible they are not needed here but hopefully they do no harm for this revision?

activity/activity_PostDigestJATS.py
gnott: I can see that as an option for an alternate method to send emails. The current `email_provider.py` is fairly new, and it was to consolidate some email sending logic. I tried to keep the functions small so they could be composed in different ways, and make them easy to test. It uses the already available Python email objects, but there is still room to wrap those into a super object.

You can also add detail to https://github.com/elifesciences/elife-bot/issues/854 where it might fit. I also planned to refactor all the email sending to not use SimpleDB, but those plans have not come back up to the top priority yet.

Regarding the usage of email sending in this activity, I can see using a replacement to reduce the number of lines of code, and maybe it will be a better and cleaner result? I will give it a go and see if the existing tests pass after changing it over.

gnott:
Sorry, I neglected to click the `re-request review` button and I will do that consistently going forward. In my defense, I think that button has only been available for a short while and I've only clicked it a couple times before.

PR #897, opened 2019-04-03
Precise email recipients for digest activities.

gnott:
To reach completion of issue https://github.com/elifesciences/issues/issues/4538, recent comments show some people would like to receive certain emails regarding digests and to not receive other emails.

There are four digest related activities that send emails: 

1. when the input zip is validated
2. sending a generated docx file as an email attachment
3. posting digest JATS to an endpoint
4. creating a Medium post for a digest

Some of these send only success emails, and some send success or error emails.

Here I opted to separate each of the currently used email recipient lists into separate variables in the settings. We can then adjust the recipients of each as people prefer.

### DIFF
--- a/activity/activity_EmailDigest.py
+++ b/activity/activity_EmailDigest.py
@@ -123,7 +123,7 @@ class activity_EmailDigest(Activity):
         sender_email = self.settings.digest_sender_email
 
         recipient_email_list = email_provider.list_email_recipients(
-            self.settings.digest_recipient_email)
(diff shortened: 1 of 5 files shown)


gnott:
I should have created this PR as a draft / WIP (if that option was available, I didn't see), because first live `settings.py` changes must be merged before this code can use those new settings.

gnott:
Settings required are merged, I think it is safe to merge this and we'll see in test results.

PR #900, opened 2019-04-05
PMC no more sending emails

gnott:
In reference to issue https://github.com/elifesciences/issues/issues/4857.

This is the simplest edit to remove all the emailing logic from the `PMCDeposit` activity. I did only a tiny bit of linting.

There is room in this file to do some more extensive refactoring: move the no-self-use functions out of the class, use the `ftp.py` provider instead of the duplicate functions in here, use the `storage_context` instead of connecting to S3 directly. I am open to suggestions whether those refactors should be in a separate PR (and at a later date), or whether it is good to just do it all now along with removing the emailing logic.

### DIFF
--- a/activity/activity_PMCDeposit.py
+++ b/activity/activity_PMCDeposit.py
@@ -1,9 +1,6 @@
 import os
-import boto.swf
 import json
 from collections import OrderedDict
-import datetime
-import time
 import zipfile
 import shutil
 import re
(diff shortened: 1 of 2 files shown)


gnott:
I like removing code too, I'm not sure why. Thanks @giorgiosironi, I will give this a merge and try it out.

PR #903, opened 2019-04-11
Fix log arguments, test case for ArchiveArticle.

gnott:
Fixes https://github.com/elifesciences/issues/issues/4827.

Arguments need to be wrapped in parentheses. I added a test case that covers where the error was too.

### DIFF
--- a/activity/activity_ArchiveArticle.py
+++ b/activity/activity_ArchiveArticle.py
@@ -100,9 +100,8 @@ class activity_ArchiveArticle(Activity):
                 with open(file_path, 'wb') as open_file:
                     self.logger.info("Downloading %s to %s", (storage_resource_origin, file_path))
                     storage.get_resource_to_file(storage_resource_origin, open_file)
-        except IOError as exception:
-            self.logger.exception(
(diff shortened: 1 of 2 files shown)


activity/activity_ArchiveArticle.py
```diff
@@ -102,7 +102,7 @@ def download_files(self, bucket_name, expanded_folder, zip_dir_path):
                     storage.get_resource_to_file(storage_resource_origin, open_file)
         except IOError as exception:
             self.logger.exception(
-                "Failed to download file %s. details: %s" % key_name, str(exception))
+                "Failed to download file %s. details: %s" % (key_name, str(exception)))
```
giorgiosironi: I thought this was going to be `"Failed to download file %s. details: %s", key_name, str(exception))`; also with `logger.exception` you shouldn't need to pass in anything about the exception because it would capture it automatically.

gnott: I rarely remember which formatting uses `%` and which uses `,`, I was more concerned with how logging an exception caused an exception, of course! I'll change it so the exception is not included in the log message, since it is already included, as you say.

PR #906, opened 2019-04-23
Change to digest JATS error message returned.

gnott:
A continuation of PR https://github.com/elifesciences/elife-bot/pull/905.

The `payload` of a digest JATS POST operation may contain some sensitive credentials. The change here is to only return the `status_code` and the response content for potential inclusion in an error email. The full `payload` is still written to the log file if it is useful in future debugging.

### DIFF
--- a/tests/activity/test_activity_post_digest_jats.py
+++ b/tests/activity/test_activity_post_digest_jats.py
@@ -395,6 +395,8 @@ eLife bot'''
 
         expected = u'''Exception blah blah blah
 
+More details about the error may be found in the log file
+
 Article DOI: 10.7554/eLife.99999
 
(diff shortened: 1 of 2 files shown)


tests/activity/test_activity_post_digest_jats.py
```diff
@@ -395,6 +395,8 @@ def test_error_email_body(self):
 
         expected = u'''Exception blah blah blah
 
+More details about the error may be found in the log file
```
giorgiosironi: `worker.log`?

gnott: Which log? the `worker.log` 👍

PR #910, opened 2019-04-30
PMC resupply

gnott:
Logic as outlined in issue https://github.com/elifesciences/issues/issues/4889 up to where we could potentially deploy it for a test with a couple zip files. The code shouldn't impact any existing PMC deposits.

There's some potential for some refactoring in the files edited here if you feel like it is a good time to do it before merging. That could include, for example, to change `PMCDeposit` activity to use the FTP provider instead of the separate function it is using right now.

### DIFF
--- a/activity/activity_CreateDigestMediumPost.py
+++ b/activity/activity_CreateDigestMediumPost.py
@@ -193,7 +193,7 @@ class activity_CreateDigestMediumPost(Activity):
             approve_status = False
 
         # check silent correction
-        if run_type == "silent-correction":
+        if run_type in ["silent-correction", "silent-correction-pmc-resupply"]:
             approve_status = False
         else:
             first_vor_status = digest_provider.approve_by_first_vor(
--- a/activity/activity_PubRouterDeposit.py
+++ b/activity/activity_PubRouterDeposit.py
@@ -105,9 +105,9 @@ class activity_PubRouterDeposit(Activity):
 
         for article in self.articles_approved:
             # Start a workflow for each article this is approved to publish
-            if self.workflow in ["PMC", "PMC_Resupply"]:
+            if self.workflow in ["PMC", "PMC-Resupply"]:
                 folder = ""
-                if self.workflow == "PMC_Resupply":
+                if self.workflow == "PMC-Resupply":
                     folder = "resupplies"
                 zip_file_name = self.archive_zip_file_name(article)
                 starter_status = self.start_pmc_deposit_workflow(article, zip_file_name, folder)
@@ -160,7 +160,7 @@ class activity_PubRouterDeposit(Activity):
             return "scopus/outbox/"
         elif workflow == "PMC":
             return "pmc/outbox/"
-        elif workflow == "PMC_Resupply":
+        elif workflow == "PMC-Resupply":
             return "pmc_resupply/outbox/"
         elif workflow == "CNPIEC":
             return "cnpiec/outbox/"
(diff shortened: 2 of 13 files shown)


activity/activity_CreateDigestMediumPost.py
```diff
@@ -193,7 +193,7 @@ def approve(self, article_id, status, version, run_type):
             approve_status = False
 
         # check silent correction
-        if run_type == "silent-correction":
+        if run_type in ["silent-correction", "silent-correction-pmc-resupply"]:
```
giorgiosironi: tempted to make `run_type` an object, but there is a whole lot more activities than run types so this decomposition seems easier to extend. It does mean going through every activity to check if a `run_type` is added, but this is not frequent.

activity/activity_PubRouterDeposit.py
```diff
@@ -105,9 +105,12 @@ def do_activity(self, data=None):
 
         for article in self.articles_approved:
             # Start a workflow for each article this is approved to publish
-            if self.workflow == 'PMC':
+            if self.workflow in ["PMC", "PMC-Resupply"]:
+                folder = ""
+                if self.workflow == "PMC-Resupply":
+                    folder = "resupplies"
```
giorgiosironi: prefer `else` for `folder = ""` rather than overwriting it, keeps a single assignment

activity/activity_PubRouterDeposit.py
```diff
@@ -180,6 +185,8 @@ def get_published_folder(self, workflow):
             return "scopus/published/"
         elif workflow == "PMC":
             return "pmc/published/"
+        elif workflow == "PMC-Resupply":
```
giorgiosironi: the (partial) names of workflows are magic strings that are used in various activities - extracting them as constants somewhere is a low hanging fruit that prevents typos.

activity/activity_CreateDigestMediumPost.py
gnott: Technically we have only two `run_type` values (now three, I guess), and they're only part of article publishing workflows. I found the existing usage of `"silent-correction"` just by searching files, and it should be similar to remove the temporary value later after we've finished the PMC resupply.

I like your idea, maybe it is not the time yet to formalise this session value into something more structured.

activity/activity_PubRouterDeposit.py
gnott: Sure, that's good too, and a good reason 👍

gnott: It also turns out coverage reports are often correct, and the line was not covered. After some attempts, I found a bug where `PMC-Resupply` needed to be added to additional places.

activity/activity_PubRouterDeposit.py
gnott: I can see options for cleaning these up, however I'd like to suggest they are out of scope for this PR. I can open a new issue for a simple low-hanging-fruit clean up of workflow values. I have an issue for a more comprehensive refactor of the pub routers, and the organising of names and outbox values is probably a good start. 

They span many activities and starters and to unify them I feel would delay testing the PMC resupply feature.

gnott:
I hope the updates and comments are welcome here, ready for another review if you find anything must be done before merging. I created a new issue https://github.com/elifesciences/elife-bot/issues/911 specifically around naming and use of constants, as well I think the relationship between the outbox names and the workflow types can be cleaned up there.

gnott:
Thanks, I will merge this and test on the `continuumtest` environment to see if it needs more changes to run as planned.

Issue #916, opened 2019-06-10
Support for new EJP CSV file names


gnott:
Regarding internal issues https://github.com/elifesciences/issues/issues/4917 (bug profiled), and https://github.com/elifesciences/issues/issues/4939 (proposed new CSV file names), it will require some editing of the `elife-bot` project.

The basic plan would be to change the `./provider/ejp.py` to detect both the old and new file names when it looks for the latest CSV files in the bucket. These are used for PoA article XML generation.

Plan after reviewing code and file names

- [x] Change the CSV file match patterns to be more flexible, matching both old and new file naming
- [x] Merge and deploy the code
- [x] Wait for EJP to update the file naming completely
- [x] Update the CSV file match patterns to be more precise for the new file naming, to avoid potential duplicate matches if someone creates a similar named query in the future
- [x] Optional: may look at improving the `ejp.py` provider to use the `storage_context` instead of connecting directly to S3 (refactoring) - edit: I'm going to skip this for now, moving all to storage_context can be done at a later date in another initiative

gnott:
PR for the first checkbox in https://github.com/elifesciences/elife-bot/pull/917

gnott:
Code to support both types of file naming is deployed. The file names changed today, after the PoA workflows ran. When the next PoA packaging is done, tomorrow possibly, we can see the result and if there are any bugs.

gnott:
Switch to the new file names for good is done, closing!

PR #917, opened 2019-06-11
CSV file matches, tests for old and new names.

gnott:
Very basic way to satisfy item one on issue https://github.com/elifesciences/elife-bot/issues/916.

The file name fragments used in matching CSV files names is shortened, which allows matching both the old and new file naming formats.

I expanded the tests to show how it works on both the old CSV file names, a mixture of old and new names together, or just the new file names.

As mentioned on the issue, once the file names are switched over, then we can expand the file name matches to be longer, for example `"ejp_query_tool_query_id_POA_Manuscript"` instead of just `"_POA_Manuscript"` just in case a query with a similar name is created. Since it is using `re.search`, the pattern could also incorporate matching the `_2019_06_10_eLife.csv` portion of file names if the date numbering is always consistent, which it has been so far.

I have omitted additional refactoring of the S3 bucket operations at this time, so this can be reviewed and possibly deployed soon, in case the EJP file name change is imminent.

### DIFF
--- a/provider/ejp.py
+++ b/provider/ejp.py
@@ -290,21 +290,21 @@ class EJP(object):
         # For each file_type, specify a unique file name fragment to filter on
         #  with regular expression search
         fn_fragment = {}
-        fn_fragment["author"] = "ejp_query_tool_query_id_152_15a"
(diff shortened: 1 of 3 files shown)


gnott:
Will merge to support pending file name changes coming soon.

Issue #912, opened 2019-05-09
CopyGlencoeStillImages require Glencoe metadata for XML with videos


gnott:
A small improvement proposed to overcome video still image errors, related to https://github.com/elifesciences/issues/issues/4907.

In article publishing workflows we verify Glencoe metadata exists in the `VerifyGlencoe` activity, however due to coordination of resupplying videos as part of a new ingest run, the video metadata can disappear for a short time while Glencoe processes video files. If the metadata is missing by the time the `CopyGlencoeStillImages` activity happens, then it does not find any video still images, the images do not end up in the CDN bucket, ultimately leading to a validation error in Lax when checking the article JSON.

The small improvement is to add into the `CopyGlencoeStillImages` activity a verify and retry logic, similar to the code in https://github.com/elifesciences/elife-bot/blob/develop/activity/activity_VerifyGlencoe.py#L74-L79, the key logic being

```
            time.sleep(60)
            return self.ACTIVITY_TEMPORARY_FAILURE
```

The sleep time will allow Glencoe to transcode the new video deposit. The `self.ACTIVITY_TEMPORARY_FAILURE` will result in repeating the `CopyGlencoeStillImages` activity 60 seconds later, when hopefully it will work.

As part of this, `CopyGlencoeStillImages` should also be made to read the article XML, and if there are videos in the XML, then it will require a non-404 response from the Glencoe metadata request, otherwise it will sleep and then retry later.

@giorgiosironi I wonder if you could please read this proposal and see if the logic is a good idea as written?

It would render `VerifyGlencoe` almost useless since `CopyGlencoeStillImages` would practically fill that role after the change.

giorgiosironi:
I agree this would be more robust in retrying. Not sure I fully understand the retry conditions but replicating the behavior from `VerifyGlencoe` is a good start.

gnott:
I wasn't sure about when the retry happens either, presumably it is when an exception is raised, possibly a `404` response from the Glencoe resource. I might also clarify the logic for retry if it can be clearer in both activities.

gnott:
In the end, the activity will now return `self.ACTIVITY_TEMPORARY_FAILURE` if the XML has videos, but Glencoe says it does not have videos. This will result in a default 10 second sleep of a bot decider before the activity is tried again. This will avoid the immediate `self.ACTIVITY_PERMANENT_FAILURE` which causes the workflow to stop when this same situation was encountered in the past.

[...]

PR #920, opened 2019-06-18
Copy glencoe retry

gnott:
Fixes https://github.com/elifesciences/elife-bot/issues/912

The commit from June 12th is where I think I figured out where to `time.sleep(60)` when trying to copy Glencoe still images. It relies on also getting a copy of the article XML (as made available from session data), and if there are videos in the XML, but Glencoe request is a `404`, then we should wait for Glencoe to finish processing videos, because they're not ready yet.

I was a little confused by all the options of poa / non-poa, Glencoe assertions or exceptions. The tests I added had to mock many of the modules, as the other tests for `do_activity()` often do.

Today, I refactored the activity code as best I could see doing it for some improvements. 
- The `standalone`, `poa` and `vor` logic is slightly more separate. 
- There is only one situation (when processing `vor`) that produces a `start_event` and that is moved.
- I moved the `end_event` building to their own methods for (hopefully) better readability
- Getting Glencoe `metadata` was separated from processing the metadata, to more clearly define when we look for requests status code assertions.

There's room for more simplification of the test cases, by testing functions other than `do_activity()` in some cases, but it might loose some features such as testing the actual messages emitted to the queue.

@giorgiosironi I wonder if there is feedback you wish to provide on this PR please?

I'm fairly confident the actual functionality and integrity of the activity remains the same, and I really only added a new situation that returns `self.ACTIVITY_TEMPORARY_FAILURE` after the `time.sleep(60)` to avoid the failure of missing still images reported on the dashboard.

### DIFF
--- a/activity/activity_CopyGlencoeStillImages.py
+++ b/activity/activity_CopyGlencoeStillImages.py
@@ -1,5 +1,4 @@
 import json
-import time
 import os
 import requests
 from activity.objects import Activity
@@ -147,9 +146,8 @@ class activity_CopyGlencoeStillImages(Activity):
         first_chars_error = str(exception)[:21]
         if first_chars_error == "article has no videos":
             if has_videos:
-                # article has videos but Glencoe 404, wait and then retry again
+                # article has videos but Glencoe 404, retry this activity again
                 end_event = self.end_event_glencoe_retry(exception, article_id, version, run)
-                time.sleep(60)
                 return end_event, self.ACTIVITY_TEMPORARY_FAILURE
             else:
                 end_event = self.end_event_glencoe_404(article_id, version, run)
--- a/activity/activity_CopyGlencoeStillImages.py
+++ b/activity/activity_CopyGlencoeStillImages.py
@@ -100,6 +100,30 @@ class activity_CopyGlencoeStillImages(Activity):
                 metadata, article_id, version, run)
         return end_event, result
 
+    def end_event_glencoe_retry(self, exception, article_id, version, run):
+        self.logger.info(exception)
+        return [self.settings, article_id, version, run, self.pretty_name, "error",
+                "Glencoe video is not available for article " + article_id + 
+                '; message: ' + str(exception)]
+
+    def end_event_glencoe_404(self, article_id, version, run):
+        self.logger.info(
+            "Glencoe returned 404, therefore article %s does not have videos", article_id)
+        return [self.settings, article_id, version, run, self.pretty_name, "end",
+                "Glencoe returned 404, therefore article has no videos"]
+
+    def end_event_checking_error(self, exception, article_id, version, run):
+        self.logger.exception("Error when checking/copying Glencoe still images.")
+        return [self.settings, article_id, version, run, self.pretty_name, "error",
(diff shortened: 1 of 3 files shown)


activity/activity_CopyGlencoeStillImages.py
```diff
@@ -36,50 +38,135 @@ def do_activity(self, data=None):
...
+
+    def handle_glencoe_metadata_assertion(self, exception, article_id, version, run, has_videos):
+        """logic for if a glencoe_check.metadata assertion is raised"""
+        self.logger.info(str(exception))
+        first_chars_error = str(exception)[:21]
+        if first_chars_error == "article has no videos":
+            if has_videos:
+                # article has videos but Glencoe 404, wait and then retry again
+                end_event = self.end_event_glencoe_retry(exception, article_id, version, run)
+                time.sleep(60)
```
giorgiosironi: we shouldn't be waiting inside an activity as it keeps that worker busy doing nothing; shouldn't this be managed at the SWF level? Can't remember if pauses can be made between activity retries. 
Can't find relevant documentation, but aren't we doing something similar elsewhere?

gnott: A very good point, thanks for mentioning it.

There is a simple, longstanding architectural "hack" in workflows here https://github.com/elifesciences/elife-bot/blob/develop/workflow/objects.py#L316, a sleep of 10 seconds. The simple SWF usage here would cause a failed activity to be tried again as immediately as possible, and in early testing that could mean trying it hundreds of times in quick succession. I added the 10 second sleep in between activity retries, which does (in practice) block a decider for that period of time, but it does free up workers for doing other things.

I think there's an SWF way of delaying a workflow, but we didn't implement that. I don't think we want to get further integrated into the SWF framework, so we probably do not want to do it the SWF way.

I don't like the sleep or blocking either. In this case we might be able to do away with the 60 second sleep, and to rely on the default 10 second decider sleep. We might be able to continue the workflow sooner if Glencoe produces the still images within seconds. The risk is the workflow will block a decider for as long as the workflow is active, if, for example, Glencoe never produces still images due to file errors, network downtime, or others. If we have at least five deciders running, we'd need five concurrent publishing workflows for articles with videos, that have no still images, to result in the bot potentially grinding to a halt (depending on how workflow decisions are made inside SWF, if it is round-robin, or prioritised differently, which I do not know).

Easiest suggestion: we remove `time.sleep(60)` here and rely on the 10 second blocking mechanism.

More complex suggestion: we look at how to improve holding workflow execution in a non-blocking manner.

giorgiosironi: I'd go for the easiest, as understanding and debugging more of SWF (a proprietary solution) is not what we should be spending time with if we can avoid it. In case we need more `worker` or `decider processes, having them sleeping will not impact CPU so the bottleneck to check and potentially expand will be memory.

gnott: Thanks, I agree for easiest, considering we do not have problems with blocking processes as it stands now.

gnott:
Tests are green, I am merging to see how end-to-end tests go.

PR #922, opened 2019-06-20
Switch all EJP CSV files over to the new naming.

gnott:
Regarding issue https://github.com/elifesciences/elife-bot/issues/916, the CSV file names are switched over now. This is using more precise name matching for the CSV file names, no longer supporting the older file names. Renamed files in the test data to reflect this and satisfy the tests for `PackagePOA` activity.


gnott:
There's not much to see here logic-wise. If tests are green I will merge it with no additional review.

PR #924, opened 2019-06-20
PMC resupply clean-up

gnott:
Regarding issue https://github.com/elifesciences/issues/issues/4889, the PMC resupply procedure is done now. We do not intend to use it again.

Here, the changes that added the new type of workflow are reversed. I did this by finding the terms in the code base and changing them until none were left.

Code tests are green (I think) so it is probably safe to merge. I was carefuly to check all the `not in` are now `!=` and `in` are `==` when matching workflow comparisons, but I might have missed something, if you get a chance to look it over @giorgiosironi - thanks!


gnott:
👍 thanks for taking a look - merging!

Issue #925, opened 2019-06-28
Refactor PMCDeposit activity


gnott:
Recently a small bug in PMC deposits was observed, where it didn't find the article XML file correctly, since the article had more than one `.xml` files in the zip package.

I am fixing that bug, and the code for this activity looks like it can be improved further.

Todo

- [x] Fix the bug for finding the correct article XML file
- [x] Refactor how the activity directories / folders are used
- [x] Use file functions from the `article_processing` provider
- [x] Refactor for functions with no self-use
- [x] Remove all the `if self.logger` conditional statements
- [x] Use the `ftp.py` provider for FTP functions
- [x] Switch all S3 operations over to use storage context from the `storage_provider`
- [x] More refactoring for better tests
- [x] Increase code coverage
- [x] Code linting

gnott:
Steps are completed in PR https://github.com/elifesciences/elife-bot/pull/930.

PR #930, opened 2019-07-18
Refactor PMCDeposit activity

gnott:
Regarding issue https://github.com/elifesciences/elife-bot/issues/925

A list of changes on the issue above are done to `activity/activity_PMCDeposit.py`.

I think the primary motivation was to fix a bug with finding the article XML file, and the original code was confused when there were more than one `.xml` file in the zip package. After reviewing the activity, I identified many ways the code could be cleaned up while working on it.

Using some existing functions from `provider/article_processing.py` (and improving them, I think) made it possible to remove some duplicate code. I think a new function was moved there too.

I also changed a hard-coded `elife` as the journal name, which instead gets the journal name from the zip file name now.

It will now return `self.ACTIVITY_TEMPORARY_FAILURE` as the result if it is unable to download the PMC zip file specified. The older logic was more convoluted. I'm not sure the retry adds much value, except to guard against temporary network or disk errors.

Added a new `FakeFTP` class for testing, which results in higher test coverage when mocking FTP upload logic.

I hope the files here are an improvement over the original code. This is not perfect, but I think it is more straightforward.

### DIFF
--- a/activity/objects.py
+++ b/activity/objects.py
@@ -49,6 +49,7 @@ class Activity(object):
 
         self.tmp_base_dir = "tmp"
         self.tmp_dir = None
+        self.directories = None
 
     def describe(self):
         """
@@ -183,13 +184,22 @@ class Activity(object):
         """
         Create the directories in the activity tmp_dir
         """
-        if not dir_names:
+        if not dir_names and self.directories and hasattr(self.directories, "values"):
             dir_names = self.directories.values()
+        if not dir_names:
+            self.logger.info("No dir_names to create in make_activity_directories()")
+            return None
+        # Now try to find or make directories
         for dir_name in dir_names:
+            if os.path.isdir(dir_name):
+                # directory exists, continue
+                continue
             try:
                 os.mkdir(dir_name)
-            except OSError:
-                pass
+            except OSError as exception:
+                self.logger.exception(str(exception))
+                raise
+        return True
 
     def open_file_from_tmp_dir(self, filename, mode='r'):
         """
--- a/activity/objects.py
+++ b/activity/objects.py
@@ -179,6 +179,18 @@ class Activity(object):
 
         return self.tmp_dir
 
+    def make_activity_directories(self, dir_names=None):
+        """
+        Create the directories in the activity tmp_dir
+        """
+        if not dir_names:
+            dir_names = self.directories.values()
+        for dir_name in dir_names:
+            try:
+                os.mkdir(dir_name)
+            except OSError:
+                pass
+
     def open_file_from_tmp_dir(self, filename, mode='r'):
         """
         Read the file from the tmp_dir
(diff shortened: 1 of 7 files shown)


gnott:
Also added `make_activity_directories()` to the base Activity object, which is an attempt at simplifying the folders an activity makes. It is related to issue https://github.com/elifesciences/issues/issues/3644

activity/objects.py
```diff
@@ -179,6 +179,18 @@ def get_tmp_dir(self):
...
+        """
+        Create the directories in the activity tmp_dir
+        """
+        if not dir_names:
+            dir_names = self.directories.values()
+        for dir_name in dir_names:
+            try:
+                os.mkdir(dir_name)
+            except OSError:
+                pass
```
giorgiosironi: ignore if it already exists? but it will also ignore permission errors or similar.

provider/article_processing.py
```diff
@@ -15,44 +17,57 @@ def list_dir(dir_name):
...
         else:
             if logger:
                 logger.info('there is no renamed file for ' + filename)
+    return file_name_map
+
+
+def rename_files_remove_version_number(files_dir, output_dir, logger=None):
+    """
+    Rename files to not include the version number, if present
+    Pre-PPP files will not have a version number, for before PPP is launched
```
giorgiosironi: outdated comment, the term `PPP` is not used anymore?

activity/objects.py
gnott: That's certainly part of the improvements that can be introduced here. Two parts at least I will try to add:
1. Check if the folder exists first before trying to make it
2. If there's an exception, log / raise the exception

It could be that we could also have different error messages depending on the type of `OSError`, but we can maybe refine those later on after we see some logged examples of real situations.

provider/article_processing.py
gnott: Agreed, I think `PMCDeposit` has been around for a long time, because I removed some old logic to account for multiple article zip files, or depositing from different buckets, probably used during the transition to eLife taking over the responsibility of doing PM deposits almost four years ago.

activity/objects.py
gnott: I expanded on these (hopefully in a good direction), with some logging, and some specific test cases too.

activity/objects.py
```diff
@@ -183,13 +184,22 @@ def make_activity_directories(self, dir_names=None):
...
         """
-        if not dir_names:
+        if not dir_names and self.directories and hasattr(self.directories, "values"):
             dir_names = self.directories.values()
+        if not dir_names:
+            self.logger.info("No dir_names to create in make_activity_directories()")
+            return None
+        # Now try to find or make directories
         for dir_name in dir_names:
+            if os.path.isdir(dir_name):
```
giorgiosironi: good not to use exceptions for normal cases like the directory already existing

gnott:
Thanks for the comments. Tests are passing, and I think all requirements are met. I'm going to merge and deploy, then I can watch the prod system tomorrow to see if there are any errors there.

Issue #921, opened 2019-06-20
PublicationEmail change logic of when to send emails


gnott:
Part 2 in the plan to not use SimpleDB anymore (https://github.com/elifesciences/elife-bot/issues/565), it includes a change to the `PublicationEmail` activity.

Instead of relying on the SImpleDB data to determine which emails were sent already, change the logic of when to send emails to look at the article type, status, and whether it is the "first vor" to determine when to send publication emails.

More changes as part of clean up
- [x] Change the logic for when to send what email (the basis of this issue) - defined as only the first version of each status (first poa or first vor)
- [x] To handle silent correction logic, put this into the `ScheduleDownstream` activity, because we can rely on the workflow session value there. Once it is queued as a `PublicationEmail`, we do not know whether the XML was part of a silent correction or not
- [x] Send emails directly by SMTP instead of using the SimpleDB email queue
- [x] Remove `get_editors()`, it is not used
- [x] Remove `send_email_testrun()` and other test run logic, we haven't used this recently
- [x] Remove logic for potentially sending duplicate emails, we don't use this, it was only for testing purposes
- [x] Switch all direct S3 bucket usage over to `storage_context` / `storage_provider`
- [x] Refactoring and increase test coverage, where possible
- [x] Linting code (as usual)
- [x] **Deployment note:** make sure the S3 outbox folder is empty when deploying this new logic, otherwise some duplicate emails will get sent for those articles in the outbox (actually on looking at it, only article types that we do not send to remain in the outbox, all other articles are cleared each day, but it is probably good to check the outbox contents when deploying to be sure)

gnott:
Regarding articles of type `Insight`, I thought the new logic might not work, but after further thought it might afterall.

We send emails about Insight articles as two types:

1. VoR and Insight is published on the same day: we can send them the VoR is published email, with another paragraph shown that says "and Insight was also published..."
2. If the Insight is published later: we send a "An Insight was published about your article" email

The new logic can work properly if when there is an Insight article to send, to check whether it is the first VoR of the Insight. If there is no research article in the same batch, then send the "An Insight was published..." email.

If there's a VoR and an Insight in the same batch, it can still send the template as described in point 1 above.

Make sure the test scenarios include these situations. I think the existing ones may already represent them, except for the non-first-VoR Insight article situation.

gnott:
[...]

PR #931, opened 2019-07-26
Refactoring of PublicationEmail activity and logic

gnott:
Regarding issue https://github.com/elifesciences/elife-bot/issues/921.

The primary purpose of opening up and changing `PublicationEmail` is to remove its reliance on SimpleDB.

The first way it depended on SimpleDB was to check whether a particular type of email was sent already to an author for a particular article, so it would not send a duplicate of that email again. This logic has been moved to the `ScheduleDownstream` activity, which will only consider the first version (first PoA or first VoR) when deciding to send an author an email that their article is published. It also does not send emails for articles that are published as a silent correction.

The second way the activity depended on SimpleDB was to use the SimpleDB email queue. In this updated code, the activity instead will send emails direct by SMTP when the activity runs, rather than using the old queue.

We also no longer consider the blacklist of articles, to not send publication emails for any article published prior to when we launched `PublicationEmail` (that would have had no email history to check duplicates against).

All direct S3 connection usage is switched over to use the `storage_provider`.

I removed old code for sending test emails and anything else that no longer applies.

There's some significant refactoring to create non-self-use functions, increase the test coverage, and try to make the logic less convoluted and nested.

There is still some potentially confusing logic when instantiating articles, setting related articles, getting author emails, but at this point I think further refactoring would not gain much more for the further time spent.


gnott:
I think I will merge this PR and deploy, monitor the results and add fixes if required.

PR #932, opened 2019-07-26
Python 3 tweaks

gnott:
Part of the issue for Python 3 support https://github.com/elifesciences/issues/issues/4711

Because there is not 100% code coverage of the project, I was wondering whether there was any Python 3 incompatible code lurking in the untested parts. I used `2to3` to get recommendations of code changes.

More or all of the changes here are not too major nor are they incompatible with Python 3 (I think). 

- A couple `print()` statements with added parentheses
- removed `long` where mentioned
- replaced many `lambda` with list comprehensions (the suggestions provided by `2to3` were looking good)
- lots of `dict.items()`, `dict.keys()` etc. when used as an iterator are wrapped in a `list` call, e.g. `for key, value in list(my_dict.items()):`, because it seems implicitly casting as a `list` is preferred over letting it automatically be converted
- EDIT: also a usage of `unicode()` replaced (which failed the tests) in `dashboard_queue.py`

### DIFF
--- a/S3utility/s3_sqs_message.py
+++ b/S3utility/s3_sqs_message.py
@@ -37,6 +37,6 @@ class S3SQSMessage(Message):
         """
         if body is not None and len(body) > 0:
             self.payload = json.loads(body)
-        if body and 'Records' in self.payload.keys():
+        if body and 'Records' in list(self.payload.keys()):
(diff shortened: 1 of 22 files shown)


gnott:
Small changes, and tests are passing, I will merge this myself.

Issue #753, opened 2018-08-08
Activity directories created can possibly be standardised


gnott:
Some activities create subdirectories in the activity's temp dir where files are saved and processed. The activity base class may be able to create them, some of the folder names could be in common. If not that then each activity may be able to define its directory names and then get an activity method to create the actual directories.

Related to feedback in PR https://github.com/elifesciences/elife-bot/pull/750

PR #933, opened 2019-07-30
Refactor creating activity directories

gnott:
Re issue https://github.com/elifesciences/issues/issues/3644
Fixes https://github.com/elifesciences/elife-bot/issues/753

Based on the sample used in the `PMCDeposit` activity refactoring, this branch continues that theme of defining a dict `self.directories` in an activity class, creating those folders using `self.make_activity_directories()`, and then the folders can be referred to from the dict of folder names.

Some tests required changing to get them to work, if they needed the folder to exist first before mocking a call.

I used an `OrderedDict()` in `PackagePOA` because the `CSV_TMP` folder is a folder inside the `CSV` folder, and then we can be assured the parent folder exists before it tries to create the child folder.

### DIFF
--- a/activity/activity_ArchiveArticle.py
+++ b/activity/activity_ArchiveArticle.py
@@ -25,6 +25,11 @@ class activity_ArchiveArticle(Activity):
         self.description = "Archive an article post-publication"
         self.logger = logger
 
+        # Local directory settings
+        self.directories = {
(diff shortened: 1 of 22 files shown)


gnott:
One change to get all tests green. I will merge this and we can see if end2end tests are also passing.

Issue #936, opened 2019-08-08
DepositCrossref activity refactoring


gnott:
Soon we will want to deposit peer review metadata to Crossref. I think a good way to run it will be to create a new bot activity for that purpose. That new activity may be able to share some of the logic in the existing `DepositCrossref` activity, possibly limited to where it checks whether an article is suitable to deposit (which holds back files that are not published yet).

In addition to any restructuring of the logic to potentially reuse, there are some standard refactoring tasks I am planning, which includes:

- [ ] Use `storage_provider` for all S3 bucket operations, instead of using direct `S3Connection`
- [ ] Send email direct by SMTP instead of using the old email queue
- [ ] Refactor the activity status tracking into the `dict` style tracking of `statuses`
- [ ] As mentioned above, see if any logic can be moved for sharing with a new activity, such as checking whether an article is suitable to deposit
- [ ] Expand on these todo items as they come up in refactoring

gnott:
Bug in the code from PR 938 deployed today.

I am deploying a bug fix in PR https://github.com/elifesciences/elife-bot/pull/939

PR #938, opened 2019-08-15
DepositCrossref refactor and new crossref provider

gnott:
In regards to issue https://github.com/elifesciences/elife-bot/issues/936.

The primary motivation was to make any logic that could be reused in the `DepositCrossref` activity available for reuse, resulting in a new `crossref.py` provider module.

Additionally, the steps of refactoring in the issue mentioned above was done, including storage logic, emailing, activity status tracking.

I moved `get_admin_email_body_foot()` over to `email_provider.py`, because it could also be reused, and is why the `PublicationEmail` module is altered on this PR.

In the end, I managed to move as much of the logic to `crossref.py` that I think could be reused, including the bucket upload and downloading, Crossref API communication, parsing XML into objects, and admin email subject and body content.

Soon, when we add a new activity for depositing peer reviews to Crossref, building that new activity will be much easier, and result in less duplicate code.

### DIFF
--- a/provider/email_provider.py
+++ b/provider/email_provider.py
@@ -243,3 +243,18 @@ def valid_recipient_object(recipient):
     if recipient.e_mail is not None and str(recipient.e_mail).strip() == "":
         return False
     return True
+
+
+def get_admin_email_body_foot(activity_id, workflow_id, datetime_string, domain):
+    """Format the footer of the body of the email"""
+    body_foot = ""
+    body_foot += "\n"
+    body_foot += "-------------------------------\n"
+    body_foot += "SWF workflow details: " + "\n"
+    body_foot += "activityId: " + str(activity_id) + "\n"
+    body_foot += "As part of workflowId: " + str(workflow_id) + "\n"
+    body_foot += "As at " + datetime_string + "\n"
+    body_foot += "Domain: " + domain + "\n"
+    body_foot += "\n"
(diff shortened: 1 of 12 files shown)


provider/email_provider.py
```diff
@@ -243,3 +243,18 @@ def valid_recipient_object(recipient):
     if recipient.e_mail is not None and str(recipient.e_mail).strip() == "":
         return False
     return True
+
+
+def get_admin_email_body_foot(activity_id, workflow_id, datetime_string, domain):
+    """Format the footer of the body of the email"""
+    body_foot = ""
```
giorgiosironi: is there a better way to build this than concatenating strings? Like reading an external `.txt` file? Don't want to go into full templating though...

gnott: I'm not sure how much we gain using an external file in this one example, but I am giving it a go. We already do use `jinja` templates in this project, instead I changed this function above to use `string.Template`. The string templates seems to be a pythonic way of doign some simple substitutions in a string.

gnott:
Thanks for looking it over and the PR approval. Hope my changes to load the email footer template are acceptable (if not, it can be changed in a subsequent PR). There's no big rush to merge this PR, but it does block me from starting on Crossref peer review logic, so I will merge it now, and we can see if this all works in `end2end` and `prod` environments.

PR #941, opened 2019-08-23
Cron refactoring and is timezone aware

gnott:
In reference to issue https://github.com/elifesciences/issues/issues/4221

I picked up the issue because last time we had to change the bot cron to take into account the changes to British Summer Time, I wanted it to be the last time we had to do it. Considering the clocks will change back in October, this will make it a non-event.

I wanted to be able to define cron workflow starts based on UTC (normal) and some using the local time zone. We set and support `"Europe/London"` as the time zone here, but the logic would only work best with timezones that have hourly increments of utc offsets (no half or quarter hour offsets), due to the legacy logic of how the current cron bases it on the UTC clock minute value for when to start workflows. We use the `pytz` for time zones.

I switched from `OptionParser` to `ArgumentParser` because a linter said `OptionParser` was deprecated. I moved the logic to parse args and load settings so it has more test coverage.

The previously long and complicated functions are split into parts. Each new function then requires fewer arguments than before, and it makes them easier to write test cases for them. For these, the starter specs are in an `OrderedDict()` in the big list of cron jobs.

Conditional starting of the `PubmedArticleDeposit` workflow no longer looks at bucket contents to decide whether a workflow should start, since that situation is already handled by the `PubmedArticleDeposit` activity itself.

The old `eval()` is gone, loading the class name dynamically using other means.

There's a more precise time window of `30` to `44` minutes of the hour, because I found sometimes it would consider running a workflow in both the `30` to `59` slot and the `45` to `59` slot. This makes the time slots not to overlap.

I also added some logging so we get a new `cron.log` log file as part of these changes.

Test coverage is increased, including checking at each time which particular starters would be conditionally invoked.

In summary, we get UK BST/GMT workflow, UTC workflows (for those where timezone does not matter), refactored for more clarity, better test covers, and a little bit of logging.


gnott:
I was a little confused with the `pytz` result at first because it wasn't doing what I expected. I was relying on `datetime.hour` to return the actual hour of the new time, such as `12:30` I wanted 12, but instead the offset of `+01:00` was taken into account and `datetime.hour` returned 11. So, `pytz` correctly set the utc offset, but I had to calculate a new `datetime` object, doing the time delta calculation and then removing the offset, so I could get the hour `12` the way I wanted. It made it easier when I split it up into its own function and write some tests to be certain it was doing what I wanted.

Issue #692, opened 2018-04-27
ScheduleCrossref for silent correction workflow, only do for the most recent article version


gnott:
``ScheduleCrossref`` activity was removed from the ``SilentCorrectionProcess`` workflow (in PR https://github.com/elifesciences/elife-bot/pull/507) when the features of the Crossref deposit were expanded. The deposit data included citations, data and text mining files, etc. If an older version was silently corrected, then Crossref may overwrite the most recent version data with older, incomplete data from a PoA article, for example.

The proposed improvement is in ``ScheduleCrossref`` to only schedule an article for Crossref deposit in a silent correction if it a silent correction of the most recent published version of the article. Any version previous to the currently published version would not be scheduled for Crossref deposit.

Could be related to issue https://github.com/elifesciences/elife-bot/issues/683 that would allow the logic to check the type of workflow being run as part of the conditional logic.

It is also possible (I think) for the logic to be compatible with silent correction and non-silent correction workflows: only schedule for Crossref deposit if it is the currently published version or greater than the published version (for ingest of new versions), or if no version exists (for ingest of v1 articles).

There is already some pub date embargo logic in the ``DepositCrossref`` workflow itself, which holds back any v1 article from being deposited until on its pub date and not before its pub date, to allow for ingesting articles early and scheduling them for coordinated press events.

@giorgiosironi @seanwiseman you may have comments on my thinking here and whether it is a good solution.

giorgiosironi:
bq. It is also possible (I think) for the logic to be compatible with silent correction and non-silent correction workflows: only schedule for Crossref deposit if it is the currently published version or greater than the published version (for ingest of new versions), or if no version exists (for ingest of v1 articles).

Need to enumerate the different scenarios, but this seems sound to me. I now understand what was the problem with blindly depositing to Crossref every silent correction.

gnott:
Also a possible task for @GiancarloFusiello if you're interested.

gnott:
I am going to pick up this issue now, along with some proposed `DepositCrossref` refactoring (for which I will create a new issue).

PR #944, opened 2019-08-28
Schedule crossref in silent corrections

gnott:
Fixes https://github.com/elifesciences/elife-bot/issues/692

For `run_type` of `"silent-correction"`, if the version of the article does not equal the highest version of the article already published, then log it and return `True`, not `"scheduling"` it for Crossref deposit.

This guards against overwriting Crossref metadata with a previous, older, version of article metadata when a silent correction is done for an older version, while allowing Crossref deposits to happen automatically if the most recent version of an article is silently corrected.

### DIFF
--- a/activity/activity_ScheduleCrossref.py
+++ b/activity/activity_ScheduleCrossref.py
@@ -54,7 +54,7 @@ class activity_ScheduleCrossref(Activity):
                 self.logger.info(
                     'ScheduleCrossref will not deposit article %s' +
                     ' ingested by silent-correction, its version of %s does not equal the' +
-                    ' highest version is %s', (article_id, version, highest_version))
+                    ' highest version which is %s', (article_id, version, highest_version))
                 return True
 
         conn = S3Connection(self.settings.aws_access_key_id,
--- a/activity/activity_ScheduleCrossref.py
+++ b/activity/activity_ScheduleCrossref.py
@@ -1,7 +1,7 @@
 import json
 
 from boto.s3.connection import S3Connection
-
+from provider import lax_provider
 from provider.execution_context import get_session
 from provider.article_structure import get_article_xml_key
 from activity.objects import Activity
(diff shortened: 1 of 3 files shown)


activity/activity_ScheduleCrossref.py
```diff
@@ -46,6 +46,16 @@ def do_activity(self, data=None):
         article_id = session.get_value('article_id')
         expanded_folder_name = session.get_value('expanded_folder')
 
+        # if is a silent-correction workflow, only deposit for the most recent article version
+        run_type = session.get_value("run_type")
+        if run_type == "silent-correction":
+            highest_version = lax_provider.article_highest_version(article_id, self.settings)
+            if str(version) != str(highest_version):
+                self.logger.info(
+                    'ScheduleCrossref will not deposit article %s' +
```
giorgiosironi: good logging

activity/activity_ScheduleCrossref.py
```diff
@@ -46,6 +46,16 @@ def do_activity(self, data=None):
         article_id = session.get_value('article_id')
         expanded_folder_name = session.get_value('expanded_folder')
 
+        # if is a silent-correction workflow, only deposit for the most recent article version
+        run_type = session.get_value("run_type")
+        if run_type == "silent-correction":
+            highest_version = lax_provider.article_highest_version(article_id, self.settings)
+            if str(version) != str(highest_version):
```
giorgiosironi: why not comparing integers rather than strings? may still require casting to `int` of course

gnott: I probably just defaulted here casrting as `str` because it seemed the safest since just about everything can be cast as a string and we wouldn't risk a runtime exception.

I couldn't remember where we use `str` and `int` for version numbers between the Lax responses, parser, and session value.

I think we're only concerned with `version` as an `int` when we want to know whether a `version` is greater than or less than another `version`. In the above, we are only concerned whether it matches the highest version value exactly or not exactly.

activity/activity_ScheduleCrossref.py
gnott: Going to make a slight change to the grammar though. You prompted me to read it again more carefully!

gnott:
Thanks for reviewing @giorgiosironi - I will merge it in if green after my grammar fix.

gnott:
The last attempt at running tests failed, I think due to the testing of Python 3 on the `ci` environment, and the linting errors uncovered. All the linting fixes are merged back into this branch and the tests are green again. I will merge this PR since the features were approved already.

PR #946, opened 2019-08-30
Fix Exception has no 'message' member (no-member).

gnott:
I noticed the linting output in a build log, where the code had `e.message` when logging an `Exception`. Switched these over to `str(exception)`, and renamed any single character variable names.


gnott:
Might leave this PR open for now, because it will probably have merge conflicts with the `python3` branch, where these same Exception message lines were changed in a different way.

gnott:
These were fixed in a different PR and the conflicts are not worth resolving, closing.

PR #947, opened 2019-08-30
Feature: tweaked Python3 installation

lsh-0:
* removes Travis + Python2.7, we shouldn't be making concessions to it anymore.
* changes travis-ci environment to 'xenial', which is 16.04, which is the environment it's currently running in now at elife
* re-introduces the mkvenv.sh script, no need to cannibalise it
* fixes a `python-docx` issue by bumping `setup-tools` alongside `pip`
* fixes a bunch of `Exception.message` cases which are no longer viable in Python3
* bumps pylint for python3.6

outstanding are some actual factual code problems that need to be addressed (which is why we use `pylint`):
* that evil `eval` in settings.py is causing these:

```
activity/activity_UpdateRepository.py:42:47: E1101: Module 'settings' has no 'live' member (no-member)
activity/activity_UpdateRepository.py:43:52: E1101: Module 'settings' has no 'end2end' member (no-member)
```

* the tests could do with a scrub as all of these are legitimate as well:

```
************* Module tests.activity.test_activity_email_video
tests/activity/test_activity_email_video.py:115:8: E1111: Assigning result of a function call, where the function has no return (assignment-from-no-return)
tests/activity/test_activity_email_video.py:141:8: E1111: Assigning result of a function call, where the function has no return (assignment-from-no-return)
tests/activity/test_activity_email_video.py:172:8: E1111: Assigning result of a function call, where the function has no return (assignment-from-no-return)
tests/activity/test_activity_email_video.py:201:8: E1111: Assigning result of a function call, where the function has no return (assignment-from-no-return)
************* Module tests.activity.test_activity_package_poa
tests/activity/test_activity_package_poa.py:125:8: E1111: Assigning result of a function call, where the function has no return (assignment-from-no-return)
tests/activity/test_activity_package_poa.py:221:8: E1111: Assigning result of a function call, where the function has no return (assignment-from-no-return)
tests/activity/test_activity_package_poa.py:224:8: E1111: Assigning result of a function call, where the function has no return (assignment-from-no-return)
************* Module tests.activity.test_activity_publication_email
tests/activity/test_activity_publication_email.py:189:12: E1111: Assigning result of a function call, where the function has no return (assignment-from-no-return)
tests/activity/test_activity_publication_email.py:288:8: E1111: Assigning result of a function call, where the function has no return (assignment-from-no-return)
tests/activity/test_activity_publication_email.py:326:8: E1111: Assigning result of a function call, where the function has no return (assignment-from-no-return)
************* Module tests.activity.test_activity_publish_final_poa
tests/activity/test_activity_publish_final_poa.py:210:12: E1111: Assigning result of a function call, where the function has no return (assignment-from-no-return)
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,6 +1,10 @@
 #!/bin/bash
 set -e
 
+if [ "$ENVIRONMENT_NAME" == "ci" ]; then
+    rm -rf venv/
+fi
+
 . mkvenv.sh
 
 echo "Virtualenv Python: $(venv/bin/python --version)"
@@ -14,6 +18,13 @@ find starter/ -maxdepth 1 -name '*.pyc' -delete
 find S3utility/ -maxdepth 1 -name '*.pyc' -delete
 
 source venv/bin/activate
+
+# fixes issues installing wheel packages, hides deprecation warnings
+pip install "pip~=19.2" --upgrade
+
+# fixes python-docx issue #594 
+# https://github.com/python-openxml/python-docx/issues/594
+pip install -U setuptools
 
 grep "git+" requirements.txt > source-requirements.txt
 #pip uninstall -r source-requirements.txt -y
--- a/install.sh
+++ b/install.sh
@@ -1,30 +1,7 @@
 #!/bin/bash
 set -e
 
-python=''
-pybinlist=("python3.6" "python3.5" "python3.4")
-
-for pybin in ${pybinlist[*]}; do
-    which "$pybin" &> /dev/null || continue
-    python=$pybin
-    break
-done
-
-if [ -z "$python" ]; then
-    echo "no usable python found, exiting"
-    exit 1
-fi
-
-echo "Found Python: $($python --version)"
-
-if [ "$ENVIRONMENT_NAME" == "ci" ]; then
-    rm -rf venv/
-fi
-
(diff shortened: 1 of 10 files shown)


install.sh
```diff
@@ -37,7 +18,14 @@ find starter/ -maxdepth 1 -name '*.pyc' -delete
 find S3utility/ -maxdepth 1 -name '*.pyc' -delete
 
 source venv/bin/activate
-pip install --upgrade pip
+
+# fixes issues installing wheel packages, hides deprecation warnings
+pip install "pip~=19.2" --upgrade
```
giorgiosironi: stop gap, but shouldn't this be done in https://github.com/elifesciences/builder-base-formula/blob/77354f7cc342eff4c724f23e92078fa47500d202/elife/python.sls#L36-L41?

giorgiosironi: though not if it is pinned maybe

gnott:
I will merge this to continue troubleshooting if required.

install.sh
lsh-0: I'm not sure it even needs to be upgraded at all to be honest, but yeah - if it's to be upgraded anywhere it should be in the 'host' so that the virtualenv doesn't need to. I'll take a look

PR #945, opened 2019-08-28
Install Python 3 venv, test on py36 in travis-ci.

gnott:
Very basic switch in `install.sh` to use Python 3, I shamelessly copied most of it from https://github.com/elifesciences/bot-lax-adaptor/blob/develop/mkvenv.sh.

I updated the travis-ci tests to run in both py27 and py36 for now.

There could be more to be done in builder things to use Python 3, please comment on what those may be, if you can think of them.

Rolling this out, the old `venv` folder may need to be deleted, or new instances created fresh.

PR #948, opened 2019-09-06
Fix bytes to str in execution_context S3Session.

gnott:
Bug fix for PR https://github.com/elifesciences/elife-bot/pull/945 which is not passing `end2end` tests.

There was (and still is) poor test coverage for `provider/execution_context.py`, where a bug with converting an S3 key `bytes` value to `str` failed.

Here, convert the value using an existing function from `provider/utils.py` in order to load it with `json.loads()`.

If the tests are green in this PR, I will merge this to see if `end2end` testing can progress futher, possibly uncovering more Python 3 compatibility glitches.

### DIFF
--- a/provider/execution_context.py
+++ b/provider/execution_context.py
@@ -4,6 +4,7 @@ from boto.s3.key import Key
 from boto.exception import S3ResponseError
 from pydoc import locate
 import json
+from provider.utils import unicode_encode
 
 
 def get_session(settings, input_data, session_key):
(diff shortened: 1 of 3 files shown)


gnott:
Next error report,

```
2019-09-06T17:26:12Z ERROR worker_5053 cannot use a string pattern on a bytes-like object
  File "/opt/elife-bot/activity/activity_VerifyGlencoe.py", line 64, in do_activity
    if glencoe_check.has_videos(xml_content):
  File "/opt/elife-bot/provider/glencoe_check.py", line 75, in has_videos
    if re.search(r'<media[^>]*mimetype="video".*?>', xml_str):
  File "/usr/lib/python3.5/re.py", line 173, in search
    return _compile(pattern, flags).search(string)
TypeError: cannot use a string pattern on a bytes-like object
```

PR #951, opened 2019-09-06
Cast content as str in UpdateRepository.

gnott:
Next error in Python 3 upgrade seems to point to updating the article XML git hub repository. I am making a guess at which content to convert from a bytes-like object to `str` based on the error in the log file, avoiding to create a more elaborate test scenario to test adding to a git repo for now.


gnott:
Related to PR https://github.com/elifesciences/elife-bot/pull/945

PR #954, opened 2019-09-11
VerifyGlencoe again just before IngestToLax.

gnott:
Potential improvement raised and discussed on https://github.com/elifesciences/issues/issues/5074.

Even though we verify the video files exist earlier in a publication workflow, the video metadata may be missing when Lax looks for it.

Here, verify the video metadata is present just prior to ingesting to Lax. `VerifyGlencoe` knows if an article has videos, and if the video metadata is not there, it will retry 10 seconds later until the metadata is available.


gnott:
I'm going to merge this myself due to the errors we've been getting recently, I'd like to deploy it ASAP if it passes `end2end` tests.

Issue #964, opened 2019-10-08
Big Query provider


gnott:
In reference to internal issue https://github.com/elifesciences/issues/issues/5105, this ticket represents the `elife-bot` side of things to use Big Query in Crossref peer review deposits.

I created a dummy `provider/bigquery.py` provider module in the `crossref-peer-review` branch of this project: https://github.com/elifesciences/elife-bot/blob/crossref-peer-review/provider/bigquery.py  It currently just returns one date which is used in tests and activity logic.

Still to do for a Big Query provider will probably include:

- [x] Credentials to connect to Big Query
- [x] Logic to connect to Big Query
- [x] Composing a query
- [x] Formatting article date data, if required, for inclusion in Crossref peer review deposits
- [x] In future, it may also gather some data about editors of an article

gnott:
Can wait until the code is merged and this issue will be completed.

PR #971, opened 2019-10-31
Crossref peer review deposits and BigQuery provider

gnott:
Here is where I am up to with adding a workflow and activity for depositing peer review data to Crossref. There are a few bits of clean up, refactoring, and also enhanced logic to add later with regards to adding missing ORCID values for editors.

I am opening this PR to demonstrate how it is currently successfully connecting to BigQuery (on my local system) to run queries, by passing in a credentials file and a project ID. The credentials also seemed to need a scope for it to work. This should be found mostly in the `provider/bigquery.py`.

The ticket about adding the bigquery provider is https://github.com/elifesciences/elife-bot/issues/964. 

Most applicable I think to my reason for opening a PR at this time is for https://github.com/elifesciences/issues/issues/5178 which I would like to bring to your attention again please @giorgiosironi. Here, my understanding is the location of the Google credentials may be specified in an environment var, and/or there may be some Google cloud config added where credential files are not required, or the project ID doesn't need to be passed to the `Client()`, or the `SCOPE` of the credentials will also already be set.

I would like to leave this PR and if you could please advise what I should change to use the credentials on the infrastructure as you choose to define them.

I think the tests I added for the `bigquery.py` provider here is mocked so much that it will be easy to change how it instantiates a `Client()` to not require a credentials file, at least I hope so.

We can discuss options about credentials on the other separate issue if you would prefer @giorgiosironi.

The logic in the PR may change slightly before it is merged, so please do not be concerned with reviewing the new code here in great detail, unless you have the time to read it and offer suggestions.


Melissa37:
@giorgiosironi can you review this or do I need to ask @Maelplaine to add it to the board he manages?

activity/activity_DepositCrossrefPeerReview.py
```diff
@@ -0,0 +1,226 @@
...
+            self.logger.info("Moving files from outbox folder to published folder")
+            to_folder = crossref.get_to_folder_name(self.published_folder, date_stamp)
+            crossref.clean_outbox(self.settings, self.publish_bucket, self.outbox_folder,
+                                  to_folder, self.good_xml_files)
+            # copy the Crossref deposit XML files to the batch folder
+            batch_file_names = glob.glob(self.directories.get("TMP_DIR") + "/*.xml")
+            batch_file_to_folder = to_folder + "batch/"
+            crossref.upload_crossref_xml_to_s3(
+                self.settings, self.publish_bucket, batch_file_to_folder, batch_file_names)
+            self.statuses["outbox"] = True
```
giorgiosironi: `else` log a warning/info message?

activity/activity_DepositCrossrefPeerReview.py
```diff
@@ -0,0 +1,226 @@
...
+
+        # Set the activity status of this activity based on successes
+        self.statuses["activity"] = bool(
+            self.statuses.get("publish") is not False and
+            self.statuses.get("generate") is not False)
+
+        # Send email
+        # Only if there were files approved for publishing
+        if self.good_xml_files:
+            self.statuses["email"] = self.send_admin_email(outbox_s3_key_names, http_detail_list)
```
giorgiosironi: also `else` could log something here

provider/bigquery.py
```diff
@@ -0,0 +1,113 @@
...
+        # populate the primary values
+        for row_key, attr_name in list(self.attr_name_map.items()):
+            setattr(self, attr_name, row_dict.get(row_key, None))
+        # populate the roles
+        self.roles = row_dict.get('Roles', [])
+
+
+def get_credentials(settings):
+    credentials = service_account.Credentials.from_service_account_file(
+        settings.big_query_credentials_file)
```
giorgiosironi: good settings, but add to `settings-example.py` for documentation?

giorgiosironi: Seems feasible to provide a separate file from `elife-bot-formula` if this is what is required

giorgiosironi: @de-code does this fit what you expect?

requirements.txt
```diff
@@ -31,3 +31,4 @@ git+https://github.com/elifesciences/digest-parser.git@cd3a47af45db64a1731cc5ae9
 git+https://github.com/Medium/medium-sdk-python.git@bdf34becf6dd24e3b1fc1048c36416bcba9fe0d6#egg=medium
 psutil==5.4.6
 pytz==2019.2
+google-cloud-bigquery==1.21.0
```
giorgiosironi: minimal new dependencies :+1:

provider/bigquery.py
de-code: Personally I would recommend not to get the credentials explicitly (what I meant by as part of the [comment](https://github.com/elifesciences/issues/issues/5178#issuecomment-544862333)).

i.e. just call:

```python
bigquery.Client(project=project)
```

And set `GOOGLE_APPLICATION_CREDENTIALS` which is a standard approach. Which has the benefit that it also works with user level credentials, especially useful for development. (I am assuming we are not using the production service account for development)

giorgiosironi: https://github.com/elifesciences/elife-bot-formula/pull/47

gnott: Specifying the path to the JSON credentials in `GOOGLE_APPLICATION_CREDENTIALS` for me worked well and was pretty painless. The scope of the credentials must already be baked in because I could get rid of the load from file and get rid of scoping them. The latest code here reflects the simplified crendentialing method.

activity/activity_DepositCrossrefPeerReview.py
gnott: By the time it reaches this point, it is just uploading the generated XML files to our bucket for later reference, so it is unlikely we'll encounter an exception or error here (unless there is a networking or S3 issue), which I think will be handled by the overall worker exception.

I will add some logging around the earlier call of `self.deposit_files_to_endpoint` on line 88 where it is communicating with Crossref.

gnott: Ah, I see what you mean though, if `publish` status is False in this block, I will add another logging event.

activity/activity_DepositCrossrefPeerReview.py
gnott: 👍 more logging is better than less logging

gnott:
Thanks for the vote of confidence on approving this @giorgiosironi. I have since refactored a little, more tests, and the logic to set ORCID values. I am waiting to confirm whether the ORCID data is ok before merging this and considering to use it.

gnott:
This workflow is, I think, feature complete. I want to eventually use the workflow because the semi-manual process I am using for the first set of articles is a little slow and cumbersome. I am going to merge this now to try depositing a handful of files once deployed, checking for any improvements and bug fixes required in a `prod` situation.

gnott:
Tests shouldn't depend on the `GOOGLE_APPLICATION_CREDENTIALS` value being in the environment, nor does it (or should it) be using that. I see a bug in the test scenario where it used to mock getting credentials from file, I am changing it in a bug fix PR where it will override the default settings instead.

Issue #965, opened 2019-10-22
Decision letter parsing workflow


gnott:
To be done is to create a workflow and activities to process decision letter parser `.zip` files as they are added to a bucket.

- [x] First step is to fill out this issue with some tasks with checkboxes

**Summary**

This issue describes an implementation of processing decision letter & author response content. The majority of the parsing logic is held in the `decision-letter-parser` code library. The `elife-bot` can be used to define and run the steps of the processing workflow.

A `.zip` file, which contains a `.docx` file and optionally some figure images and/or video files is copied to an S3 bucket. An S3 notification from that bucket is read by a queue listener in the `elife-bot`. As a result, the bot will start a workflow. That workflow will have activities run in series to validate the input, generate JATS output from the `.docx` file contents, copy file assets to another S3 bucket, POST the JATS output to an API endpoint. This workflow will notify people if it encounters an error, or when everything completed successful, it will notify people of the completed status.

**Bot workflow and activity build out in brief:**

**Workflow parts to build**

- [x] Create workflow: name `IngestDecisionLetter`
- [x] Create a workflow starter for `IngestDecisionLetter`
- [x] Add code tests for the workflow and starter
- [x] Start an `IngestDecisionLetter` workflow execution when a file is uploaded to the input bucket

**Activities in this workflow to create** 

_(separate issues can accompany each for more detailed discussion)_

- [x] Note: will probably want to create a new provider, named `letterparser.py` perhaps, that can store any code to be shared by these activities, such as logic to unzip the contents, build from `.docx` files, and such
- [x] ValidateDecisionLetterInput (https://github.com/elifesciences/elife-bot/issues/966)
- [x] GenerateDecisionLetterJATS (https://github.com/elifesciences/elife-bot/issues/967)
- [x] DepositDecisionLetterIngestAssets (https://github.com/elifesciences/elife-bot/issues/968)
- [x] PostDecisionLetterJATS (https://github.com/elifesciences/elife-bot/issues/969)
- [x] DecisionLetterReceipt (https://github.com/elifesciences/elife-bot/issues/970)

**External infrastructure required for completion**

- [x] S3 bucket for decision letter input files (https://github.com/elifesciences/issues/issues/5146)
- [x] S3 bucket to store decision letter parser assets output (https://github.com/elifesciences/issues/issues/5109)
- [x] API endpoint to receive decision letter JATS XML (to be provided by ExeterPremedia)

**Assistance required from infrastructure team**

[...]

PR #973, opened 2019-11-08
DO NOT MERGE - Simple call to pandoc via Docker in letterparser.

gnott:
Re: build out of decision letter parsing workflow in issue https://github.com/elifesciences/elife-bot/issues/965

Infrastructure related issue https://github.com/elifesciences/issues/issues/5194

Here is my attempt to connect the `decision-letter-parser` library in the simplest way, and parse a simple `.docx` file using `pandoc` via a Docker image.

I hope this is a helpful illustration of one possible way this may work. I am able to run the test case (when I have Docker running), and it is also magicaly working on travis-ci testing.

I will open this PR as a draft for illustrative purposes for now, with room to refactor a grow the new provider slightly before possibly merging it, allowing further development of the workflow in separate PRs.


gnott:
I forgot to select "Draft" on the PR and I'm not sure if it can be changed now it's created. I'll put on a "DO NOT MERGE" title to it.

requirements.txt
```diff
@@ -9,6 +9,8 @@ git+https://github.com/elifesciences/elife-pubmed-xml-generation.git@814bcf15dac
 git+https://github.com/elifesciences/ejp-csv-parser.git@9fdfa6a37c00b6134f8a2a46c784b0b11b19c654#egg=ejpcsvparser
 git+https://github.com/elifesciences/jats-generator.git@ea0d4d90a28198eb2953cc982da85392fbfee360#egg=jatsgenerator
 git+https://github.com/elifesciences/package-poa.git@2649777a020661e507de4087d447f50ca8c2fe57#egg=packagepoa
+docker==4.0.2
```
giorgiosironi: this will probably be brought in as part of `decision-letter-parser`, but it's good to list it explicitly at the top level and to pin the version

tests/provider/test_letterparser_provider.py
```diff
@@ -0,0 +1,23 @@
...
+import provider.letterparser_provider as letterparser_provider
+
+
+class TestLetterParserProvider(unittest.TestCase):
+
+    def test_docker_pandoc_output(self):
+        """test calling pandoc via Docker"""
+        file_name = 'tests/fixtures/letterparser/sections.docx'
+        # fake config for specifying Docker image
+        config = {'docker_image': 'knsit/pandoc:v2.5'}
```
giorgiosironi: This may not help running `elife-bot` into a Docker container in the future, but there are ways to do Docker-in-Docker if that comes to pass.

giorgiosironi:
Not sure about the failure, `docker run hello-world` runs correctly on `elife-bot--ci` so the daemon is available.

giorgiosironi:
The error message is a problem as the library just returns `None` but doesn't throw an exception reporting what error lead to that.

giorgiosironi:
After the failure, `elife-bot--ci` doesn't seem to have pulled the image (checking with `docker image ls`).

provider/letterparser_provider.py
```diff
@@ -0,0 +1,16 @@
...
+import log
+
+IDENTITY = "process_%s" % os.getpid()
+LOGGER = log.logger("letterparser_provider.log", 'INFO', IDENTITY, loggerName=__name__)
+
+
+def docker_pandoc_output(file_name, config):
+    try:
+        return parse.docker_pandoc_output(file_name, config)
+    except docker.errors.APIError:
```
giorgiosironi: there is no possible recovery from a failure here, so remove `try`/`except` and let the exception bubble up

giorgiosironi: Removing leads to the error message:
```
E       docker.errors.ImageNotFound: 404 Client Error: Not Found ("no such image: knsit/pandoc:v2.5: No such image: knsit/pandoc:v2.5")
```

giorgiosironi: Manually pulling with `docker pull knsit/pandoc:v2.5`.

giorgiosironi: Run out of space since it's a 1.39GB image. We need to put the `/var/lib/docker` folder on the external volume, however here that volume has a special name `/bot-tmp` rather than `/ext`.

tests/provider/test_letterparser_provider.py
gnott: We could probably avoid potential complications of Docker-in-Docker if for a future dockerised `worker.py` includes `pandoc` executable in the image, then we don't need to use Docker to call pandoc.

giorgiosironi: My assumption is that, given, the 1+ GB size of this image, there is a lot of work involved in setting up `pandoc` on a host or in a Docker image. Also a microservice solution where `pandoc` runs behind an HTTP API could be investigated. In any case, all speculation so don't need to think about it now.

gnott: This was the first docker image I started using, and recently it looks like there is a more official pandoc docker image https://hub.docker.com/r/pandoc/core. With some fiddling I am getting the library's tests to use `pandoc/core:2.6` which is thankfully 52.6MB, far smaller than the other image. I will move to switch over the `decision-letter-parser` library to it, and then change the version used here in this PR.

provider/letterparser_provider.py
gnott: Space required is now down to 50+ MB using the alternate docker image. All green! I hope it wasn't too much work mapping the `/ext` which might not be required now the image is a more managable size?

giorgiosironi: This is not the first Docker image being used (also `strip-coverletter` has one plus a few that aid in testing like `sftp`); we might want to get that in nevertheless to make the Docker setup future-proof.

tests/provider/test_letterparser_provider.py
gnott: I was looking around other pandoc resources, and on their pandoc extras page is Pandoc as a service project which might be useful in the future https://github.com/mrded/pandoc-as-a-service

gnott:
Thanks for the approval, but my plan now is I think I will close this PR, change this branch a bit, and open a new PR for merging the very basic letterparser provider.

PR #981, opened 2019-11-20
Schedule crossref peer review

gnott:
Re: recent discussion on https://github.com/elifesciences/issues/issues/5223
Re: older ticket for overall plan https://github.com/elifesciences/elife-crossref-feed/issues/123

I chose to add a new activity named `ScheduleCrossrefPeerReview` for `"scheduling"` articles for peer review deposits because the logical checks are unique from other activities we have of this type. It will only attempt to schedule it if it is not part of a silent correction and the XML contains a `<sub-article>` tag, or it is a silent correction but only the most recent version of a silent correction.

In the cron, the `DepositCrossrefPeerReview` workflow is scheduled for each hour on the half hour.

Fixed a bug in the tests for the activity `ScheduleCrossref` where if you ran the full test suite, the override of a session variable would be retained for some reason. Instead it makes a `copy.copy()` of the session before changing the `run_type` on it.

I can go into more detail on this PR as required. It can perhaps serve as one example of PRs for review in our latest plan for developer team collaborations.

### DIFF
--- a/activity/activity_ScheduleCrossrefPeerReview.py
+++ b/activity/activity_ScheduleCrossrefPeerReview.py
@@ -59,16 +59,16 @@ class activity_ScheduleCrossrefPeerReview(Activity):
                 self.emit_monitor_event(
                     self.settings, article_id, version, run, self.pretty_name, "end", log_message)
                 return True
-        else:
-            # check article XML has <sub-article> tag before queuing it
-            if self.xml_sub_article_exists(expanded_folder_name) is False:
-                log_message = (
-                    '%s finds version %s of %s has no sub-article for peer review depositing' %
-                    (self.name, version, article_id))
-                self.logger.info(log_message)
-                self.emit_monitor_event(
-                    self.settings, article_id, version, run, self.pretty_name, "end", log_message)
-                return True
+
+        # check article XML has <sub-article> tag before queuing it
+        if self.xml_sub_article_exists(expanded_folder_name) is False:
+            log_message = (
+                '%s finds version %s of %s has no sub-article for peer review depositing' %
+                (self.name, version, article_id))
+            self.logger.info(log_message)
+            self.emit_monitor_event(
+                self.settings, article_id, version, run, self.pretty_name, "end", log_message)
+            return True
 
         try:
             xml_file_name = lax_provider.get_xml_file_name(
--- a/activity/activity_ScheduleCrossrefPeerReview.py
+++ b/activity/activity_ScheduleCrossrefPeerReview.py
@@ -61,7 +61,7 @@ class activity_ScheduleCrossrefPeerReview(Activity):
                 return True
         else:
             # check article XML has <sub-article> tag before queuing it
-            if not self.xml_sub_article_exists(expanded_folder_name):
+            if self.xml_sub_article_exists(expanded_folder_name) is False:
                 log_message = (
                     '%s finds version %s of %s has no sub-article for peer review depositing' %
                     (self.name, version, article_id))
--- a/activity/activity_ScheduleCrossrefPeerReview.py
+++ b/activity/activity_ScheduleCrossrefPeerReview.py
@@ -52,8 +52,8 @@ class activity_ScheduleCrossrefPeerReview(Activity):
             if str(version) != str(highest_version):
                 log_message = (
                     ('%s will not deposit article %s' +
-                    ' ingested by silent-correction, its version of %s does not equal the' +
-                    ' highest version which is %s') %
+                     ' ingested by silent-correction, its version of %s does not equal the' +
+                     ' highest version which is %s') %
                     (self.name, article_id, version, highest_version))
                 self.logger.info(log_message)
                 self.emit_monitor_event(
--- a/activity/activity_ScheduleCrossrefPeerReview.py
+++ b/activity/activity_ScheduleCrossrefPeerReview.py
@@ -1,6 +1,6 @@
 import json
 
-from provider import lax_provider
+from provider import crossref, lax_provider
 from provider.article_processing import download_jats
 from provider.storage_provider import storage_context
 from provider.execution_context import get_session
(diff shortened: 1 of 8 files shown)


activity/activity_ScheduleCrossrefPeerReview.py
```diff
@@ -0,0 +1,119 @@
...
+            return False
+
+        return True
+
+    def xml_sub_article_exists(self, expanded_folder_name):
+        jats_file = download_jats(
+            self.settings, expanded_folder_name, self.get_tmp_dir(), self.logger)
+        with open(jats_file, 'r') as open_file:
+            article_xml = open_file.read()
+            if '<sub-article' in article_xml:
```
giorgiosironi: Substrings check on XML documents are too low level, this needs to be parsed to check the tag name?
(e.g. `<sub-articletitle>` would pass the check but you don't want that one to pass it)

giorgiosironi: It's probably an XPath given this tag could be somewhere in the document?

activity/activity_ScheduleCrossrefPeerReview.py
```diff
@@ -0,0 +1,119 @@
...
+        if run_type == "silent-correction":
+            highest_version = lax_provider.article_highest_version(article_id, self.settings)
+            if str(version) != str(highest_version):
+                log_message = (
+                    ('%s will not deposit article %s' +
+                     ' ingested by silent-correction, its version of %s does not equal the' +
+                     ' highest version which is %s') %
+                    (self.name, article_id, version, highest_version))
+                self.logger.info(log_message)
+                self.emit_monitor_event(
```
giorgiosironi: these `emit_monitor_event` calls always add a lot of boilerplate that hides the important logic in the activity. What about a [`contextmanager`](https://docs.python.org/2/library/contextlib.html):
```
with self.monitoring_events(article_id, version, run):
     // logic
```
with the `contextmanager` wrapping this in `try`/`except`, performing logging and emitting events?

I know this is a separate change, but it would greatly simplify the code here, which means shorter reviews.

giorgiosironi: e.g. can be a separate PR that is merged before this one, and this one is updated to shake off the code that is not needed anymore

workflow/workflow_SilentCorrectionsProcess.py
```diff
@@ -68,6 +68,17 @@ def __init__(self, settings, logger, conn=None, token=None, decision=None,
                         "schedule_to_start_timeout": 300,
                         "start_to_close_timeout": 60 * 5
                     },
+                    {
```
giorgiosironi: proposal for an helper function:
```
def define_activity(
    activity_name, # used for the `activity_*` arguments
    input,            
    version=1,
    control=None,
    heartbeat_timeout=60 * 5,
    schedule_to_close_timeout=60 * 5,
    schedule_to_start_timeout=300,
    start_to_close_timeout=60 * 5
):
    return {
        "activity_type": activity_name,
        "activity_id": activity_name,
        "version": version,
        "input": input,
        "control": control,
        "heartbeat_timeout"=heartbeat_timeout,
        "schedule_to_close_timeout"=schedule_to_close_timeout,
        "schedule_to_start_timeout"=schedule_to_start_timeout,
        "start_to_close_timeout"=start_to_close_timeout,
    }
```
The call becomes:
```
    define_activity("ScheduleCrossrefPeerReview, input=data)
```
but leaves open the possibility to tune the defaults, such as the timeouts.

Would do here, but only apply to this activity definition and refactor the rest in a follow-up PR.

activity/activity_ScheduleCrossrefPeerReview.py
gnott: I was looking for the simplest way to keep junk out of the outbox. These files are later further filtered. I can change this to use the sub-article verification logic used in the `DepositCrossrefPeerReview` activity, which leverages the XML parsing libraries.

workflow/workflow_SilentCorrectionsProcess.py
gnott: These workflow definitions have not changed in six years, and it was always an intent to turn them into something. It turns out most of the time default timeouts are used, and rarely they are not. If we do this it will be more easy to see which activities in a workflow use different values. We can write up a new issue and subsequent code to PR.

activity/activity_ScheduleCrossrefPeerReview.py
gnott: Good idea, but I'd have to think about it more to comment further. 

If I were to improve this and many of the bot activities, I would be trying to encapsulate each part of the procedural steps in `do_activity()` into separate methods, which can sometimes be moved outside of the class itself with some refactoring, which more clearly demonstrate what purpose they serve.

I find this strategy makes it easier to write test cases for the activity without having to mock eight different providers, and you can more easily create edge case tests for each part.

activity/activity_ScheduleCrossrefPeerReview.py
gnott: Changed in commit https://github.com/elifesciences/elife-bot/pull/981/commits/e6028263f797e467d09b5524f3aa8079e4cf9832, uses existing parsing function from the `crossref.py` provider.

activity/activity_ScheduleCrossrefPeerReview.py
gnott: I added this to an issue in this repo `https://github.com/elifesciences/elife-bot/issues/982`

workflow/workflow_SilentCorrectionsProcess.py
gnott: I think I understand the general idea, some points for investigation on a new issue https://github.com/elifesciences/elife-bot/issues/983  

It will probably need a closer look based on what I remember about various workflows, such as the `activity_id` value can have significance (where we `VerifyGlencoe` more than once in a workflow, each activity needs to have a unique `activity_id` from each other).

Each activity also has defaults set on each, which may serve as a default value to use when returning, if instantiating the activity class to get these values when running each step of a workflow does not cause any problems.

gnott:
One code improvement, and I opened two new issues for future refactoring opportunities. Are there further comments before we attempt to try this on `end2end` and possible `continuumtest` environment?

gnott:
I am merging now to see if it can be deployed. I can fix or rollback anything that looks to be not working today.

PR #986, opened 2019-11-22
Decision letter parsing via letterparser_provider

gnott:
Very similar to draft PR https://github.com/elifesciences/elife-bot/pull/973 that was green and previously approved.

The differences here are
- updated to use the latest `letterparser` library sha, which now specifies a default Docker image name for calling pandoc
- changed the `letterparser_provider.py` to call `parse_file()` instead of the more specific call that just invoked the docker method of pandoc
- changed the test to pass a blank config dict, which will cause the default Docker image to be used

This is more generic than the last, which allows the local `pandoc` executable to be used if present, otherwise Docker is used.

(to note, I am undecided yet whether you should be able to explicity call pandoc via Docker even if you _do_ have `pandoc` available, but that situation hasn't yet presented itself and I think it is too early to think that far ahead).


provider/letterparser_provider.py
```diff
@@ -0,0 +1,16 @@
...
+
+IDENTITY = "process_%s" % os.getpid()
+LOGGER = log.logger("letterparser_provider.log", 'INFO', IDENTITY, loggerName=__name__)
+
+
+def parse_file(file_name, config):
+    try:
+        return parse.parse_file(file_name, config)
+    except docker.errors.APIError:
+        LOGGER.info('Error connecting to docker')
```
giorgiosironi: if this happens nothing is returned? Add a `raise`?

giorgiosironi: (that will re-raise the original exception)

giorgiosironi: (this is what I mean: https://stackoverflow.com/a/24623518/91590)

gnott: In the end it will return `None`, and with no parsed data, nothing will be produced. I am not against `raise` here, but it will eventually be handled in a graceful way. I think any non-Docker exceptions would be raised here. The robustness of this function may be enhanced later too as the workflow logic is built.

I will `raise` for now so we can potentially get this PR merged and create new PRs.

giorgiosironi: I can't say much without seeing the caller. Would it have to check the return value with an `if`?

gnott: There's a different option to parse a `.docx` in the `letterparser` library that is basicaly a one-step call. It illustrates how I should probably check how that would respond to bad data or exceptions in the library itself. I think `parse_file()` here in this provider would possibly not be used _per se_, maybe for debugging purposes or lower level operations if chosen.

giorgiosironi: What isn't clear is where the data go if there is no return value and no output arguments?

gnott: I remember in the past sometimes linters complain when there is a return statement that is not required, but it doesn't complain if I add `return None` to this one. I will add that too.

gnott: In latest commit https://github.com/elifesciences/elife-bot/pull/986/commits/2799288eb87cb8dd728c68dac955ab1101c3c2af I hope it looks better. It raises the specific docker exception, and has tests for that and for if a different exception is raised when trying to parse.

It looks like nothing is returned if an exception if raised.

gnott:
This PR, based on an earlier approved one, is a small amount of code, and has gone through a round of review and revision. I'm not in a hurry to merge it, but it is blocking further development of this workflow, so I will merge it now, and we can revise it later as required.

PR #987, opened 2019-12-04
Letterparser config

gnott:
Re issue https://github.com/elifesciences/issues/issues/5233 to use pandoc docker image.

Following a similar pattern to how this project loads the `crossref.cfg` config file, the file name and config section is specified in `settings.py`, and that is loaded by the provider.

Also updated to use the latest `letterparser` and `elifearticle` libraries, which are not specifically related to loading a config, but newer is better for future development of the activities and workflow.

I hope it is ok I add you as a reivewer @giorgiosironi, if you choose to review.

### DIFF
--- a/settings-example.py
+++ b/settings-example.py
@@ -159,6 +159,10 @@ class exp():
     packagepoa_config_file = 'packagepoa.cfg'
     packagepoa_config_section = 'elife'
 
+    # Decision letter parser
+    letterparser_config_file = 'letterparser.cfg'
+    letterparser_config_section = 'elife'
+
     # PubMed FTP settings
     PUBMED_FTP_URI = ""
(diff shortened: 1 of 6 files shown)


settings-example.py
```diff
@@ -159,6 +159,10 @@ class exp():
     packagepoa_config_file = 'packagepoa.cfg'
     packagepoa_config_section = 'elife'
 
+    # Decision letter parser
+    letterparser_config_file = 'letterparser.cfg'
```
giorgiosironi: I like that we specify explicitly the file name/path, no need to hardcode it

gnott:
Thanks @giorgiosironi, I'll merge now and continue building the workflow logic in segments.

Issue #983, opened 2019-11-20
Workflow definition refactoring and default values


gnott:
Suggested in comment https://github.com/elifesciences/elife-bot/pull/981#discussion_r348386220

The bulk of a workflow definition is a long list of JSON values, one for each activity along with values for various SWF timeout and workflow values. Often the values used are the defaults: five minutes for timeouts and some other values that rarely change.

These could be refactored to reduce the amount of code in each workflow definition, and making it easier to see when the default values are overridden.

- [x] Add to the base `Workflow` object some logic for workflow definition defaults
- [x] ~Consider to use the default values as specified in the particular activity objects as the default values for timeouts (if not otherwise specified in a workflow definition)~ - Answer: not going to do this for now, because workflows can independently set the timeout values, and they may not always match what the default timeout values are specified on the activity itself
- [x] Establish a clear and simple way to override values, when needed, for example in `VerifyGlencoe` the activity has a timeout of 5 minutes, but in the `IngestArticleZip` workflow we give it 15 minutes
- [x] To note: in SWF you must define a default value in order to register an activity, but you can then specify different values for that activity when it is run as part of a workflow execution.

gnott:
The majority of tidying up is done, and I will close this ticket.

PR #990, opened 2019-12-14
Ingest decision letter workflow, starter, S3 notification

gnott:
A start on issue https://github.com/elifesciences/elife-bot/issues/965, this addresses the first five checkboxes.

To build out the workflow in small bites, this PR defines the workflow, its starter, and starts a workflow execution after an S3 notification comes into the message queue.

The workflow steps right now only includes `PingWorker`; the real activities to this workflow will be added later. The ping should mean, if all is correct here, that it will cause no trouble, and also do very little at this time.

I also started on issue https://github.com/elifesciences/elife-bot/issues/983 here too, where `workflow/helper.py` module contains a function to define a workflow step, using default values where possible. In future all the workflows can be refactored to use this, if the example here is good.

It was informative to look through the old workflow logic a little, where I noticed the `control` value for workflow activity steps is not used at all. I left it in as a possible value to customise, even though we haven't used it yet, and probably never will. We could possibly remove `control` all together as a separate PR, if this simplification is a good idea (more detail you can see on this line https://github.com/elifesciences/elife-bot/blob/develop/workflow/objects.py#L170 `'control data' is hardcoded in the `boto` call, and the value of `control` is not used).

In summary, this PR represents the basics to add a new workflow, ability to start that workflow, and to start a workflow from an S3 notification from a particularly named bucket. The actual effect of this workflow should be a `ping` and nothing more. This will allow adding activities one-by-one to this workflow and we can review and approve them each separately.

### DIFF
--- a/queue_workflow_starter.py
+++ b/queue_workflow_starter.py
@@ -86,10 +86,15 @@ def process_data_ingestarticlezip(workflow_data):
     return data
 
 
-def process_data_initialarticlezip(workflow_data):
+def process_data_s3_notification_default(workflow_data):
+    """definition for default data in response to an S3 notification"""
     data = {'info': S3NotificationInfo.from_dict(workflow_data),
             'run': str(uuid.uuid4())}
     return data
+
+
+def process_data_initialarticlezip(workflow_data):
+    return process_data_s3_notification_default(workflow_data)
 
 
 def process_data_postperfectpublication(workflow_data):
@@ -103,9 +108,11 @@ def process_data_pubmedarticledeposit(workflow_data):
 
 
 def process_data_ingestdigest(workflow_data):
-    data = {'info': S3NotificationInfo.from_dict(workflow_data),
-            'run': str(uuid.uuid4())}
-    return data
+    return process_data_s3_notification_default(workflow_data)
+
+
+def process_data_ingestdecisionletter(workflow_data):
+    return process_data_s3_notification_default(workflow_data)
 
 
 workflow_data_processors = {
@@ -114,7 +121,8 @@ workflow_data_processors = {
     'SilentCorrectionsIngest': process_data_initialarticlezip,
     'PostPerfectPublication': process_data_postperfectpublication,
     'PubmedArticleDeposit': process_data_pubmedarticledeposit,
-    'IngestDigest': process_data_ingestdigest
+    'IngestDigest': process_data_ingestdigest,
+    'IngestDecisionLetter': process_data_ingestdecisionletter
 }
 
 
--- a/starter/starter_IngestDecisionLetter.py
+++ b/starter/starter_IngestDecisionLetter.py
@@ -1,0 +1,78 @@
+import os
+# Add parent directory for imports
+parentdir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+os.sys.path.insert(0, parentdir)
+
+import boto.swf
+import json
+from optparse import OptionParser
+from S3utility.s3_notification_info import S3NotificationInfo
+import starter.starter_helper as helper
+from starter.starter_helper import NullRequiredDataException
+
+
+class starter_IngestDecisionLetter():
+    def __init__(self):
+        self.const_name = "IngestDecisionLetter"
+
+    def start(self, settings, run, info):
+
+        # Log
+        logger = helper.get_starter_logger(settings.setLevel, helper.get_starter_identity(self.const_name))
+
+        if hasattr(info, 'file_name') is False or info.file_name is None:
+            raise NullRequiredDataException("filename is Null. Did not get a filename.")
+
+        input_data = S3NotificationInfo.to_dict(info)
+        input_data['run'] = run
+
+        workflow_id, \
+        workflow_name, \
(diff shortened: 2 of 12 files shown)


queue_workflow_starter.py
```diff
@@ -103,9 +108,11 @@ def process_data_pubmedarticledeposit(workflow_data):
...
 
 def process_data_ingestdigest(workflow_data):
-    data = {'info': S3NotificationInfo.from_dict(workflow_data),
-            'run': str(uuid.uuid4())}
-    return data
+    return process_data_s3_notification_default(workflow_data)
+
+
+def process_data_ingestdecisionletter(workflow_data):
+    return process_data_s3_notification_default(workflow_data)
```
giorgiosironi: good removal of duplication

queue_workflow_starter.py
```diff
@@ -114,7 +121,8 @@ def process_data_ingestdigest(workflow_data):
     'SilentCorrectionsIngest': process_data_initialarticlezip,
     'PostPerfectPublication': process_data_postperfectpublication,
     'PubmedArticleDeposit': process_data_pubmedarticledeposit,
-    'IngestDigest': process_data_ingestdigest
+    'IngestDigest': process_data_ingestdigest,
+    'IngestDecisionLetter': process_data_ingestdecisionletter
```
giorgiosironi: I think you can add a `,` at the end of this line to make diffs shorter in the future

starter/starter_IngestDecisionLetter.py
```diff
@@ -0,0 +1,78 @@
+import os
+# Add parent directory for imports
+parentdir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
```
giorgiosironi: this is weird; is it required if execute `venv/bin/python starter/starter_IngestDecisionLetter.py`?

starter/starter_IngestDecisionLetter.py
```diff
@@ -0,0 +1,78 @@
...
+
+    import settings as settingsLib
+    SETTINGS = settingsLib.get_settings(ENV)
+
+    STARTER_OBJECT = starter_IngestIngestDecisionLetter()
+
+    # note: this starter must be started by an S3Notification and not directly from command line
+    RUN = None
+    INFO = None
+    STARTER_OBJECT.start(settings=SETTINGS, run=RUN, info=INFO)
```
giorgiosironi: there's a lot of boilerplate in here so surprised there is nothing to extract that is shared with other starters?

queue_workflow_starter.py
gnott: I've noticed having a `,` at the end of a list doesn't harm anything, and it's also not required, so I wasn't sure what is the preferred style. I like adding the `,` as you suggest for better diffs and it works for every line.

starter/starter_IngestDecisionLetter.py
gnott: Last time I tried to remove these it resulted in an error invoking the starter. I cannot remember if it was due to a missing `__init__.py` file from the root folder, how the worker runs, or something to do with the NewRelic wrapper, or how the service daemon was invoked.

This hack is present on most, if not all, of the starters right now. 

I suggest if we want to try out removing them again, we could try it on a separate PR, which we can merge and test through to the `end2end` testing to be sure. If it fails there we can revoke the changes and/or diagnose the exact reason why it fails (if it still does). We could create a new issue describing the plan to test / debug / revoke. Do you think this is a good idea @giorgiosironi ?

starter/starter_IngestDecisionLetter.py
gnott: Starters were at one time going to be reviewed and refactored. They are very similar but not all the same. I think some starter helpers were added recently. Also, some refactoring of the `if __name__ == "__main__":` parts I did so some in order to get higher test coverage. 

These I think also might be good to be described in a separate issue and completed as part of a separate PR. I thought I had an issue for this (which would have been from years ago) but it may have been closed or it was not exactly this particular topic of refactor.

gnott:
Ah, I see you approved this PR - thanks for reviewing @giorgiosironi! I will create two issues for the things I think could be done in separate PRs so those suggestions are carried forward and the momentum of continuous improvement can go on.

starter/starter_IngestDecisionLetter.py
gnott: https://github.com/elifesciences/elife-bot/issues/991 issue I created for this

starter/starter_IngestDecisionLetter.py
gnott: https://github.com/elifesciences/elife-bot/issues/992 issue I created for starter refactoring

gnott:
I added `sections.docx` to the continuumtest bucket for these files and it resulted in a completed `IngestDecisionLetter` workflow, where I conclude the code here is functioning correctly.

starter/starter_IngestDecisionLetter.py
giorgiosironi: Definitely an issue for a different PR, try to remove from the least important `starter` first.

PR #994, opened 2020-01-08
Workflow definitions refactoring

gnott:
Re issue https://github.com/elifesciences/elife-bot/issues/983

An example of definiting a workflow step was merged recently, used in the `IngestDecisionLetter` workflow.

I continued a similar pattern of refactoring for the remaining workflow definitions.

There were many with a standard of 10 or 15 minute timeouts in a specific format, for which I created `define_workflow_step_10()` and `define_workflow_step_15()` to achieve a cleaner result.

There are still some with odd or longer timeout values, which still have these specific timeout values.

Only one worklfow activity step, for `VerifyGlencoe`, do we need to explicitly override the `activity_id` value (as `VerifyGlencoeAgain`, when we repeat the same activity twice in both the regular ingest and the silient correction ingest).

I hope the diff on this is not too hard to read. I carefully copy and pasted items, and I also double-checked each before staging the changes, so I'm fairly sure this will not break anything.

### DIFF
--- a/workflow/helper.py
+++ b/workflow/helper.py
@@ -27,3 +27,53 @@ def define_workflow_step(
         ("schedule_to_start_timeout", schedule_to_start_timeout),
         ("start_to_close_timeout", start_to_close_timeout),
     ])
+
+
+def define_workflow_step_10(
+        activity_type,
+        activity_input,
+        activity_id=None,
+        version="1",
+        control=None,
+        heartbeat_timeout=60 * 10,
+        schedule_to_close_timeout=60 * 10,
+        schedule_to_start_timeout=60 * 5,
+        start_to_close_timeout=60 * 10):
+    """
+    Workflow step definition with 10 minute timeout defaults
+    """
+    return define_workflow_step(
+        activity_type=activity_type,
+        activity_input=activity_input,
+        activity_id=activity_id,
+        version=version,
+        control=control,
+        heartbeat_timeout=heartbeat_timeout,
+        schedule_to_close_timeout=schedule_to_close_timeout,
+        schedule_to_start_timeout=schedule_to_start_timeout,
+        start_to_close_timeout=start_to_close_timeout)
+
+
+def define_workflow_step_15(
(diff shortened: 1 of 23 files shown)


workflow/helper.py
```diff
@@ -27,3 +27,53 @@ def define_workflow_step(
         ("schedule_to_start_timeout", schedule_to_start_timeout),
         ("start_to_close_timeout", start_to_close_timeout),
     ])
+
+
+def define_workflow_step_10(
```
giorgiosironi: weird to have a number in the name, but the duplication is minimal so if it works for readability :+1:

workflow/workflow_ApproveArticlePublication.py
```diff
@@ -35,28 +32,8 @@ def __init__(self, settings, logger, conn=None, token=None, decision=None,
...
-                        "activity_id": "PublishToLax",
-                        "version": "1",
-                        "input": data,
-                        "control": None,
-                        "heartbeat_timeout": 60 * 5,
-                        "schedule_to_close_timeout": 60 * 5,
-                        "schedule_to_start_timeout": 300,
-                        "start_to_close_timeout": 60 * 5
-                    },
+                    define_workflow_step("PingWorker", data),
```
giorgiosironi: a lot of code removed :+1:

workflow/helper.py
giorgiosironi: maybe they could be `_short` and `_medium` rather than `_10` or `_15`, or some other qualitative difference if they are tuned to 5 and 20 minutes later on

gnott: I didn't like the strictness of the number in the name either. Thanks for the suggestion! I will change `_10` to `_short` and `_15` to `_medium` (the default one I guess is "very_short" at 5 minutes but will remain the same name).

gnott:
I also forgot to commit changes to `workflow_SilentCorrectionsIngest.py` in the previous commits, now included after renaming the define workflow step functions.

PR #998, opened 2020-01-13
Fix S3Monitor workflow activity version call.

gnott:
Bug fix to https://github.com/elifesciences/elife-bot/pull/994. The wrong activity version value was used for `S3Monitor`.


gnott:
Re issue https://github.com/elifesciences/issues/issues/5303

Issue #999, opened 2020-01-14
Reduce amount of logging to decider.log


gnott:
Observed as part of issue https://github.com/elifesciences/issues/issues/5303

When a workflow retries an activity over and over again, the `decider.log` log file size can increase fairly quickly due to the amount of logging done there.

Planned initial tasks

- [x] Stop logging workflow history to the `decider.log` by default
- [x] ~~Look for additional sources of logging that provides little value in a normal, non-debug-mode situation~~ should be fine, I think the `events` in the decision dict is the culprit

Issue #1000, opened 2020-01-14
Refactor and simplify decider.py


gnott:
When investigating the logging done by `decider.py` in issue https://github.com/elifesciences/elife-bot/issues/999, I notice the code is pretty old, and the test coverage is poor

Some todo tasks:

- [x] Refactor and simplify the code, as is possible
- [x] Write more tests for smaller part of code -- currently not covered by automated tests -- and mock AWS connections
- [x] Lint the code

PR #1001, opened 2020-01-14
Do not log decision events in non-debug mode.

gnott:
Fixes https://github.com/elifesciences/elife-bot/issues/999

So we are not caught with large decider log files, this is a very small change proposed.

When processing a workflow decision, the `decider.py` gets the entire workflow event history from AWS SWF service by following all the paging tokens and puts this data into a dict's `events` value. For a stuck and repeatedly retried workflow, these can reach in the thousands of events, logged every 10 seconds.

In this PR, a separate copy of `decision` is made and its `events` value is set to a blank list `[]` if the decider is not in debug mode. `debug=False` is the default value, and currently there's no way to specify `debug=true`. 

I created the debug flag primarily to signify how there is the potentially to log more decision data, in the future, if we want to expand into that, but based on recent development of this project, we do not normally need to dig into that much detail to determine why a workflow is failing. If we changed how workflow histories are processed, then at that time we could turn on `debug` in some way to figure out why it cannot process a workflow correctly.

I'm open to suggestions to pass the `debug` value in from where the process is started (if you think so), or any other suggestions.

I created an additional issue to cover refactoring, linting, and more test coverage in the future (issue https://github.com/elifesciences/elife-bot/issues/1000).

### DIFF
--- a/decider.py
+++ b/decider.py
@@ -97,10 +97,10 @@ def decide(settings, flag, debug=False):
 
 
 def trimmed_decision(decision, debug=False):
-    """trim events from a copy of decision prior to logging if not debug"""
+    """trim data from a copy of decision prior to logging if not debug"""
     decision_trimmed = copy.copy(decision)
     if not debug:
-        # set events value to a blank list to reduce logging
+        # removed to limit verbosity
         decision_trimmed['events'] = []
     return decision_trimmed
 
--- a/decider.py
+++ b/decider.py
@@ -1,4 +1,5 @@
 import boto.swf
+import copy
 import log
 import json
 import random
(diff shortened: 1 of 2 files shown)


decider.py
```diff
@@ -92,6 +95,16 @@ def decide(settings, flag):
 
     logger.info("graceful shutdown")
 
+
+def trimmed_decision(decision, debug=False):
+    """trim events from a copy of decision prior to logging if not debug"""
+    decision_trimmed = copy.copy(decision)
+    if not debug:
+        # set events value to a blank list to reduce logging
+        decision_trimmed['events'] = []
```
giorgiosironi: could put it a conventional value to make it clear this has been removed, like `['***removed to limit verbosity***']`

gnott: Much better wordings, thanks! Words were failing me yesterday I suppose.

Issue #966, opened 2019-10-29
ValidateDecisionLetterInput activity


gnott:
From issue https://github.com/elifesciences/elife-bot/issues/965

The purpose of this activity in summary will be to download a `.zip` file from an S3 bucket, unzip it, and check its validity. It may also save some values to a session for later use.

Tasks may include

- [x] Check the `.zip` file file name is the correct format, and an article ID can be gotten from it
- [x] Check the zip contains a `.docx` file
- [x] ~~Look at other (optional) assets in the zip file (I'm not sure what can be checked at this point without counting how many files should exist based on the `.docx` file contents)~~ will postpone until later when the workflow is used more
- [x] Try to parse the `.docx` file using `pandoc` and detect if there are any errors in producing JATS output
- [x] Check it can find a `decision letter` and `author response` sections from the `.docx` file (note: currently just counting there is at least two articles)
- [x] If there are any validity errors, send an email to the designated recipients including details about the invalid parts

PR #1002, opened 2020-01-14
New download_helper, refactor from digest provider.

gnott:
I was starting in on logic for decision letter parsing (https://github.com/elifesciences/elife-bot/issues/966), based on how digest workflows work.

There was code in the `digest_provider.py` for downloading the zip files from the S3 bucket that could be reused.

Here I moved it over to a new module, `download_helper.py`, and refactored all the existing digest activities to use it.

Rather than mixing this into a PR with new decision letter workflow code, I created this smaller PR first.

### DIFF
--- a/provider/digest_provider.py
+++ b/provider/digest_provider.py
@@ -9,6 +9,7 @@ from docx.opc.exceptions import PackageNotFoundError
 from digestparser import jats, build, conf
 import provider.utils as utils
 from provider.storage_provider import storage_context
+from provider.download_helper import download_file
 import provider.lax_provider as lax_provider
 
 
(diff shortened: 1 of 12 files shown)


provider/digest_provider.py
```diff
@@ -148,32 +140,6 @@ def image_file_name_from_s3(settings, article_id, bucket_name):
...
-    "download the digest filename from a bucket or storage to the to_dir"
-    if not resource_origin:
-        return None
-    filename_plus_path = to_dir + os.sep + filename
-    with open(filename_plus_path, 'wb') as open_file:
-        storage.get_resource_to_file(resource_origin, open_file)
-    return filename_plus_path
-
-
-def download_digest_from_s3(settings, filename, bucket_name, bucket_folder, to_dir):
```
giorgiosironi: simply dead code, I imagine

giorgiosironi: (now that the generic version exists)

gnott: Some slight renaming to make these functions more generic sounding, and then they are essentially exact replacements for these old ones with `digest` in their name. So now instead of `download_digest()` using `download_helper.download_file()`, and instead of `download_digest_from_s3()` using `download_helper.download_file_from_s3()`.

PR #1003, opened 2020-01-16
Validate decision letter

gnott:
Fixes https://github.com/elifesciences/elife-bot/issues/966

Merging blocked by https://github.com/elifesciences/builder-private/pull/346 to be merged first to provide settings.

Major parts of this PR

- New activity named `ValidateDecisionLetterInput`, which takes a zip file deposited into a bucket, parses it using the `letterparser` library, and checks for any errors or missing data we need
- `provider/letterparser_provider.py` is expanded to interface with the `letterparser` library, handle exceptions and return the status of each step
- comprehensive tests for the new activity and the `validate_articles()` function in the provider
- added the new activity as a step to the `IngestDecisionLetter` workflow

I think in the near future some of the steps to transform a zip file to the XML output may be refactored so it can be reused, as appropriate, in other activities, except making sure the activity that validates the input can keep track of each step to monitor for errors was the goal at this time.

In future, I think the additional asset files, such as figure `.jpg` or `.tif` files, in the zip file will need to be checked at this early step too, but I hope to add that when it becomes apparent how to do this effectively. For example, possibly checking the XML for `<graphic>` or `<media>` tags and then counting and / or comparing the names with the asset files in the zip file supplied, and log and error if they mismatch. I prefer to keep moving ahead to build all the planned workflow steps in order and then improve them as later revisions.


activity/activity_ValidateDecisionLetterInput.py
```diff
@@ -0,0 +1,141 @@
...
+            # create the email
+            email_message = email_provider.message(subject, sender_email, recipient)
+            email_provider.add_text(email_message, body)
+            # send the email
+            email_provider.smtp_send(connection, sender_email, recipient,
+                                     email_message, self.logger)
+        return True
+
+
+def error_email_subject(filename):
```
giorgiosironi: no clear templating system to use here? even just a separate file whose content can be read into a string to call `.format()` on?

giorgiosironi: (not for the subject, for the body of course)

provider/letterparser_provider.py
```diff
@@ -23,3 +24,69 @@ def parse_file(file_name, config):
...
+
+    # check for any articles at all
+    if not articles:
+        valid = False
+        error_message = 'No articles to check'
+        error_messages.append('No articles to check')
+        logger.info(error_message)
+
+    # check for two article objects
+    min_count = 2
```
giorgiosironi: magic number is suspicious; should be a visible constant at the top of the file?

gnott: The number is certainly magic, and I agree it deserves to be a constant, good catch!

I think I may expand this into checking the title values of the articles to see there is at least one of each `Decision letter` and `Author response` too.

This is, of course, a place we can expand on the strict validation before continuing with the workflow steps.

gnott: I'm adding another constant named `ARTICLE_TITLE_MAP` which holds the article title match snippet and count of each type expected.

activity/activity_ValidateDecisionLetterInput.py
gnott: Another good observation. Even though the body is very short, I am moving into `email_provider.py` following a similar string Template safe_substitute() logic that is already in use (which is hopefully acceptable still).

gnott: Added a new issue for myself https://github.com/elifesciences/elife-bot/issues/1004 to reduce another one or two of these error email body functions

Issue #996, opened 2020-01-10
Refactor to remove old Python 2.7 related code


gnott:
- [x] Refactor and clean-up code if there is Python 2 specific logic that can be removed
- [x] Look for any additional changes to fully drop support for Python 2.7

PR #1005, opened 2020-01-17
More py27

gnott:
Fixes https://github.com/elifesciences/elife-bot/issues/996

This PR includes what I hoped to change in a more in-depth, complicated, removal of additiona Python 2.7 logic. It may not be complete but it is the things I could find as a result of some keyword searching.

It includes:

- in `provider/utils.py` renamed `unicode_decode()` to `bytes_decode()`, which I think is a more accurate name now
- simplfied the unicode encode / decode logic in `provider/utils.py`
- consistent use of `urllib.parse`, including `urllib.parse.unquote_plus()` and `urllib.parse.urlencode()`
- reading binary files in code and test scenarios consistently the same way
- in email digests tests, able to change it to compare the file name and email attachment string, without having to also support what Python 2 was generating
- can use `assertCountEqual()` in `test_glencoe_check.py`

That's about it, give or take.

This is an optional follow-up issue I mentioned on https://github.com/elifesciences/issues/issues/4726

### DIFF
--- a/activity/activity_EmailDigest.py
+++ b/activity/activity_EmailDigest.py
@@ -91,7 +91,7 @@ class activity_EmailDigest(Activity):
         return self.ACTIVITY_PERMANENT_FAILURE
 
     def output_path(self, output_dir, file_name):
-        """for python 3 should always return bytes which is what we want"""
+        """for python 3 cast file_name to str so it can be joined with the path value"""
         return os.path.join(output_dir, str(file_name))
 
     def generate_output(self, digest_content):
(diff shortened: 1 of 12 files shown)


activity/activity_EmailDigest.py
```diff
@@ -91,11 +91,8 @@ def do_activity(self, data=None):
         return self.ACTIVITY_PERMANENT_FAILURE
 
     def output_path(self, output_dir, file_name):
-        "for python 2 and 3 support, only encode sometimes"
-        try:
-            return os.path.join(output_dir, str(file_name)).encode('utf8')
-        except AttributeError:
-            return os.path.join(output_dir, str(file_name))
+        """for python 3 should always return bytes which is what we want"""
```
giorgiosironi: should always return `str`, not bytes?

gnott: Yes, you're right. I got a little turned around. I think the `file_name` may be bytes, especially if unicode characters are involved, but then it is cast to `str` so it can relabily be concatenated with the path value.

Issue #1004, opened 2020-01-16
Use email_provider.error_email_body() in ValidateDigestInput activity


gnott:
Planning to merge a template and function in PR https://github.com/elifesciences/elife-bot/pull/1003. 

- [x] When `error_email_body()` is merged it can be used in the `ValidateDigestInput` activity to reduce duplicate code.
- [x] May be able to use something like it in the `PostDigestJATS` activity, which includes more content in the body though (yes, for sending success emails)
- [x] On closer review now, it looks like renaming the function to `simple_email_body()` will be more general purpose for both success and error emails, and to accept varied email body content with too much duplicate code

PR #1006, opened 2020-01-17
Email body refactoring for sending simple error and success emails

gnott:
Fixes https://github.com/elifesciences/elife-bot/issues/1004

After starting in on this refactoring issue, I noticed with a small change the previously named `error_email_body()` in `provider/email_provider.py` would be better renamed as `simple_email_body()`, which I've done here. As a result, it is more suited to formatting both simple success and error emails.

Similar code in other activities is refactored to use `simple_email_body()`. If the email body content contains more than error messages, the body content is formatted separately, but now using fewer lines of code and is more specific to the particular activity data.

To note / a question: when searching for places to refactor, there are a few other email body functions around that we may want to also turn into templates. Examples are in `PackagePoA`, `PublishFinalPOA`, `PubmedArticleDeposit`, `PubRouterDeposit`, and a little in `PublicationEmail`. These would possibly be better as jinja2 templates, but it may be possible to just use string replace templates. @giorgiosironi what do you think in terms of priority whether all the email body formatting were in templates, would this be worthwhile for now or in the future?

### DIFF
--- a/activity/activity_EmailDigest.py
+++ b/activity/activity_EmailDigest.py
@@ -118,7 +118,7 @@ class activity_EmailDigest(Activity):
         "email the digest as an attachment to the recipients"
         success = True
 
-        datetime_string = time.strftime('%Y-%m-%d %H:%M', time.gmtime())
+        datetime_string = time.strftime('%Y-%m-%dT%H:%M:%S.000Z', time.gmtime())
         body = email_provider.simple_email_body(datetime_string)
         subject = success_email_subject(digest_content)
         sender_email = self.settings.digest_sender_email
--- a/activity/activity_EmailDigest.py
+++ b/activity/activity_EmailDigest.py
@@ -4,7 +4,7 @@ import time
 from digestparser import output
 import digestparser.utils as digest_utils
 from S3utility.s3_notification_info import parse_activity_data
-from provider import digest_provider, download_helper, email_provider
+from provider import digest_provider, download_helper, email_provider, utils
 from activity.objects import Activity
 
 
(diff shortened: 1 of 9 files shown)


giorgiosironi:
> These would possibly be better as jinja2 templates

Do the templates contain conditionals or loops? That would be the Godzilla threshold that would call for Jinja or a similar full templating language.

activity/activity_EmailDigest.py
```diff
@@ -118,8 +118,8 @@ def email_digest(self, digest_content, output_file):
         "email the digest as an attachment to the recipients"
         success = True
 
-        current_time = time.gmtime()
-        body = success_email_body(current_time)
+        datetime_string = time.strftime('%Y-%m-%dT%H:%M:%S.000Z', time.gmtime())
```
giorgiosironi: a constant on the base activity class could be useful to avoid `%Y-%m-%dT%H:%M:%S.000Z` spreading everywhere?

giorgiosironi: or maybe it just needs a `datetime_provider` module?

gnott:
> Do the templates contain conditionals or loops? That would be the Godzilla threshold that would call for Jinja or a similar full templating language.

I think this would not be known for sure until it was implemented. I'd estimate the complexity of the logic is low, and it might be possible to offload the conditional logic to the programming code itself.

This project already uses `jinja2` templates for sending `PublicationEmail` emails to authors, and it wouldn't be a big leap to also use it for internal and administator emails.

I'll open a new issue regarding this where it can be discussed and prioritised, or rejected.

activity/activity_EmailDigest.py
gnott: Good point again! There are a couple date formatting constants already in `provider/utils.py`, but this variant is not there yet. I'll add a third constant for this one.

gnott: I'm naming the new const as `DATE_TIME_FORMAT`. I've only applied it to the files already edited in this PR, and I'll create a new issue where this new value can be used throughout, because there's more than a dozen places this date time formatting string is used in this project.

gnott:
Thanks for approving @giorgiosironi - I'll merge this when tests on the most recent commit are shown to pass.

PR #1012, opened 2020-01-22
Decider refactoring

gnott:
Fixes https://github.com/elifesciences/elife-bot/issues/1000

The diff on this PR may be hard to read, if so -- sorry!

Steps I followed on refactoring of `decider.py`

- Linted the code in general
- Added test scenarios to cover as much of the code as possible (starting from about 40% coverage)
- Break apart the complicated `decide()` function into about three parts
- Further linting for variable naming
- Switched `OptionParser` usage over to `ArgumentParser`, moving the helper functions to `provider/utils.py`

I changed the `FakeLogger` testing object so it concatenates `loginfo` values instead of replacing the value each time the info log is written. I did this so I could assert what decider was writing to the log. I changed the tests that depended on the old `FakeLogger` so they check the log value `.endswith()` the string they expected in their assertions.

A note on the new functions in `provider/utils.py` - I copied these over from `cron.py`, where I recently changed it to use more testable and modern coding for running as main(). After this PR is merged, `cron.py` can be changed to use the now central helper functions (or I could do it in this PR too, but I thought that would be less clean).

I suspect if I messed up logically in `decider.py` we will know very quickly when `end2end` tests are run, since they'd have no way to finish without a functioning decider!


gnott:
Also noting:

- refactored away the use of `eval` when importing the workflow object
- code coverage for `decider.py` is now 97%

provider/utils.py
```diff
@@ -109,3 +110,18 @@ def get_doi_url(doi):
     Given a DOI, get the URL for the DOI
     """
     return "https://doi.org/%s" % doi
+
+
+def console_start_env():
```
giorgiosironi: hopefully these will be reusable in all sorts of scripts

tests/activity/classes_mock.py
```diff
@@ -237,23 +237,45 @@ def list(self, prefix='', delimiter='', headers=''):
...
+        if args:
+            try:
+                rendered_msg = msg % args
+            except TypeError:
+                # if args were supplied as a tuple
+                rendered_msg = msg % args[0]
+        else:
+            rendered_msg = str(msg)
+        # append message to the log
+        self.loginfo = '\n'.join([self.loginfo, rendered_msg])
```
giorgiosironi: if `loginfo` is turned into a list of strings rather than a single string, it can make checking easier e.g. `self.loginfo[-1]` for the last line

tests/provider/test_utils.py
```diff
@@ -108,5 +110,22 @@ def test_get_doi_url(self):
         self.assertEqual(doi_url, "https://doi.org/10.7554/eLife.08411")
 
 
+class TestConsoleStart(unittest.TestCase):
+
+    def test_console_start(self):
+        env = 'foo'
+        expected = env
+        testargs = ['cron.py', '-e', env]
+        with patch.object(sys, 'argv', testargs):
```
giorgiosironi: minimal patching :+1:

provider/utils.py
gnott: I originally added this because coverage reports show the `if __name__ == "__main__":` portion of code is excluded from coverage percentages, and I wanted to bump up the coverage, which of course is better for testing potential errors.

tests/activity/classes_mock.py
gnott: I set to follow the other properties (which are plain strings), and what would normally be in a log file, but I like your suggestion of a list format, plus we have the opportunity to do it now and update the tests to use the list method. I will make it so! Of course anything that wants to check the full log could just join the list or just `str(list)` it.

tests/provider/test_utils.py
gnott: These are also from the `tests/test_cron.py` version of it (which I will refactor to use the `utils.py` in a subsequent PR). I remember having a little difficulty trying to patch a system library, but I figured it out eventually.

gnott:
I'm also going to convert `tests/test_data/decision.json` to pretty printed JSON.

Issue #967, opened 2019-10-29
GenerateDecisionLetterJATS activity


gnott:
From issue #965

The `GenerateDecisionLetterJATS` activity can generate final JATS output, the input being a `.zip` file containing a `.docx` file and optionally some figure or video assets. One option is to save the generated JATS somewhere for reference later, and/or for the next activity which will be to POST it to an endpoint.

- [x] Download the input `.zip` from an S3 bucket, as required
- [x] Convert the `.docx` to rough JATS XML using `pandoc` via the `decision-letter-parser` code library
- [x] Build article objects by parsing the rough JATS (also via `decision-letter-parser`)
- [x] Generate final JATS XML from the article objects (via `decision-letter-parser`) - The two steps above may be done in one operation depending on any content customisation options that can be included in a two-step process)
- [x] Set the final name of asset files (which goes into the final JATS XML)
- [x] Save the final JATS output to an S3 bucket (for reviewing purposes)
- [x] Possibly store the S3 location of the JATS in a session if we want to use it without having to generate it again in subsequent activities
- [x] ~~If there is an error in generating JATS or renaming / locating figure assets then email the designated recipients about the errors~~ for now I think we can skip the emailing at this stage, since prior to this the Validate activity would send an email if XML could not be produced, and later when XML is sent to an API endpoint, it can email if there is an error at that point

PR #1018, opened 2020-01-25
Generate decision letter

gnott:
Fixes https://github.com/elifesciences/elife-bot/issues/967

Will require settings to be merged prior to this PR https://github.com/elifesciences/builder-private/pull/347

More functions are moved into the `letterparser_provider.py` so they can be used by multiple activities. As a result, in `activity/activity_ValidateDecisionLetterInput.py` the multiple steps to convert a `.zip` into decision letter output is split into two parts: zip -> article objects, and articles to XML.

`GenerateDecisionLetterJATS` is the new activity. It will download the `.zip`, generate XML, upload it to a folder in the decision letter output bucket, and store a couple values into a session for use later when this XML is posted to an API endpoint.

Folder name and file name values are formatted from strings in `settings.py`.

The new activity is added to the `IngestDecisionLetter` workflow so we can try it out.


activity/activity_GenerateDecisionLetterJATS.py
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        # Load the config
+        self.letterparser_config = letterparser_provider.letterparser_config(self.settings)
+
+    def do_activity(self, data=None):
+        """
+        Activity, do the work
+        """
+        if self.logger:
```
giorgiosironi: there should always be a `self.logger`; in tests, pass in a mock that just ignores all calls

activity/activity_ValidateDecisionLetterInput.py
```diff
@@ -63,25 +62,25 @@ def do_activity(self, data=None):
...
+            letterparser_provider.process_zip(
+                self.input_file,
+                config=self.letterparser_config,
+                temp_dir=self.directories.get("TEMP_DIR"),
+                logger=self.logger))
 
-        # Convert docx to articles
-        self.statuses["build"], self.articles = letterparser_provider.docx_to_articles(
-            docx_file_name, config=self.letterparser_config, logger=self.logger)
+        self.copy_statuses(statuses)
```
giorgiosironi: this is storing data in `self` so `set_` as a verb can make it more explicit?

activity/activity_ValidateDecisionLetterInput.py
```diff
@@ -63,25 +62,25 @@ def do_activity(self, data=None):
...
 
-        # Validate content of articles
-        self.statuses["valid"], error_messages = letterparser_provider.validate_articles(
-            self.articles, logger=self.logger)
+        # part 2 articles to XML
+        self.xml_string, statuses = letterparser_provider.process_articles_to_xml(
```
giorgiosironi: the logic is reasonably isolated from the activity, which helps with reuse and testability

activity/activity_GenerateDecisionLetterJATS.py
```diff
@@ -0,0 +1,129 @@
...
+        self.articles, asset_file_names, statuses, error_messages = (
+            letterparser_provider.process_zip(
+                self.input_file,
+                config=self.letterparser_config,
+                temp_dir=self.directories.get("TEMP_DIR"),
+                logger=self.logger))
+        # part 2 articles to XML
+        self.xml_string, statuses = letterparser_provider.process_articles_to_xml(
+            self.articles,
+            self.directories.get("TEMP_DIR"),
```
giorgiosironi: Just to check: is this just using the temporary directory to write new files into it, or relying on anything being present there? The latter would be a smell

activity/activity_GenerateDecisionLetterJATS.py
```diff
@@ -0,0 +1,129 @@
...
+            self.logger,
+            pretty=True,
+            indent="    ")
```
giorgiosironi: don't know what this `indent` parameter does but seems weird compared to highly structured input like the article :thinking:

activity/activity_GenerateDecisionLetterJATS.py
gnott: This directory name is passed along to the function in the `letterparser` library, where it is the location to unzip files from a `.zip` file. Outside of that it isn't used much.

activity/activity_GenerateDecisionLetterJATS.py
gnott: The `indent` value is passed to `minidom` when generating the XML output: https://docs.python.org/3/library/xml.dom.minidom.html?highlight=minidom#xml.dom.minidom.Node.toprettyxml  if `pretty` is `True`.  

This happens in `letterparser.generate.output_xml()` (https://github.com/elifesciences/decision-letter-parser/blob/develop/letterparser/generate.py#L232)

activity/activity_GenerateDecisionLetterJATS.py
gnott: This is a holdover from old lettuce tests I think when the logger was not always present. I can remove this 👍

activity/activity_ValidateDecisionLetterInput.py
gnott: Sure, it sounds like a good naming option.

gnott: This small function is present in two activities - a tiny bit of duplicated code as a result, but I'm not sure how useful it is for reuse, and if so, where to put it.

giorgiosironi: Very small amount of code so wouldn't worry about reuse

Issue #991, opened 2020-01-06
Remove hacky parentdir directory imports in starters


gnott:
As described in https://github.com/elifesciences/elife-bot/pull/990#discussion_r358311879, there are still some old hacky ways to import parent directories in starter code. The object is to test whether or why these are required, and if they are still required, refactor it so they can import what they need properly in a normal way.

# DoD:

- [x] No more `os.path.dirname(os.path.dirname(...` style imports required

# Tasks / plan

- [x] Remove all the weird imports
- [x] Test it on `end2end` testing - revert the PR if it fails testing
- [x] Debug further (if it failed) and add more todo tasks to fix it

gnott:
A note about some old memories about why this import line may be present.

**Possible reason 1**
There was a time early on where what was the `elife-tools` code existed in a folder that was a folder at the same level as the `elife-bot` folder. In order to import it (because `elife-tools` was not a pip-instalable module) some parts needed to also import from above the `elife-bot` folder.

**Possible reason 2**
Another possible reason is for the dynamic loading of activity objects. I think this was more likely happening in the `worker.py`, which no longer uses this parent directory import mess. It was because when the module name to dynamically import was concatenated, it needed to resolve the module name into folder names.

**Possible reason 3**
It's possible in Python 2 it was hard to do a relative import of a module, which is no longer a problem with Python 3

**Possible reason 4**
It could be that early on some key folders did not have a `__init__.py` file to make it possible to import from them easily.

If these imports solved one of these four reasons at some time in the past, they probably no longer apply, and we'll have no headaches in removing these weird imports. The one place that may matter is in the dynamic importing of modules, and if we run into that, then we can improve the code to make it work more clearly.

gnott:
After testing the latest merged PR experiment, I think the problem is somewhat related to the above reasons but slightly different.

I was able to execute a `SilentCorrectionsIngest` workflow fine as a result of adding a zip file to the S3 bucket.

The probably happens when I tested the starter for `PubmedArticleDeposit` locally.

If I am in `elife-bot` folder and I try to execute:

```
python starter_PubmedArticleDeposit.py -e dev
```

I get

```
Traceback (most recent call last):
  File "starter/starter_PubmedArticleDeposit.py", line 8, in <module>
    import log
ImportError: No module named 'log'
```

My understanding so far is, if a starter is run as a script, then the "package" structure will begin at `starter/`, and so it is unable to import `log.py` from the root directory of the project.

[...]

PR #1020, opened 2020-01-27
Clean up imports in starter_AdminEmail.

gnott:
A first experiment towards issue https://github.com/elifesciences/elife-bot/issues/991

`AdminEmail` workflow is old, we considered to remove it, but haven't yet. The `cron.py` starts one every four hours, it gets some metadata from SWF, and sends an email with some stats. 

It has no effect on other workflows, so it looks like a good place to remove old parent directory imports.

If this continues to run after deploying to `prod`, we can remove it from another one or two starters that are lower priority, like for `SilentCorrectionIngest` and `PubmedArticleDeposit`, which are triggered by the SQS queue listener.

### DIFF
--- a/starter/starter_AdminEmail.py
+++ b/starter/starter_AdminEmail.py
@@ -1,8 +1,3 @@
-import os
-# Add parent directory for imports
-parentdir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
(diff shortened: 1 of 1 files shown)


gnott:
I received an admin email at 21:45 UTC, so this removal was a success.

PR #1023, opened 2020-01-30
Parent directory and starter imports cleaned up

gnott:
Fixes https://github.com/elifesciences/elife-bot/issues/991

Simple removal of the old parent directory imports, removed the import of any modules the linter said were not used.


gnott:
All that remains, I think, is to observe it on `end2end` and `continuumtest` environments, as possible, and monitor `prod` after it is deployed. It it time to give it a go. I can watch it today.

PR #1024, opened 2020-01-30
No indent for decision letter XML output.

gnott:
I tested generating decision letter output again after fixing the bucket names, and it worked; the output bucket contained a `.xml` file.

I downloaded it to check and see all looked well, and then I realised it was _too_ pretty.

I remember now the best output to produce is pretty XML -- which is a new line for each tag -- but do not indent, otherwise we get extra whitespace before every tag.

### DIFF
--- a/activity/activity_ValidateDecisionLetterInput.py
+++ b/activity/activity_ValidateDecisionLetterInput.py
@@ -78,7 +78,7 @@ class activity_ValidateDecisionLetterInput(Activity):
             self.directories.get("TEMP_DIR"),
             self.logger,
             pretty=True,
-            indent="    ")
+            indent="")
 
         self.set_statuses(statuses)
 
--- a/activity/activity_GenerateDecisionLetterJATS.py
+++ b/activity/activity_GenerateDecisionLetterJATS.py
@@ -79,7 +79,7 @@ class activity_GenerateDecisionLetterJATS(Activity):
             self.directories.get("TEMP_DIR"),
             self.logger,
             pretty=True,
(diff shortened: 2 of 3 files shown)


activity/activity_ValidateDecisionLetterInput.py
```diff
@@ -78,7 +78,7 @@ def do_activity(self, data=None):
             self.directories.get("TEMP_DIR"),
             self.logger,
             pretty=True,
-            indent="    ")
+            indent="")
```
giorgiosironi: isn't there a default for this argument? But no problem in being explicit

giorgiosironi:
No regression test? Realizing it's cumbersome to unit test activities so may require a lot of stubbing for little value.

activity/activity_ValidateDecisionLetterInput.py
gnott: `indent=""` is the default, yes, but I like it visible in the activity too, although now I realise `""` is the only value we'd want to consider for optimal whitespace.

gnott:
There are tests for this activity as well as the function in the `letterparser_provider.py` that generates XML. It is not concerned with whitespace though, and it doesn't compare the full XML output to a test fixture at this time. It only checks the XML is produced, and that it contains an expected DOI value at this time. I planned the tests can be expanded, if need be, based on actual errors we observe when we use this workflow more.

gnott:
Tested this again on `continnumtest` environment, and the XML in the output bucket has not too extra whitespace 👍

Issue #1019, opened 2020-01-27
PoA packaging workflow to rely on S3 notifications


gnott:
This issue is a more detailed list of actions to finish point 1 on issue https://github.com/elifesciences/elife-bot/issues/565, where the end goal is to remove all dependence on SimpleDB.

The proposed new PoA packaging procedure in summary:
- a PoA `.zip` file input arrives into the `elife-ejp-poa-delivery` bucket
- results in an S3 notification, which is added to a new SQS queue, the queue having a lengthy message retention time (between 4 and 14 days we're thinking)
- daily a the `cron.py` initiates the PoA packaging workflows at a specific time (this already happens currently)
- for each message in the queue, start a `PackagePOA` workflow for each `.zip` file (this happens now, but the list of `.zip` files is taken from SimpleDB data from S3 bucket polling)

Steps to completion:
- [x] create new SQS queue with long message retention time (@giorgiosironi) - possibly named `poa-incoming-queue`
- [x] add S3 notification configuration to the `elife-ejp-poa-delivery` bucket manually (@giorgiosironi)
- [x] (optional) mirror the `elife-ejp-poa-delivery` S3 notification settings in the builder / infrastructure definitions (@giorgiosironi)
- [x] refactor the existing `starter/cron_NewS3POA.py` file to use SQS queue data instead of SimpleDB data for getting PoA `.zip` file names
- [x] Note: prior to deploying this new PoA logic to `prod` environment, may want to clear out the existing SQS messages so the files do not get processed again

I considered to try and integrate with existing SQS queue listeners use in the article ingest workflows, but since those are all based around processing SQS messages immediately upon receiving them, I think keeping the delayed PoA packaging logic separate from those is probably less messy. There may be some connection and message processing logic that can be consolidated.

The first three steps @giorgiosironi if you could please lend some help on setting up queues and infrastructure? I'm happy to give it a go, but you may have suggestions and requirements I'm not fully aware of. Thanks!

gnott:
Suggestion of new queue name: `poa-incoming-queue`

giorgiosironi:
> since those are all based around processing SQS messages immediately upon receiving them, I think keeping the delayed PoA packaging logic separate from those is probably less messy.

Yes.

giorgiosironi:
Review https://github.com/elifesciences/builder/pull/606 which will apply queue and notifications to `end2end` and `continuumtest`; plus create the queue only on prod.

[...]

Issue #565, opened 2017-12-14
Replacement for SimpleDB


gnott:
Here is my proposal for how to remove SimpleDB as a requirement in the `elife-bot` project.

A very brief update: Some of the parts that depended on SimpleDB have been removed, but some still remain. For a while I thought about how the state could be stored elsewhere, only recently I think the combination of logic and data available makes it possible to remove what SimpleDB provides entirely from the project.

The proposed solution comes as three major parts, which can be done separately and/or concurrently, and can also be tracked in other tickets for each of the ~three~ four parts. These parts are:

1. Change the PoA packaging workflow to rely on S3 notifications instead of polling S3 buckets
2. `PublicationEmail` to be switched over to explicit email sending logic instead of relying on duplicate email checking
3. Send all the emails from the bot directly as SMTP mails, and do away with the email sending queue
4. Remove all references to the `simpleDB` provider in all other code

Todos:

- [x] Remove bucket polling in favour of S3 notifications for PoA workflow
- [x] Refactor and switch over PublicationEmail to use SMTP sending and different way to not send duplicates
- [x] Replace all places emails are added to the email queue to instead send emails immediately by SMTP - Ticket https://github.com/elifesciences/elife-bot/issues/854
- [x] Remove all remaining `simpleDB` provider imports in any other files, and then delete the `simpleDB.py` module
- [x] Remove any documentation in the project about SimpleDB
- [x] Can probably remove `settings.prefix` and `settings.delimiter` from `settings.py`, which I think was only used by SimpleDB logic

A little more about each of these solution parts for discussion, which can be duplicated over to their own tickets later.

**1. Change the PoA packaging workflow**
The last thing that relies on S3 bucket polling data is the PoA packaging workflow. It is run once a day at a particular time of day. At that time, it queries SimpleDB records to find the files that have been uploaded in the past 24 hours, and processes those.

My suggested solution is:
* enable S3 notifications on the bucket the receives these PoA zip files
* send those notifications to a new SQS queue (@giorgiosironi if you could please comment and help with infrastructure to create the queue and config)
* A new, separate, PoA queue listener will get messages from this queue when it intends to process the zip files, currently only once a day, at a particular time as specified in the `cron.py`. (In future we can package the files more than once a day, but again only at particular times, due to synchronicity with the `.csv` files also required).

Essentially saving state, but only for a limited time, of what files arrived, so the messages on the queue would ideally stay in flight for at least 72 hours, maybe 7 days might be safe.

[...]

Issue #992, opened 2020-01-06
Refactor and standardise workflow starters


gnott:
As mentioned in https://github.com/elifesciences/elife-bot/pull/990#discussion_r358312510, there is similar code across many different starter modules, indicating a strong possibility there could be simpler and reusable ways to call a starter.

# DoD

- [x] Less code duplication across all the starter module code
- [x] More standardised list of starter parameters (if possible)
- [x] Clearer and more standard way to gather starter data (such as bucket object lists or other data sources used in starters)

# Tasks

- [x] Do a survey of all the starters, their parameters and logic, looking for similarities and differences
- [x] Move logic to a base starter object or to some starter helper libraries or functions that can be reused across all starters
- [x] Starter object (to be created) accepts a `logger` to be passed to it
- [x] Remove any starter code that is no longer in use
- [x] Aim to reduce the amount of code and complexity of code used in all the starters, so they are easier to understand
- [x] Use `ArgumentParser` instead of `OptionParser` for console start logic, and if possible standarise the parsing of arguments into `provider/utils.py`
- [x] Can probably use `helper.get_starter_logger()` everywhere for logging
- [x] Improve the test coverage or test scenarios

# Starter module list for tracking progress

- [x] starter/cron_FiveMinute.py
- [x] starter/cron_NewS3POA.py
- [x] starter/objects.py
- [x] starter/starter_AdminEmail.py
- [x] starter/starter_ApproveArticlePublication.py
- [x] starter/starter_CopyGlencoeStillImages.py
- [x] starter/starter_DepositCrossref.py
- [x] starter/starter_DepositCrossrefPeerReview.py
- [x] starter/starter_DepositDOAJ.py
- [x] starter/starter_FTPArticle.py
- [x] starter/starter_IngestAcceptedSubmission.py
- [x] starter/starter_IngestArticleZip.py
- [x] starter/starter_IngestDecisionLetter.py
- [x] starter/starter_IngestDigest.py
- [x] ~starter/starter_InitialArticleZip.py~ removed in PR https://github.com/elifesciences/elife-bot/pull/1299
- [x] starter/starter_LensArticlePublish.py
- [x] starter/starter_PMCDeposit.py
- [x] starter/starter_PackagePOA.py
- [x] starter/starter_Ping.py
- [x] starter/starter_PostPerfectPublication.py
- [x] starter/starter_ProcessArticleZip.py
- [x] starter/starter_PubRouterDeposit.py
[...]

PR #1026, opened 2020-01-31
Cron for PackagePOA using SQS queue with S3 notifications

gnott:
Fixes https://github.com/elifesciences/elife-bot/issues/1019
Re issue https://github.com/elifesciences/elife-bot/issues/565 for SimpleDB removal

**Required to merge before this PR** Settings for queue name in https://github.com/elifesciences/builder-private/pull/349

Summary of changes:

Following the simple plan of repurposing the existing `starter/cron_NewS3POA.py` module to get its data from an SQS queue instead of from SimpleDB data that was collected by polling the S3 bucket.

No longer required was getting data from SimpleDB or from the SWF history, so these providers were removed.

Since the logic based on timestamp calculations (`last_startDate`) to get the appropriate data from SimpleDB, that is also removed. This also led to removing this option from the `starter/starter_PackagePOA.py`, which now only supports passing in a file name to process.

Some parts of `cron_NewS3POA.py` could be moved `starter/starter_helper.py`, for potential reuse by other starters. This also included to refactor away using `eval()` to instantiate the starter object.

For the queue processing, my proposed method is to get messages from the queue (default is up to 100 messages), read each, try to start a workflow, and delete the message from the queue if a workflow was started. I think this should work well because there should only be this one consumer of that queue, it avoids looping in a `while()` loop, and in case there is a message that cannot be processed, it should be read only once (I think).

There are some rough edges in places, but this is ready for a first review I think.

The changes to starters may cause a conflict with open PR https://github.com/elifesciences/elife-bot/pull/1023, but it should be limited to two starters, and I can resolve those when it happens.

**Notes for merging and deployment:**
- [ ] Merge and deploy at a chosen time after all PoA packaging for the day is completed, and before that, empty old messages from the SQS queue so files are not packaged twice

### DIFF
--- a/starter/cron_NewS3POA.py
+++ b/starter/cron_NewS3POA.py
@@ -31,38 +31,12 @@ class cron_NewS3POA(object):
         logger = log.logger(logFile, settings.setLevel, ping_marker_id)
 
         # Start a ping workflow as a marker
-        self.start_ping_marker(ping_marker_id, settings)
+        helper.start_ping_marker(ping_marker_id, settings, logger)
 
         # Get data from SQS queue
         sqs_conn = sqs_connect(settings)
         sqs_queue = get_sqs_queue(sqs_conn, settings)
         process_queue(sqs_queue, settings, logger)
-
-    def start_ping_marker(self, workflow_id, settings):
-        """
-        Start a ping workflow with a unique name to serve as a time marker
-        for determining last time this was run
-        """
-
-        workflow_id = workflow_id
-        workflow_name = "Ping"
-        workflow_version = "1"
-        child_policy = None
-        execution_start_to_close_timeout = None
-        input = None
-
-        conn = boto.swf.layer1.Layer1(settings.aws_access_key_id, settings.aws_secret_access_key)
-        try:
-            response = conn.start_workflow_execution(settings.domain, workflow_id, workflow_name,
-                                                     workflow_version, settings.default_task_list,
-                                                     child_policy, execution_start_to_close_timeout,
-                                                     input)
-
-        except boto.swf.exceptions.SWFWorkflowExecutionAlreadyStartedError:
-            # There is already a running workflow with that ID, cannot start another
-            message = ('SWFWorkflowExecutionAlreadyStartedError: There is already ' +
-                       'a running workflow with ID %s' % workflow_id)
-            print(message)
 
 
 def sqs_connect(settings):
--- a/starter/cron_NewS3POA.py
+++ b/starter/cron_NewS3POA.py
@@ -77,7 +77,7 @@ def start_package_poa_workflow(sqs_message, settings, logger):
         logger.info('Started %s workflow for document %s' % (starter_name, document))
         return True
     except:
-        logger.exception('Error: starting %s for document %s' % (starter_PackagePOA, document))
+        logger.exception('Error: starting %s for document %s' % (starter_name, document))
 
 
 if __name__ == "__main__":
--- a/starter/cron_NewS3POA.py
+++ b/starter/cron_NewS3POA.py
@@ -10,7 +10,8 @@ from S3utility.s3_sqs_message import S3SQSMessage
 Cron job to check for new article S3 POA and start workflows
 """
 
-MAX_MESSAGE_COUNT = 100
+# Note: boto accepts a number between 1 and 10
+MAX_MESSAGE_COUNT = 10
 
 
 class cron_NewS3POA(object):
--- a/starter/cron_NewS3POA.py
+++ b/starter/cron_NewS3POA.py
@@ -51,9 +51,10 @@ def process_queue(sqs_queue, settings, logger):
     """reach each message from the queue, start a workflow, then delete the message"""
     message_count = 0
     while True:
-        messages, status = get_queue_messages(sqs_queue, MAX_MESSAGE_COUNT, logger)
-        if not status:
-            logger.error('Breaking process queue read loop, failed to get messages from queue')
+        try:
+            messages = get_queue_messages(sqs_queue, MAX_MESSAGE_COUNT, logger)
+        except:
+            logger.exception('Breaking process queue read loop, failed to get messages from queue')
             break
 
         # check if any messages to process
@@ -91,10 +92,10 @@ def process_queue(sqs_queue, settings, logger):
 
 def get_queue_messages(sqs_queue, num_messages, logger):
     try:
-        return sqs_queue.get_messages(num_messages), True
+        return sqs_queue.get_messages(num_messages)
     except:
         logger.exception('Exception in getting messages from SQS queue')
-    return [], False
+        raise
 
 
 def start_package_poa_workflow(sqs_message, settings, logger):
--- a/starter/cron_NewS3POA.py
+++ b/starter/cron_NewS3POA.py
@@ -6,12 +6,11 @@ os.sys.path.insert(0, parentdir)
 import boto.swf
 import log
 import time
-import importlib
 
 import provider.simpleDB as dblib
 import provider.swfmeta as swfmetalib
 from provider import utils
-import starter
+import starter.starter_helper as helper
 
 """
 Cron job to check for new article S3 POA and start workflows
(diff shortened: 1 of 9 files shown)


starter/cron_NewS3POA.py
```diff
@@ -27,106 +30,62 @@ def start(self, settings):
...
+    queue = sqs_conn.get_queue(settings.poa_incoming_queue)
+    queue.set_message_class(S3SQSMessage)
+    return queue
+
+
+def process_queue(sqs_queue, settings, logger):
+    """reach each message from the queue, start a workflow, then delete the message"""
+    messages = None
+    if sqs_queue:
+        messages = sqs_queue.get_messages(MAX_MESSAGE_COUNT)
```
giorgiosironi: what if there's more? Wait the next day? Should probably be in a loop?

giorgiosironi: i.e. this limit is dictated by how big an API response can be, rather than by any business logic of batching in POAs

starter/cron_NewS3POA.py
```diff
@@ -27,106 +30,62 @@ def start(self, settings):
...
+    if not messages:
+        logger.info('no messages available')
+        return
+    for message in messages:
```
giorgiosironi: The inside the `for` loop needs to be:
- inside a `try` block
- with a `finally` that deletes the message (otherwise an unreadable message will be retried arbitrarily at the next execution, over and over)
- with an `except` that logs errors to a channel where they can be addressed, as the message won't be there anymore

starter/cron_NewS3POA.py
```diff
@@ -27,106 +30,62 @@ def start(self, settings):
...
+        helper.import_starter_module(starter_name, logger)
+        starter_object = helper.get_starter_module(starter_name, logger)
+        starter_object.start(settings=settings, document=document)
+        logger.info('Started %s workflow for document %s' % (starter_name, document))
+        return True
+    except:
+        logger.exception('Error: starting %s for document %s' % (starter_PackagePOA, document))
+
 
 if __name__ == "__main__":
```
giorgiosironi: you probably want a way to start the workflow with input data that you specify from the command line, for example to re-execute a message manually after it has failed, it has been logged, and a code fix has been deployed

giorgiosironi: Probably this is covered in the `starter_PackagePOA` module itself rather than in this cron script

starter/starter_PackagePOA.py
```diff
@@ -9,129 +9,92 @@
...
+        if doc:
+
+            document = doc["document"]
+
+            # Get a unique id from the document name for the workflow_id
+            id_string = None
+            try:
+                id_string = document.split("_")[0]
+            except:
+                id_string = "000"
```
giorgiosironi: can't something more unique be generated? Don't expect this to happen often, but would still be useful to distinguish between different cases unless `000` is a convention meaning something?

starter/starter_PackagePOA.py
```diff
@@ -9,129 +9,92 @@
...
-import provider.simpleDB as dblib
 
 """
 Amazon SWF PackagePOA starter
 """
 
 class starter_PackagePOA():
 
-    def start(self, settings, document=None, last_updated_since=None):
+    def start(self, settings, document=None):
```
giorgiosironi: when does this *not* have a document? Since the cron executes the loop, I only see the cron-based and the manual use case, both of which have one document being passed in

starter/starter_PackagePOA.py
```diff
@@ -9,129 +9,92 @@
...
+                    settings.default_task_list,
+                    child_policy,
+                    execution_start_to_close_timeout,
+                    input)
+
+                logger.info('got response: \n%s' %
+                            json.dumps(response, sort_keys=True, indent=4))
+
+            except boto.swf.exceptions.SWFWorkflowExecutionAlreadyStartedError:
+                # There is already a running workflow with that ID, cannot start another
```
giorgiosironi: fair, I think we usually ignore this. Want to log an error rather than an info level message?

starter/starter_helper.py
```diff
@@ -1,33 +1,97 @@
...
+    Given an starter name as starter_name,
+    attempt to lazy load the module when needed
+    """
+    try:
+        module_name = "starter." + starter_name
+        importlib.import_module(module_name)
+        return True
+    except ImportError:
+        if logger:
+            logger.exception('')
```
giorgiosironi: I know that `logger.exception` already adds information extracted magically from the exception object, but still another unique message should be used here to help locate where the log came from

starter/starter_helper.py
```diff
@@ -1,33 +1,97 @@
...
+        return False
+
+
+def start_ping_marker(workflow_id, settings, logger):
```
giorgiosironi: weird way to log but already existing

starter/cron_NewS3POA.py
gnott: I wanted to avoid an infinate loop if a message that couldn't be processed is encountered, but I think your suggestion (as I interpret it) to log and then delete any message that is unsuitable should guard against that. I will refactor!

starter/cron_NewS3POA.py
gnott: The `starter_PackagePOA` is the one I've used in the past if we want to process a particular `.zip` PoA input file again. That would be the manual procedure to process just a single file. This `cron_NewS3POA.py` is to read the queue and then start 1 to n `PackagePOA` workflows, using this particular starter and knowing the zip file name.

starter/starter_PackagePOA.py
gnott: A good observation, thanks. The optional value made sense when you could supply either `document` and/or `last_updated_since`. Now there is only one parameter for the input data, it is mandatory for anything to happen.

starter/starter_PackagePOA.py
gnott: We can probably use `document` in the workflow execution id. I cannot remember what characters are allowed or disallowed there in SWF, but the zip file names we expect are fairly simple.

starter/starter_PackagePOA.py
gnott: I think examples of these will get more closer attention when starter refactoring in issue https://github.com/elifesciences/elife-bot/issues/992 happens. I will change this from `info` to `error` though.

This exception may have happened on occassion if a long running workflow execution is still in progress and the cron tried to start another, such as in `S3Monitor` workflows or similar, or it could catch a duplicate workflow in a short time period due to a glitch. These examples are probably all historical. I expect in the refactoring we can catch this particular exception just once for all the times that starters try to start a workflow execution so there is less duplicate code.

starter/starter_helper.py
gnott: I think I just copy and pasted this from the existing starter code. I'm adding unique messages to the two `logger.exception()`'s here.

starter/starter_helper.py
gnott: These insert data into the SWF history so it can be relied on. Getting rid of this would be accompanied with issue https://github.com/elifesciences/issues/issues/5052 so the cron doesn't rely on the workflow history to determine when it is time to start a workflow execution within a certain time window.

gnott:
Improved the queue reading logic in `starter/cron_NewS3POA.py` I hope, along your suggested lines @giorgiosironi with a `try`, `except`, `finally`. There is some additional handling of connection exceptions, and deleting messages that are not applicable.

starter/cron_NewS3POA.py
```diff
@@ -22,106 +23,97 @@ def start(self, settings):
...
+    while True:
+        messages, status = get_queue_messages(sqs_queue, MAX_MESSAGE_COUNT, logger)
+        if not status:
+            logger.error('Breaking process queue read loop, failed to get messages from queue')
+            break
+
+        # check if any messages to process
+        if not messages:
+            logger.info('no messages available')
+            break
```
giorgiosironi: good exit condition

starter/cron_NewS3POA.py
```diff
@@ -22,106 +23,97 @@ def start(self, settings):
...
+            finally:
+                sqs_queue.delete_message(message)
+
+
+def get_queue_messages(sqs_queue, num_messages, logger):
+    try:
+        return sqs_queue.get_messages(num_messages), True
+    except:
+        logger.exception('Exception in getting messages from SQS queue')
+    return [], False
```
giorgiosironi: re-raise the exception here? https://refactoring.guru/replace-error-code-with-exception
The calling code can catch it if it wants to log something in its context as well.

gnott: I considered this yesterday too, to rely on the exception alone. I thought it would complicate the caller more, but I'll give it a go!

gnott:
Good I did some testing on `continuumtest` manually before merging, as it turns out the number of messages to read from the queue at one time is limited to 10. The 100 messages raises and exception. I am dialing that back down to 10.

gnott:
I am checking and pruning the `prod` SQS queue for old files, and merging + deploying this code ASAP to be ready for a first run on Thursday at the regular time.

PR #1028, opened 2020-02-03
Success in PDF covers if stated assertion fails.

gnott:
Re issue https://github.com/elifesciences/issues/issues/5364

There are some cases where a personalised cover PDF will not be created, sometimes resulting in a status `500` from the service call. This results in a failed workflow. It doesn't really cause too much trouble if the activity is the last one in a workflow, but it is hard to determine why the workflow failed without having to dig deeply into the list of activities to see if it failed earlier on, in where it archives the article to a `.zip` file, for example.

Code change here is a small improvement to not fail if an error `500` happens. The exception is still logged to file. If an exception beyond the assertions on the status code value happens (something that is not defined in the code) it will still result in a failed workflow. This might be due to anything in the process that does not end in an HTTP response from the service.

### DIFF
--- a/activity/activity_GeneratePDFCovers.py
+++ b/activity/activity_GeneratePDFCovers.py
@@ -62,7 +62,7 @@ class activity_GeneratePDFCovers(Activity):
             self.logger.error(error_message)
             self.emit_monitor_event(self.settings, article_id, version, run,
                                     self.pretty_name, "error", error_message)
-            return self.ACTIVITY_PERMANENT_FAILURE
+            return self.ACTIVITY_SUCCESS
 
         except Exception as e:
             error_message = str(e)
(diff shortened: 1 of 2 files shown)


activity/activity_GeneratePDFCovers.py
```diff
@@ -62,7 +62,7 @@ def do_activity(self, data):
             self.logger.error(error_message)
             self.emit_monitor_event(self.settings, article_id, version, run,
                                     self.pretty_name, "error", error_message)
-            return self.ACTIVITY_PERMANENT_FAILURE
+            return self.ACTIVITY_SUCCESS
```
giorgiosironi: this will ignore any error, including the service being completely broken. Can't we do anything better?

gnott: A very simple option is we could return a `ACTIVITY_TEMPORARY_FAILURE` which would result in retying this activity every 10 seconds until the workflow ultimately reaches `Timed Out` status. That would be slightly better than it resulting in `Failed` to me. The covers that will never be generated will always result in `Timed Out`.

I'll add more details of other options to consider into issue https://github.com/elifesciences/issues/issues/5364.

gnott:
I'll do a new PR with different logic, after changing the proposed solution.

PR #1030, opened 2020-02-04
Fix starter import in cron_NewS3POA.

gnott:
There was a problem with the old `eval` method of loading a `PackagePOA` starter by the `cron_NewS3POA` script, due to too aggresive pruning of the imports. `import starter` seems to be required still using this old method.

An updated PoA cron process is coming soon, which also no longer uses `eval`, but for now this needs to be fixed for current operations.


gnott:
Tested this on `continuumtest` environment and I'm fairly sure it is good now.

PR #1033, opened 2020-02-04
Remove unused code from cron_FiveMinute.py

gnott:
It was odd in bug fix PR https://github.com/elifesciences/elife-bot/pull/1030 where the new PoA file cron script was unable to import the starter class it wanted, since in `starter/cron_FiveMinute.py` was starting workflows fine.

On closer inspection, `cron_FiveMinute.py` doesn't actually import any starters, even though the code is there.

This PR removes the code that is not run anyway, to simplify things, prior to more extensive starter refactoring.

### DIFF
--- a/starter/cron_FiveMinute.py
+++ b/starter/cron_FiveMinute.py
@@ -63,35 +63,6 @@ class cron_FiveMinute(object):
                        'a running workflow with ID %s' % workflow_id)
             print(message)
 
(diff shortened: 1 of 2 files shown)


gnott:
Looking back, I remember whey this code was here, because in the past every five minutes this started a `SendQueuedEmail` workflow execution. This five minute cron piece may in the future be removed entirely, but removing this unused code is a good start.

PR #1034, opened 2020-02-05
Remove SimpleDB and S3Monitor

gnott:
Fixes https://github.com/elifesciences/elife-bot/issues/565

With PoA logic switched over to SQS and S3 notifications now, I see nothing to stop us from no longer using

- `S3Monitor`, to poll and track S3 bucket files
- `SimpleDB`

Basically, I deleted all the main files for these workflows and providers, removed any references to them elsewhere (not very widely used anymore, as planned), removed mentions in documentation, test fixtures, and removed settings we no longer require.

When this is merged and deployed, I will continue on issue https://github.com/elifesciences/issues/issues/5388 to delete the data eLife has stored in SimpleDB, which is now obsolete.


gnott:
Safe to merge I think, here goes!

Issue #968, opened 2019-10-29
DepositDecisionLetterIngestAssets activity


gnott:
From issue https://github.com/elifesciences/elife-bot/issues/965

This activity will copy figure image files and video files to an S3 bucket that are part of a decision letter & author response. The image files may be `.tif`, `.jpg`, `.png`, possibly `.gif`. The video files may be `.mp4` or some other format (I'm not sure exactly which types of videos are acceptable, and they will just be copied over as received).

- [x] ~~Rename asset files taken from the `.zip` file in the input S3 bucket~~ - No, going to leave the files as named for now, we can improve file renaming as an enhancement later
- [x] Copy the files to the output S3 bucket, into a subfolder like `elife00666/`

These files can then be accessed by typesetters. 

Additional assumptions:

- We do not need to convert the image files into a different format or into dimensions
- These particular files will not need to be made public from this output bucket
- These files do not need to be served by a CDN
- These files do not need to be processed by a IIIF image server

PR #1035, opened 2020-02-06
Decision letter assets activity

gnott:
Fixes https://github.com/elifesciences/elife-bot/issues/968

New activity named `DepositDecisionLetterIngestAssets`. It unzips the decision letter zip file, and saves each asset file to the output bucket into the specific subfolder for the article.

It is added to the `IngestDecisionLetter`.

### DIFF
--- a/activity/activity_DepositDecisionLetterIngestAssets.py
+++ b/activity/activity_DepositDecisionLetterIngestAssets.py
@@ -90,8 +90,9 @@ class activity_DepositDecisionLetterIngestAssets(Activity):
 
         # deposit assets to the bucket
         try:
-            self.deposit_assets_to_bucket(
-                output_bucket_name, bucket_folder_name, self.asset_file_names)
+            deposit_assets_to_bucket(
+                self.settings, output_bucket_name, bucket_folder_name,
+                self.asset_file_names, self.logger)
         except:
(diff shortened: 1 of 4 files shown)


gnott:
Might need to ask Alfred to run the tests again, there might be a temporary issue there.

activity/activity_DepositDecisionLetterIngestAssets.py
```diff
@@ -0,0 +1,124 @@
...
+        manuscript = letterparser_provider.manuscript_from_articles(self.articles)
+        bucket_folder_name = letterparser_provider.output_bucket_folder_name(
+            self.settings, manuscript)
+
+        self.logger.info(
+            "%s asset file names from %s: %s" % (self.name, self.input_file, self.asset_file_names))
+
+        # deposit assets to the bucket
+        try:
+            deposit_assets_to_bucket(
```
giorgiosironi: the try could span more of the activity, as previous functions from the `letterparser_provider` can fail

gnott: It would be unlikely for earlier steps to fail, since in the workflow, there are activities prior to this one that will have vetted the input content. But, I'm all for more exception handling. I will wrap another portion of code in a separate exception clause.

PR #1039, opened 2020-02-08
Do not generate PDF covers if article has no subjects

gnott:
Re: issue https://github.com/elifesciences/issues/issues/5364

I compared the XML of articles that fail personalised PDF cover generation with articles that work; the ones that fail do not have any `<subj-group subj-group-type="heading">` tags. These values result in a `subjects` key in the article JSON from the RAML API.

I think the personalised covers PDF service uses the eLife API data to determine which background image to use in the PDF file. I haven't seen the logs of the PDF service, but I suspect the error `500` we see has something to do with the missing `subjects` value and/or it cannot locate a `.jpg` file for when subjects do not exist.

Here, I added `lax_provider` as a dependency of the `GeneratePDFCovers` activity.

If there is no `subjects` key in the snippet for this article ID version from Lax, then we do not attempt to generate a PDF for it.

In this situation, and if there is an exception or an error, it only logs this data to the `elife-bot` instance's `worker.log` file, and the error doesn't appear on the publishing dashboard, since we haven't emitted any messages to the queue. I wasn't sure whether this is a good way to go, but it satisfies the basic requirement of not resulting in a `Failed` workflow in SWF if the PDF generation service will fail in this known case.

I also linted and cleaned up the test case scenario file for long lines and duplicated test data.

### DIFF
--- a/activity/activity_GeneratePDFCovers.py
+++ b/activity/activity_GeneratePDFCovers.py
@@ -43,7 +43,7 @@ class activity_GeneratePDFCovers(Activity):
                         (article_id, self.name))
                     return self.ACTIVITY_SUCCESS
         except Exception:
-            self.logger.exception("Exception in %s getting data snippet from Lax" % (self.name))
+            self.logger.exception("Exception in data snippet in %s" % (self.name))
             return self.ACTIVITY_PERMANENT_FAILURE
 
         # generate the PDF
--- a/activity/activity_GeneratePDFCovers.py
+++ b/activity/activity_GeneratePDFCovers.py
@@ -1,4 +1,5 @@
 import provider.article as articlelib
+from provider import lax_provider
 from activity.objects import Activity
 
 """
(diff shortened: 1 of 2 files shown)


activity/activity_GeneratePDFCovers.py
```diff
@@ -29,6 +30,23 @@ def do_activity(self, data):
...
+                self.logger.error("Article %s has no Lax data snippet in %s." % (article_id, self.name))
+                return self.ACTIVITY_PERMANENT_FAILURE
+            else:
+                if 'subjects' not in snippet:
+                    self.logger.info(
+                        "Article %s has no subjects in %s, not creating a PDF." %
+                        (article_id, self.name))
+                    return self.ACTIVITY_SUCCESS
+        except Exception:
+            self.logger.exception("Exception in %s getting data snippet from Lax" % (self.name))
```
giorgiosironi: error message is a bit too specific, could even be `self.logger` problem or anything else

gnott: Anything is possible! I am changing it to a more generic, but unique, message `Exception in data snippet`.

Issue #969, opened 2019-10-29
PostDecisionLetterJATS activity


gnott:
From issue https://github.com/elifesciences/elife-bot/issues/965

This activity will take decision letter & author response JATS XML and issue a POST request to an API endpoint. The endpoint will be provided by typesetters.

The details about the actual POST are not final yet. My assumption is it may follow the pattern of the digest API endpoint in regards to the authentication, specifying the article DOI, and encapsulating the XML data. On this assumption of similarity, the tasks to build it would include:

- [x] Test API endpoint receiving JATS XML as a `POST`
- [x] Download the JATS XML from an S3 bucket (or generate it again from the original `.zip` file contents)
- [x] Refactor to reuse code from `PostDigestJATS` activity, where applicable
- [x] Assemble a POST request payload, containing `apiKey`, `accountKey`, `doi`, `type`, `content`
- [x] Invoke the POST request
- [x] Send an email when a successful post is completed to certain recipients
- [x] Send an email upon error to other recipients

The API endpoint is not available yet. In the meantime, the POST request can be disabled until the endpoint can be connected up and tested.

gnott:
In progress, the current status is some manual testing of `POST`ing JATS XML to the typesetter's endpoint before doing too much code writing.

PR #1040, opened 2020-02-11
Post decision letter JATS activity

gnott:
Re issue https://github.com/elifesciences/elife-bot/issues/969

Blocked by a future PR to add new settings before this PR can be merged.

The simple story about this new activity is 
- it will download the decision letter JATS XML from a bucket, the object's location is already saved in a workflow's session data,
- it extracts a DOI from the XML,
- it issues a `POST` to an endpoint
- it sends success or error emails depending on the response of the `POST`

There is still some testing underway with the real endpoint, but it shouldn't alter how the logic looks as presented here.

Test coverage is fairly good, but I'm sure there's room for improvement and comments!

### DIFF
--- a/activity/activity_PostDecisionLetterJATS.py
+++ b/activity/activity_PostDecisionLetterJATS.py
@@ -1,6 +1,7 @@
 import os
 import json
 import time
+from requests.exceptions import HTTPError
 from provider import (
     download_helper, email_provider, letterparser_provider, requests_provider, utils)
 from provider.execution_context import get_session
@@ -24,6 +25,7 @@ class activity_PostDecisionLetterJATS(Activity):
         # Track some values
         self.xml_file = None
         self.doi = None
+        self.post_error_message = None
 
         # Local directory settings
         self.directories = {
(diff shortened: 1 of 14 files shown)


gnott:
I continued to refactor, creating a new `requests_provider.py` provider so the posting of Digest and Decision letter content to the endpoint can share some code.

provider/requests_provider.py
```diff
@@ -0,0 +1,54 @@
...
+        full_error_message = (
+            "%s\npayload: %s" %
+            (response_error_message, payload))
+        logger.error(full_error_message)
+        return response_error_message
+    logger.info(
+        ("Success posting %s to endpoint %s: status_code: %s\nresponse: %s" +
+         " \npayload: %s") %
+        (identifier, url, resp.status_code, resp.content, payload))
+    return True
```
giorgiosironi: use exceptions here over return codes that can be mistaken as true by `if post-to_endpoint(...)` statements:
- positive case: return `None` (hence not return anything)
- negative case: `raise RequestsProviderError(...)` passing in all the information you want in this custom exception

activity/activity_PostDecisionLetterJATS.py
```diff
@@ -0,0 +1,197 @@
...
+        """prepare and POST jats to API endpoint"""
+        url = self.settings.typesetter_decision_letter_endpoint
+        payload = requests_provider.jats_post_payload(
+            'decision', doi, jats_content, self.settings.typesetter_decision_letter_api_key)
+        if payload:
+            return requests_provider.post_to_endpoint(
+                url, payload, self.logger, 'decision letter JATS')
+        return None
+
+    def send_email(self, doi, jats_content):
```
giorgiosironi: from here on it's all email-related logic; consider what could be pushed down into the provider as it doesn't seem related to the main purpose of the activity, talking to a remote (Exeter?) API

provider/requests_provider.py
gnott: I've changed this to something that will hopefully work better.

First off, a new exception catching around the request call itself.

Then, if the status code of the response is not `200`, then raise a `requests.exceptions.HTTPError` exception.

The error message that was formerly returned so it could be added to an email is used as the `HTTPErrror()` message, and that exception can then be added to the body of an error email.

Both this new activity and the `PostDigestJATS` activity are refactored to use the new exception catching.

gnott: In commit https://github.com/elifesciences/elife-bot/pull/1040/commits/65a3fa6247bcea967fa0f08732e58506294c5d95

gnott: I can switch over to a custom exception instead of `HTTPError` if that is unsuitable, what do you think?

activity/activity_PostDecisionLetterJATS.py
gnott: There are some slight differences in some of the email functions for digests and decision letters, but I've managed to consolidate them, move them.

gnott:
An additional commit https://github.com/elifesciences/elife-bot/pull/1040/commits/be738a92aa2f66118eceb1c1b4872045ba6d9656, after testing and discussion about the API endpoint authentication method, we need to also add the `apiKey` and `accountKey` to the `params` of the `POST`.

I also moved the `accountKey` value to be specified in `settings.py` instead of being hard-coded as `1`, so this PR will be blocked until a change to settings in the builder repo is merged.

giorgiosironi:
Unblocked `builder-private` and merged, but I don't have much context to do a code review here. Hope that was helpful.

gnott:
Sounds ok to me, it's an option to review, and I think it is ready to deploy.

Issue #970, opened 2019-10-29
DecisionLetterReceipt activity


gnott:
From issue https://github.com/elifesciences/elife-bot/issues/965

Similar to a proposed as-yet-built activity for the digest workflow (https://github.com/elifesciences/issues/issues/4593), here when a `IngestDecisionLetter` workflow reaches this activity, it notifies people that the `.zip` file they uploaded was processed successfully and in full. 

If no error email was sent by the workflow (due to a known, detectable error), and a runtime exception was logged, the person who deposited the `.zip` file to the bucket may not be aware of the exception.

- [x] Send an email to designated recipients (editorial, production)
- [x] The email can contain the ~~DOI,~~ original `.zip` file name processed ~~, and how it reached the final step successfully.~~ Update: for now we can just get by with the zip file name in the email subject line I think. In future we can expand the email body contents to include other helpful things in the future

This activity may not be required if the original depositors of the `.zip` file receive an email from the immediately previous activity, `PostDecisionLetterJATS`, which would potentially serve as an indicator that the final step of the workflow was reached for a decision letter & author response deposit.

PR #1043, opened 2020-02-14
Decision letter receipt

gnott:
Fixes https://github.com/elifesciences/elife-bot/issues/970

I'll consider this mildly blocked by PR https://github.com/elifesciences/elife-bot/pull/1040, so this next activity can be added to the workflow without causing a merge conflict.

Todo: 
- [x] Add this new activity to the `IngestDecisionLetter` worklfow

A very simple email sent to the recipients who would also be receiving validation error emails. The file name is in the email subject line.

The email body is currently empty, there being not much in particular I can think is required to include right now. In future maybe we can include the DOI and/or the full JATS produced for this decision letter content. I think just sending an email to denote the end of a successful workflow will be enough to get started using this new workflow.

### DIFF
--- a/activity/activity_DecisionLetterReceipt.py
+++ b/activity/activity_DecisionLetterReceipt.py
@@ -1,0 +1,66 @@
+import json
+import time
+from S3utility.s3_notification_info import parse_activity_data
+from provider import email_provider
+from activity.objects import Activity
+
+
+class activity_DecisionLetterReceipt(Activity):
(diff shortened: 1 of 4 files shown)


gnott:
Should be ready to review now, if you choose to, thanks!

gnott:
Small activity added here to a workflow that is in testing, and I will merge and deploy it now. We can expand on it in future when people start to use it.

PR #1053, opened 2020-03-12
Also purge digest data from CDN.

gnott:
Re issue https://github.com/elifesciences/issues/issues/5493

This makes it a standard thing to also purge the digest data from CDN caches at the same time it purges article data.

### DIFF
--- a/provider/fastly_provider.py
+++ b/provider/fastly_provider.py
@@ -15,7 +15,11 @@ class FastlyApi:
         response.raise_for_status()
         return response
 
-KEYS = ['article/{article_id}v{version}', 'article/{article_id}/videos']
+KEYS = [
+    'article/{article_id}v{version}',
+    'article/{article_id}/videos',
+    'digest/{article_id}',
+    ]
 
 def purge(article_id, version, settings):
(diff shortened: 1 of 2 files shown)


provider/fastly_provider.py
```diff
@@ -15,7 +15,11 @@ def purge(self, surrogate_key, service_id):
         response.raise_for_status()
         return response
 
-KEYS = ['article/{article_id}v{version}', 'article/{article_id}/videos']
+KEYS = [
+    'article/{article_id}v{version}',
+    'article/{article_id}/videos',
+    'digest/{article_id}',
```
giorgiosironi: the only service that uses this surrogate keys is [`iiif`](https://github.com/elifesciences/builder/blob/master/projects/elife.yaml#L1420), which is already in the list of services

gnott:
This activity is included in the `IngestArticleZip` and the `SilentCorrectionsIngest` workflows, and so it should invalidate digest images when any article version is ingested or silent correction ingested. This should cover any case where the digest image itself was silently deposited or for regular deposits.

PR #1055, opened 2020-03-16
Starter object and refactor Ping starter

gnott:
In regards to issue https://github.com/elifesciences/elife-bot/issues/992.

As a first attempt to refactor some of the starter logic, I selected the `Ping` workflow starter as a simple example.

To a new module, `starter/objects.py`, I created a base `Starter()` object, from which the Ping starter now inherits. The base `Starter` holds the logic for starting a workflow execution and to catch exceptions there.

In `starter_Ping.py`, I kept a method named `start()` for backwards compatibility. Other parts of the exisitng code depends on this method, which now invokes the new `start_workflow()` method anyway.

The `workflow_params` are set and collected in an `OrderedDict()` rather than having lots of separate property values floating around.

I wanted to keep the first PR on the starter refactoring to be as simple as possible, and this is it! I think this can gradually be applied to other starters and the code can be modified to account for different workflow parameters and logic in different circumstances as they are refactored.

### DIFF
--- a/starter/objects.py
+++ b/starter/objects.py
@@ -64,7 +64,7 @@ class Starter():
             raise
 
 
-def get_workflow_params(workflow):
+def default_workflow_params():
 
     workflow_params = OrderedDict()
     workflow_params['workflow_id'] = None
@@ -73,10 +73,5 @@ def get_workflow_params(workflow):
     workflow_params['child_policy'] = None
     workflow_params['execution_start_to_close_timeout'] = None
     workflow_params['input'] = None
-
-    if workflow == "Ping":
-        workflow_params['workflow_id'] = "ping_%s" % int(random.random() * 10000)
-        workflow_params['workflow_name'] = "Ping"
-        workflow_params['input'] = "1"
 
     return workflow_params
--- a/starter/objects.py
+++ b/starter/objects.py
@@ -1,5 +1,6 @@
 import random
 import json
+from collections import OrderedDict
 import boto.swf
 
 import log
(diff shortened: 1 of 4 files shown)


starter/starter_Ping.py
```diff
@@ -1,70 +1,46 @@
...
-        return (workflow_id, workflow_name, workflow_version, child_policy,
-                execution_start_to_close_timeout, input)
 
 if __name__ == "__main__":
 
     ENV = utils.console_start_env()
     SETTINGS = utils.get_settings(ENV)
 
-    o = starter_Ping()
+    STARTER_OBJECT = starter_Ping(SETTINGS)
```
giorgiosironi: just `STARTER` conflicts with anything?

starter/objects.py
```diff
@@ -0,0 +1,82 @@
...
+
+    workflow_params = OrderedDict()
+    workflow_params['workflow_id'] = None
+    workflow_params['workflow_name'] = None
+    workflow_params['workflow_version'] = None
+    workflow_params['child_policy'] = None
+    workflow_params['execution_start_to_close_timeout'] = None
+    workflow_params['input'] = None
+
+    if workflow == "Ping":
```
giorgiosironi: weird to have this in what is generic code; eventually going away?

starter/objects.py
```diff
@@ -0,0 +1,82 @@
...
+            self.logger.info(
+                'got response: \n%s', json.dumps(response, sort_keys=True, indent=4))
+
+        except boto.swf.exceptions.SWFWorkflowExecutionAlreadyStartedError:
+            # There is already a running workflow with that ID, cannot start another
+            message = (
+                'SWFWorkflowExecutionAlreadyStartedError: There is already ' +
+                'a running workflow with ID %s' % workflow_params.get('workflow_id'))
+            print(message)
+            self.logger.info(message)
```
giorgiosironi: this is a benign error, I guess

starter/objects.py
gnott: This will evolve, and a good observation, I will probably change this again before merging this PR. This method of collecting workflow parameters is only used by a few older starters, but it's not ideal going forward. The workflow name like `Ping` should be associated with the `Ping` starter, but the defaults for other workflow parameters can be populated by a generic function.

starter/objects.py
gnott: I left in the old `print()` call for if you use the starter as a console script you will see something in the console if there is an exception.

starter/starter_Ping.py
gnott: 👍

gnott:
I changed how the workflow params are populated, first to get default values, then the particular ones for a Ping workflow are in its starter. I think it will be more clear if this will work when I refactor a few more starters in the next PR, so if these tests are still green I'll merge this PR and continue to the next batch, thanks @giorgiosironi!

PR #1058, opened 2020-03-18
XML processing instructions in xmlio rewriting.

gnott:
Re issue https://github.com/elifesciences/issues/issues/5509

XML processing instruction data was no retained when XML is parsed, and not included in output when XML is rewritten using the `elifetools` `xmlio.py` module function.

I updated `elifetools` to support input and output of these processing instruction nodes.

Here the values are collected and added back using the `article_processing` provider. I've also now updated three activities that call `xmlio.parse()` and `xmlio.output()` to also support processing instructions.

### DIFF
--- a/activity/activity_PublishFinalPOA.py
+++ b/activity/activity_PublishFinalPOA.py
@@ -226,7 +226,8 @@ class activity_PublishFinalPOA(Activity):
         # Register namespaces
         xmlio.register_xmlns()
 
-        root, doctype_dict = xmlio.parse(xml_file, return_doctype_dict=True)
+        root, doctype_dict, processing_instructions = xmlio.parse(
+            xml_file, return_doctype_dict=True, return_processing_instructions=True)
 
         soup = self.article_soup(xml_file)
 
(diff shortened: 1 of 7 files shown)


activity/activity_PublishFinalPOA.py
```diff
@@ -226,7 +226,8 @@ def convert_xml(self, doi_id, xml_file, filenames, new_filenames):
         # Register namespaces
         xmlio.register_xmlns()
 
-        root, doctype_dict = xmlio.parse(xml_file, return_doctype_dict=True)
+        root, doctype_dict, processing_instructions = xmlio.parse(
```
giorgiosironi: ah, assuming this tuple come from the `xmlio` library they could have made it a named tuple or a stronger type. Now we're stuck with it

gnott: I didn't want to make a breaking change at this time but it should get cleaned up at some point. `elife-tools` library is due for a refresh.

gnott:
I'm curious to see if `end2end` tests are fine with this change so I will merge to see!

PR #1060, opened 2020-03-20
Refactored starters which have simple console arguments

gnott:
Continuing issue https://github.com/elifesciences/elife-bot/issues/992

Refactored `default_workflow_params()` to also set the `domain` and `task_list` value.

These starters are similar because they only gather the `ENV` value from the command line arguments, or are started by an S3 notification message.

For the `PublicationEmail` starter, the `allow_duplicates` value is not used by the activity anymore, so that value in the starter data is removed.

The non-S3 notification starters are fairly simple and similarly changed.

For the S3 notification starters - `IngestDigest` and `IngestDecisionLetter`, there are a few additional features

- retained the custom logging identity value instead of using the default one by overloading `__init__()` of the object
- the `run` and `info` values are passed around for use in the input data
- `NullRequiredDataException` exception catching is kept
- the starter `"__main__"` is not functional since these only start from an S3 notification, so I removed the extra code

I tried my best to not mess up any values, the tests are passing, we can test some of these on the `continuumtest` environment before deploying it to `prod` in case something in the S3 notification ones have a bug.

### DIFF
--- a/starter/starter_IngestDecisionLetter.py
+++ b/starter/starter_IngestDecisionLetter.py
@@ -58,8 +58,3 @@ class starter_IngestDecisionLetter(Starter):
                 'Exception starting workflow execution for workflow_id %s' %
                 workflow_params.get('workflow_id'))
             self.logger.exception(message)
-
-
-if __name__ == "__main__":
-    # note: this starter must be started by an S3Notification and not directly from command line
-    pass
--- a/starter/starter_IngestDecisionLetter.py
+++ b/starter/starter_IngestDecisionLetter.py
@@ -1,64 +1,65 @@
-import boto.swf
 import json
 from provider import utils
 from S3utility.s3_notification_info import S3NotificationInfo
+from starter.objects import Starter, default_workflow_params
 import starter.starter_helper as helper
 from starter.starter_helper import NullRequiredDataException
 
 
-class starter_IngestDecisionLetter():
-    def __init__(self):
+class starter_IngestDecisionLetter(Starter):
+
+    def __init__(self, settings=None, logger=None):
+        super(starter_IngestDecisionLetter, self).__init__(
+            settings, logger)
         self.const_name = "IngestDecisionLetter"
+        # logging
+        if not self.logger:
+            self.logger = helper.get_starter_logger(
+                self.settings.setLevel, helper.get_starter_identity(self.const_name))
+
+    def get_workflow_params(self, run, info):
+        workflow_params = default_workflow_params(self.settings)
+        workflow_params['workflow_id'] = "%s_%s" % (self.const_name,
+                                                    info.file_name.replace('/', '_'))
+        workflow_params['workflow_name'] = self.const_name
+        workflow_params['workflow_version'] = "1"
+        workflow_params['execution_start_to_close_timeout'] = str(60 * 15)
+
+        input_data = S3NotificationInfo.to_dict(info)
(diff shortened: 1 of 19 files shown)


starter/starter_IngestDecisionLetter.py
```diff
@@ -1,64 +1,65 @@
...
-    ENV = utils.console_start_env()
-    SETTINGS = utils.get_settings(ENV)
-
-    STARTER_OBJECT = starter_IngestIngestDecisionLetter()
-
     # note: this starter must be started by an S3Notification and not directly from command line
-    RUN = None
-    INFO = None
-    STARTER_OBJECT.start(settings=SETTINGS, run=RUN, info=INFO)
+    pass
```
giorgiosironi: needed at all? remove the `if`?

starter/starter_IngestDigest.py
```diff
@@ -10,59 +10,60 @@
...
-class starter_IngestDigest():
-    def __init__(self):
+class starter_IngestDigest(Starter):
+
+    def __init__(self, settings=None, logger=None):
+        super(starter_IngestDigest, self).__init__(
+            settings, logger)
         self.const_name = "IngestDigest"
+        # logging
+        if not self.logger:
```
giorgiosironi: if `self.const_name` was assigned before the call to `super`, could this default logger be pushed up into the parent `__init__`? Maybe a follow-up, wouldn't add to the changes here

starter/starter_PublishPOA.py
```diff
@@ -1,53 +1,49 @@
...
-            print(message)
-            logger.info(message)
 
 if __name__ == "__main__":
 
     ENV = utils.console_start_env()
     SETTINGS = utils.get_settings(ENV)
 
-    o = starter_PublishPOA()
+    STARTER = starter_PublishPOA(SETTINGS)
```
giorgiosironi: good indication of global variables in uppercase, by the way

starter/starter_IngestDecisionLetter.py
gnott: I was undecided at the time whether leaving the snippet with a comment about how it is started was better than removing it entirely. Based on your comment, I think removing it entirely on these is probably better - thanks!

starter/starter_IngestDigest.py
gnott: Yes, I believe it could thanks for the suggestion @giorgiosironi! In fact, a little later on I noticed on the issue describing some steps to refactor the starters, I noted that the log files would be instantiated a consistent way using the starter_helper module. I forgot this, and used the older method for setting the log file for starters. There are currently two methods in use, and changing this will help get closer to potentially using a consistent one way to set the logs when all starter refactoring is completed.

gnott: I'm making it so the `name` can be passed in to set `self.name`, and then that can be used. We can eventually set a `self.name` for every starter so the logging identity of each is consistently printed.

starter/starter_PublishPOA.py
gnott: I credit the linter for suggesting the uppercase of certain variables.

gnott:
Some testing and observations of cron triggered starters on the `continuumtest` bot instance are looking good, no errors to report. I will deploy these to the `prod` `elife-bot` and Tuesday we should see everything remain to work.

PR #1063, opened 2020-03-24
Add CLOCKSS as a pub router deposit recipient.

gnott:
Re issue https://github.com/elifesciences/issues/issues/5503

Adding CLOCKSS to all the necessary spots to be able to send some samples manually.

Configuring the sending of files daily will come later.


gnott:
This is a new recipient, follows previous similar configurations, and I'm going to give it a try, merging now.

Issue #1067, opened 2020-03-31
Improve handling of wand.exceptions.CorruptImageError in imageresize.py


gnott:
An example today from ingesting eLife article `50469` included a `.tif` file which resulted in an exception when attempting to resize the image to `.jpg`. An example stack trace message in the `worker.log`:

```
2020-03-30T12:08:35Z ERROR worker_8130 error resizing image elife-50469-fig5-v2
Traceback (most recent call last):
  File "/opt/elife-bot/provider/imageresize.py", line 39, in resize
    image.save(file=image_buffer)
  File "/opt/elife-bot/venv/lib/python3.5/site-packages/wand/image.py", line 5042, in save
    file.write(self.make_blob())
  File "/opt/elife-bot/venv/lib/python3.5/site-packages/wand/image.py", line 4921, in make_blob
    self.raise_exception()
  File "/opt/elife-bot/venv/lib/python3.5/site-packages/wand/resource.py", line 222, in raise_exception
    raise e
wand.exceptions.CorruptImageError: Insufficient memory (case 4) `' @ error/jpeg.c/JPEGErrorHandler/322
2020-03-30T12:08:35Z ERROR worker_8130 An error occurred during Convert Images To JPG
```

The longer lasting effect seemed to be a large amount of virtual memory was tied up with the Python worker process that tried to resize it, reaching 1.7Gb used.

I think we can make a small improvement to handle this particular exception

- [x] In `provider/imageresize.py`, if unable to `image.save()` the resized image due to `wand.exceptions.CorruptImageError` then log the exception and destroy the image to free up memory usage

gnott:
Re issue https://github.com/elifesciences/issues/issues/5516

PR #1068, opened 2020-03-31
Destroy image on exception in imageresize provider, plus tests.

gnott:
Fixes https://github.com/elifesciences/elife-bot/issues/1067.

It looked like a good idea to call `image.destroy()` if any exception is raised when resizing an image. There could be other reasons for why image resizing fails beyond `CorruptImageError` and we do not want the in-memory image object to remain afterward.

I also expanded and refactored the tests to cover the exception code.

There's many lines of the `provider/imageresize.py` provider that are not covered in the test scenarios, but I didn't add any additional tests here. They include the logic for resizing by width, height, and all the file renaming logic. We used to use these when we resized image into multiple sizes, but right now we only reformat `.tif` to `.jpg` with no changes to the dimensions.

### DIFF
--- a/provider/imageresize.py
+++ b/provider/imageresize.py
@@ -9,6 +9,7 @@ def resize(format, filep, info, logger):
     try:
 
         with Image(file=filep, resolution=96) as tiff:
+            image = None
             image_format = format.get('format')
             if image_format is not None:
                 image = tiff.convert(image_format)
@@ -43,6 +44,9 @@ def resize(format, filep, info, logger):
     except Exception as e:
         message = "error resizing image %s" % info.filename
         logger.error(message, exc_info=True)
+        # free up memory
+        if image:
+            image.destroy()
         raise RuntimeError("%s (%s)" % (message, str(e)))
 
     filename = info.filename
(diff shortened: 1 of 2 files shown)


provider/imageresize.py
```diff
@@ -43,6 +44,9 @@ def resize(format, filep, info, logger):
     except Exception as e:
         message = "error resizing image %s" % info.filename
         logger.error(message, exc_info=True)
+        # free up memory
+        if image:
```
giorgiosironi: seems a reasonable check. Can't guarantee `destroy` will never throw an exception, but it's worth a try in that case, which is already failing

gnott: This seemed like a simple way to try and free up memory without nesting exception handling, since the whole procedure is wrapped in a `try`. In the future if the `image.destroy()` becomes a problem here, the exception handling could be split up instead of it covering the huge block of code. I could open a separate issue for that if you think it is worthy. The test coverage of this provider isn't great either, and I can put that in the issue too. Yes I think I'll create an issue for later!

PR #1070, opened 2020-04-03
Declare image earlier in imageresize so always exists.

gnott:
Re issue https://github.com/elifesciences/issues/issues/5543 where an image file conversion raised exception.

I suspected the `image` would potentially not exist in the case of certain exceptions, and it should have been declared earlier. Due to what looks like an EPS file, it results in the exception:

`UnboundLocalError: local variable 'image' referenced before assignment`

Here, declare `image` before anything really happens, it should (conceivably) always exist when  an exception is raised.

### DIFF
--- a/provider/imageresize.py
+++ b/provider/imageresize.py
@@ -6,10 +6,10 @@ from wand.image import Image
 def resize(format, filep, info, logger):
 
     image_buffer = None
+    image = None
     try:
 
         with Image(file=filep, resolution=96) as tiff:
-            image = None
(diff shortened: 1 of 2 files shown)


gnott:
It's possible because of the use of the `with` context the `image` is still not present, but we can try this first and see if we get another exception in the future to confirm we need to do more on this.

gnott:
I added an additional test case illustrating when the image data is unacceptable for `wand.Image()` and it seems to be happy with the exception handling logic.

PR #1076, opened 2020-04-16
Publication email add recipients from article XML

gnott:
Re issue https://github.com/elifesciences/issues/issues/5081

This improvement to the `PublicationEmail` activity is to also send emails to corresponding authors in the article XML that are not already included in the list of recipients we get from an external source.

The `authors_from_xml()` function gets the email, first name, and last name from the article XML from two possible places (which are populated by the `elife-tools` parser, by way of the `article` provider). These two places are either an inline `<email>` tag for the author, or from their affiliation.

`merge_recipients()` takes any email address from the article XML that is not in the original list of recipients and adds it to the full list of recipients.

Along with test scenarios for the new code, when working on this code, I noticed the existing tests could be improved. Specifically, I populated the email template `.json` header files so addiitonal exceptions were not raised, and then I added comparisons with the `admin_email_content` string, where the tests should fail if an article XML was not parsed correctly or a template could not be populated.


activity/activity_PublicationEmail.py
```diff
@@ -284,12 +284,16 @@ def send_emails_for_articles(self, articles):
...
-                    authors = self.get_authors(article.related_insight_article.doi_id)
+                    article_authors = self.get_authors(article.related_insight_article.doi_id)
+                    xml_authors = authors_from_xml(article.related_insight_article)
+                    authors = self.merge_recipients(article_authors, xml_authors)
                 else:
                     authors = None
                     self.log_cannot_find_authors(article.doi)
             else:
-                authors = self.get_authors(article.doi_id)
+                article_authors = self.get_authors(article.doi_id)
```
giorgiosironi, line 294: some duplication across branches, can a private method be extracted reconciling their differences?

activity/activity_PublicationEmail.py
```diff
@@ -723,3 +746,28 @@ def choose_email_type(article_type, is_poa, was_ever_poa, feature_article):
...
+            for email in author.get('email'):
+                xml_author['e_mail'] = email
+                xml_author['first_nm'] = str(author.get('given-names'))
+                xml_author['last_nm'] = str(author.get('surname'))
+                authors.append(xml_author)
+        elif author.get('affiliations'):
+            # add author for each affiliation email
+            for aff in author.get('affiliations'):
+                if aff.get('email'):
+                    xml_author = OrderedDict()
```
giorgiosironi, line 768: similar duplication, extract a function to build `xml_author`?

activity/activity_PublicationEmail.py
gnott: I agree, I also noticed the duplication and it wasn't ideal. I have a full refactor of this now, two separate object methods and separate check for if there are no authors extracted.

activity/activity_PublicationEmail.py
gnott: Yes, I agree here too, this looked like a little bit of duplication. I've separated out the `OrderedDict()` structure. The logic of getting data from the author record or its affiliation(s) remains, since it would not look much better if further refactored, I think.

gnott:
Feedback well spotted, thanks for taking a look @giorgiosironi, it is improved with taking into account your second opinion. Should be ready to merge and deploy now if tests are still green.

PR #1077, opened 2020-04-17
Validate digest image file name by default.

gnott:
Fixes https://github.com/elifesciences/issues/issues/4678

As part of validating digest input, also look at the image file name and its extension is supported. It is integrated along with the other validation rules we can use to check input.

As a result of this, the test scenarios that used only a `.docx` as input and expected a valid result were changed to use a `.zip` input instead, changing the test fixture `DIGEST 99997.docx` to `DIGEST 99997.zip` which includes an image file for this fixture.

I also fixed a bug with a test assertion message that used `string.format()`, the multiline string needed to be wrapped in parentheses to make use of `.format()` properly.


gnott:
I also removed test scenarios that had a `.docx` file input, because we currently always expect and image to accompany every digest, and the contents are to be inside a `.zip` file, so a `.docx` on its own is not supported at this time.

provider/digest_provider.py
```diff
@@ -142,13 +152,25 @@ def image_file_name_from_s3(settings, article_id, bucket_name):
...
         return False
     if not digest_content.image.file:
         return False
     return True
 
 
+def validate_image(digest_content):
+    """validate the image file by its name"""
+    if not has_image(digest_content):
+        return False
```
giorgiosironi, line 167: boolean results can make it difficult to differentiate the reason for validation failure, consider using exceptions in this layer (where a type or a message can be specified)

gnott: I think I understand what you're describing and I've changed it. A simple check for whether the input exists before trying to process it works for me, but if it is better to rely on exceptions, that is fine too. I hope I did it as hoped.

gnott: It still returns a boolean but it also has exception handling, seems to be what I ended up with, otherwise it gets more complicated with error message concatenation. Good enough for now I hope.

PR #1081, opened 2020-04-30
Escape chars in email headers prior to json loads.

gnott:
Fixes https://github.com/elifesciences/issues/issues/4185

There were some recent changes to the `PublicationEmail` activity, and I also found the older issue above, where backslash characters caused errors when building email headers from the templates.

In this PR, I've added a specific fix to escape `\` and `"` characters in an article's `article_title` as an improvement to this. These are two known, possibly common, characters that may appear in an article title. They also proved to cause a `json.decoder.JSONDecodeError` exception when the JSON string from rendering the email header template file has a character that is not valid JSON. The character replacements fix the potential string to JSON loading problem.

Test cases are added for some specific examples.

I didn't do any linting or try to increase test coverage, which looks like a potential future task, since these files are still a little messy. I wanted to keep this PR simple and not more confusing with a bunch of linting changes.

### DIFF
--- a/provider/templates.py
+++ b/provider/templates.py
@@ -297,6 +297,8 @@ class Templates(object):
 
             jinja_env = self.get_jinja_env()
             tmpl = self.get_jinja_template(jinja_env, template_name)
+            # fix special characters in subject
+            article = article_title_char_escape(article)
             headers_str = tmpl.render(author=author, article=article)
             headers = json.loads(headers_str)
             # Add the email format as specified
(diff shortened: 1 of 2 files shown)


gnott:
The email templates are from long ago, and at the time I guess I figured a `.json` file to specify the email header data would be suitable. At least for the immediate and most normal situations it has been, which is probably why we've only observed one error, that I can remember.

Maybe it would be better to change the header templates to a YAML file, or something else, that could still accept robust string replacements, while it can still be parsed into a data structure. Or, I've had decent results using `.cfg` files for configurations, where the string value uses Python string formatting rules.

gnott:
Thanks for the PR approval, I guess I may as well merge this one to close the old issue. I'm not sure when I'll get around to email template refactoring and this has a remote chance of protecting us from the bug until then.

PR #1086, opened 2020-05-09
Ingest digest to endpoint refactoring for exception handling and send error email

gnott:
A solution for issue https://github.com/elifesciences/issues/issues/4480

The plan to improve the `IngestDigestToEndpoint` activity was to reduce the reliance on exception handling if something goes wrong when issuing a POST of a digest to the endpoint. This activity is part of an article publication workflow, so when it was first introduced, we didn't want any errors to delay the publication of research articles.

Most recently, the plan included to send an email if an exception is raised, but to still return True so the workflow can continue.

I started with adding the ability to send emails to the activity.

A large block of code was wrapped in a single exception, and I change it to wrap more specific steps, so more specific error messages can be included in the error email.

Data is collected from four different sources, and the mocking required in the tests was extensive. The tests were also confusing and hard to edit. I decided to refactor the activity code, splitting it up in to separate functions, adding a `digest_content` dict to hold data, and some other refactorings.

If any exception is raised, and error email is sent, including some details about the reason why.

The existing tests were split up a bit into success scenarios, failure scenarios, and exception catching, in an attempt to reduce the number of mocks required to test `do_activity()`. There are still many mocks, but slightly fewer, as a result.

I added new tests for coverage of the exception handling, and did some linting.

As a result of all the changes, I expect the diff will be lengthly and possibly hard to follow. If reviewing, it might make sense to just read the two files (activity module, and test module) separately from start to finish, without looking at what is deleted or added.

@giorgiosironi - I hope this is something like what you expected for improving the exception handling described in the issue.


gnott:
Tests for this PR are all green and passing. I think I may go ahead and merge it and deploy. I will be around tomorrow to solve any errors observed.

PR #1093, opened 2020-05-26
Library upgrades and pin GitPython version.

gnott:
Re issue https://github.com/elifesciences/issues/issues/5670

Began as a failed test on travis-ci for one library, due to the version of GitPython specified. The version used in `elife-bot` was no pinned, but this resulted in a newer version in the `prod` instance, but old developer environments had an older version. This wasn't a problem because it seems both version 2 and 3 of GitPython was compatible for how it was used, but I think pinning it in the project is probably a good idea for now.

There's the opportunity in this PR or in a subsequent PR that we upgrade many other Python library versions, which I don't think have been looked at for a while.


lsh-0:
> There's the opportunity in this PR or in a subsequent PR that we upgrade many other Python library versions

ignoring how it may be implemented, how do you feel about having two files for elife-bot with one file describing the requirements needed in general and the other a complete listing of all dependencies (including dependencies of dependencies) with exact pinned versions?

gnott:
It sounds good to me to specify general and specific requirements, it is all open for comment, and a good idea, thanks @lsh-0. Some of the other libraries are like this, the ones with a `setup.py` for setuptools, I specified just names of packages, and then in `requirements.txt` there are some specific version numbers used for testing and development. I'll open a new Github issue for a later time we can review and modify the requirements.

gnott:
I will merge, since tests are green and PR approved - thanks!

PR #1098, opened 2020-06-01
Remove unused PubMed FTP credentials.

gnott:
Re issue https://github.com/elifesciences/issues/issues/5648

The final step now we send PubMed deposits by SFTP, and FTP will no longer be supported beyond August, 2020, we can remove these old, now unused, credentials.

### DIFF
--- a/settings-example.py
+++ b/settings-example.py
@@ -167,12 +167,6 @@ class exp():
     letterparser_config_file = 'letterparser.cfg'
     letterparser_config_section = 'elife'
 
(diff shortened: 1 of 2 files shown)


lsh-0:
we'll need to update the settings in the `elife-bot-formula` and `builder-private` as well I think

gnott:
Yes, good point, I have a PR for `builder-private` but I'll check `elife-bot-formula` over now.

PR #1128, opened 2020-08-18
Change cron PoA workflows and quarter hour segments.

gnott:
Re issue https://github.com/elifesciences/issues/issues/5989

Run `PackagePOA` and `PublishPOA` workflows at multiple times per day at specific times of the day.

A small change to the `cron.py` module breaks up an hour into four quarters cleanly know, instead of there being a mixture of first half, then third quarter and fourth quarter, the scheduling timeslots are in 15 minute increments.

### DIFF
--- a/cron.py
+++ b/cron.py
@@ -92,10 +92,10 @@ def conditional_starts(current_datetime):
         ]))
 
         # POA Publish at specific hours of the day UK time
-        if (local_current_time.tm_hour == 12
+        if (local_current_time.tm_hour == 10
+                or local_current_time.tm_hour == 12
                 or local_current_time.tm_hour == 14
-                or local_current_time.tm_hour == 16
-                or local_current_time.tm_hour == 18):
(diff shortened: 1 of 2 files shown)


cron.py
```diff
@@ -122,9 +122,9 @@ def conditional_starts(current_datetime):
         # Jobs to start at quarter past the hour
         LOGGER.info("Quarter past the hour")
 
-        # POA Packaging at UK local time, hourly between 6:15 and 17:15
+        # POA Packaging at UK local time, hourly between 6:15 and 13:15
         if (local_current_time.tm_hour >= 6
-                and local_current_time.tm_hour <= 17):
+                and local_current_time.tm_hour <= 13):
```
FAtherden-eLife, line 127: Should this be 15 rather than 13?

gnott: Yes! Good catch, thanks Fred.

PR #1131, opened 2020-10-06
PackagePOA, start at twenty minutes past the hour.

gnott:
Re issue https://github.com/elifesciences/issues/issues/6138

For better stability if the files exported from EJP arrive a little late, move the `PackagePOA` workflows to start at `:20` past the hour, instead of `:15` past.

### DIFF
--- a/cron.py
+++ b/cron.py
@@ -118,11 +118,15 @@ def conditional_starts(current_datetime):
                 ("start_seconds", 60 * 31)
             ]))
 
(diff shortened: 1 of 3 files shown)


gnott:
I am adding in `six==1.13.0` to `requirements.txt` to see if it will allow travis-ci tests to run, while still allow `docker` to be compatible, which apparently requires `six==1.10.0`.

gnott:
Tests are green now, I will merge and if there are any problems with requirements then hopefully end2end tests will catch them before deploying.

PR #1139, opened 2020-12-01
fix, drops pip version requirement in install.sh

lsh-0:
this is to fix an infinite resolution lookup

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -20,7 +20,7 @@ find S3utility/ -maxdepth 1 -name '*.pyc' -delete
 source venv/bin/activate
 
 # fixes issues installing wheel packages, hides deprecation warnings
-pip install "pip~=19.2" --upgrade
+pip install pip --upgrade
 
 # fixes python-docx issue #594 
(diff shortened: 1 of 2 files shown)


gnott (approved):
Thanks @lsh-0 for doing this. I assume the test pipeline failures are due to old configurations that will not pass until we merge a fix like this. I ran into a situation where a prod pipeline did not complete when running `install.sh` so things are already not working, and merging this may fix it.

Longer term of course I think there's a separate issue for reviewing and cleaning up all the requirements specified in this project.

Considering the broken-ness right now, I will merge this and if anything is worse we can revert it.

gnott:
Hmm those failed tests may be important, I think possibly the "source" requirements are not an issue and the ` --no-deps` flag might need to be added to where it installs the `requirements.txt` too. I will try that out with a commit on this branch to see.

gnott:
No, that's not going to work, I'll try to resolve those requirements again, after reversing the last commit.

PR #1146, opened 2020-12-15
Send email after first FTP exception in PMCDeposit

gnott:
Re issue https://github.com/elifesciences/issues/issues/6205

In order to make `PMCDeposit` sending FTP timeouts more visible and easier to monitor, the code here will send a simple email including the exception message for the first time an exception is encountered. It sets a value for the `ftp_exception` session attribute when the email is sent, so the email is not sent an additional time for the same workflow execution.

I considered to increment the `ftp_exception` session value each time an exception is reached (although an email would only send once) but counting how many exceptions is not part of the requirements of this feature, so it will just set the value to `1` each time an exception is raised in the attempts to send by FTP as a flag only.

### DIFF
--- a/activity/activity_PMCDeposit.py
+++ b/activity/activity_PMCDeposit.py
@@ -1,4 +1,5 @@
 import os
+import time
 import json
 import zipfile
 import re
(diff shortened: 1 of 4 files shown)


gnott:
I will merge and deploy now to see if there are any errors in prod environment today when running this new code.

PR #1168, opened 2021-02-18
Add BCC address to list of email recipients.

gnott:
Re issue https://github.com/elifesciences/issues/issues/6268

A small change, tested it manually on the real SMTP server. It does consider if the `To` value is a `str` or `list`, but it currently would only understand one `BCC` recipient (not a list of `BCC` recipients).

Hopefully looks ok @lsh-0? I feel like another set of eyes in case it can be improved.

### DIFF
--- a/provider/email_provider.py
+++ b/provider/email_provider.py
@@ -1,4 +1,5 @@
 "provider for sending email"
+import copy
 import os
 import smtplib
 import unicodedata
@@ -117,10 +118,11 @@ def ses_send(connection, sender, recipient, message):
 
 def smtp_send(connection, sender, recipient, message, logger=None):
     "send a MIMEMultipart email to the recipient from sender by SMTP"
-    # strip the BCC header from the message, if present, prior to sending
-    message.__delitem__('BCC')
+    # strip the BCC header from a copy of the message, if present, prior to sending
+    message_to_send = copy.copy(message)
+    message_to_send.__delitem__('BCC')
     try:
-        connection.sendmail(sender, recipient, message.as_string())
+        connection.sendmail(sender, recipient, message_to_send.as_string())
     except smtplib.SMTPSenderRefused:
         if logger:
             logger.error('error in smtp_send: %s ', traceback.format_exc())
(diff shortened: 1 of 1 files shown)


lsh-0 (approved):
looks fine, perhaps add a link to the ticket under discussion for more context as to why BCC recipients are becoming regular recipients.

gnott:
I've attempted to explain, or maybe I made it worse!

The other potential change here is to treat the `BCC` value completely differently, so it has no chance of being included in the `message.as_string()` output as a `BCC:` header. Optionally, we could explicity strip the `BCC` value from `message` just prior to `message.as_string()`, and it might be more clear that the value isn't included. We are relying on the SMTP server to ignore or strip out the `BCC:` header, since it is not a valid header to begin with.

lsh-0:
> We are relying on the SMTP server to ignore or strip out the BCC: header, since it is not a valid header to begin with.

Could we remove that ambiguity by doing a `message.BCC = None` ?

lsh-0:
or maybe `email_message.replace_header('BCC', None)` 

https://docs.python.org/3/library/email.message.html#email.message.EmailMessage.replace_header

gnott:
Yeah, it will be helpful to show it's not included.

Searching for a oneliner, I think the winner will be

`message.pop("BCC", None)`

gnott:
`email_message.replace_header()` isn't bad either, but it might raise a `KeyError` if the header doesn't exist?

lsh-0:
I don't mind, just offering up suggestions.

gnott:
Much appreciated your input!

gnott:
Hmm looks like `message.pop()` is not working, I'll continue with your suggestion.

gnott:
I want to use `__delitem__()` looks the cleanest, and then I had concerns it might be destructive to the `message` so we'll make a copy of it first before modifying it.

lsh-0:
`__delitem__` is a magic method, it's for dealing with `del object['property']` behaviour

gnott:
Yes, part of my original thought was to use `del` but avoid having to check for if the key exists or having to handle exceptions. I may change my mind yet again, but I think I'll leave it to tomorrow, obviously I should review this PR again before merging.

gnott:
Thanks for comments @lsh-0, I'm going to close this PR in favour of a different solution in PR https://github.com/elifesciences/elife-bot/pull/1169

PR #1169, opened 2021-02-18
Do not set bcc on email message, pass when sent.

gnott:
Re issue https://github.com/elifesciences/issues/issues/6268

A re-do of PR https://github.com/elifesciences/elife-bot/pull/1168 which I will close.

Testing of email BCC recipients and an implementation in the previous PR didn't sit right with me; since `BCC` is not a valid email message header, and now there is more clarification of how to implement BCC, the email message itself shouldn't have a `BCC` header set. Then we don't have to remove it from the headers prior to sending the email message.

To make a BCC recipient work, we pass it in at the time of sending the message, where it can then be added to the list of email recipients. It will not be present in the message headers.

The BCC feature was recently added and was only used by the `PublicationEmail` activity, so refactoring the function arguments should have little backwards compatibility problems.

### DIFF
--- a/activity/activity_PublicationEmail.py
+++ b/activity/activity_PublicationEmail.py
@@ -420,10 +420,10 @@ class activity_PublicationEmail(Activity):
 
         message = email_provider.simple_message(
             headers["sender_email"], str(author.get('e_mail')), headers["subject"], body,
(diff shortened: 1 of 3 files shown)


lsh-0 (approved):

gnott:
Thanks @lsh-0 - I'll merge and try it out now.

PR #1230, opened 2021-06-17
Accept header to request to Lax versions endpoint.

gnott:
Re issue https://github.com/elifesciences/issues/issues/6732

A new settings value `lax_article_versions_accept_header` will specify the value for the `Accept:` header in a request to the API endpoint specified in the setting value `lax_article_versions`.

Previously, the API endpoint will return the most recent content version as a response, and specifying a specific content version will make transitioning to a newer version of the article versions data easier.

We can specify `Accept:` header values for other endpoints later as we like.

[new] code is linted with `black`.

### DIFF
--- a/provider/lax_provider.py
+++ b/provider/lax_provider.py
@@ -16,9 +16,13 @@ class ErrorCallingLaxException(Exception):
     pass
 
 
-def lax_request(url, article_id, verify_ssl, request_type='version', auth_key=None):
+def lax_request(
+    url, article_id, verify_ssl, request_type="version", auth_key=None, headers=None
+):
     "common request logic to Lax"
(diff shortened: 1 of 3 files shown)


lsh-0 (approved):

gnott:
Ready for `end2end` test pipeline

gnott:
Repeatable failure in `end2end` testing, which I've tracked down to how I didn't use `dict.update()` properly, since it returns `None`, the auth headers will be missing. I'll fix this in the next PR.

lsh-0:
yeah, this one bites me fairly often as well. I switch back and forth between languages where the 'update' type function returns a new collection and Python where it modifies it in place.

PR #1245, opened 2021-07-08
Different headers for zip only to SWH endpoint.

gnott:
Discovered by testing against the Software Heritage staging site the new procedure to support the sending of large files, the first upload request, with a zip and atom XML file, can have a `Content-Type: multipart/form-data` header, but sending a zip file only needs to have a `Content-Type: application/zip` header.

Meaning, when using Python `requests` library, the first upload can use the `files` argument, but the next uploads need to use the `data` argument.

This is still to be tested and confirmed with the Software Heritage staging site endpoint again.

### DIFF
--- a/provider/software_heritage.py
+++ b/provider/software_heritage.py
@@ -201,18 +201,15 @@ def swh_post_request(
 
     headers = {"In-Progress": "%s" % str(in_progress).lower()}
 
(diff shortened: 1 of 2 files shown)


gnott:
Re issue https://github.com/elifesciences/issues/issues/6773

PR #1279, opened 2021-08-30
glencoe, support for new kitchen sink msid

lsh-0:


provider/glencoe_check.py
```diff
@@ -80,7 +75,24 @@ def pad_msid(msid):
...
+
+    The kitchen sink however is its own article with its own set of videos.
+    Its msid is still changed from 1234567890 to a generated one during testing *except* for video hrefs.
+    See `elife-spectrum/update-kitchen-sinks-from-github.sh`."""
+    msid = int(msid)
+    kitchen_sink_id = 1234567890
+    if msid == kitchen_sink_id:
+        return str(msid)
+    if msid > 100000:
         return pad_msid(msid[-5:])
```
gnott, line 96: I think here if `msid` is now an int, the list sequence operation needs a str. I'll change this to test it

provider/glencoe_check.py
```diff
@@ -61,13 +61,8 @@ def metadata(msid, settings):
...
     return resp.json()
 
-
 def jpg_href_values(metadata):
-
-    return list((seq(list(metadata.items()))
-                .filter(lambda key_value: 'jpg_href' in key_value[1])
-                .map(lambda k_v: k_v[1]['jpg_href'])))
-
+    return [val['jpg_href'] for val in metadata.values() if 'jpg_href' in val]
```
gnott, line 65: I like this restatement 👍

lsh-0:
> there was an example of a generated kitchen sink msid value of 91234567890, and when checking Glencoe metadata the first 9 will need to be lopped off.

that's right, I patched the random number generator during testing to always return a 9. A generated msid for the kitchen sink will  then always have a prefix. Outside of testing the possible range is much higher.

PR #1314, opened 2021-09-22
Fix article_id zerofill for video JPG file names.

gnott:
Bug fix to PR https://github.com/elifesciences/elife-bot/pull/1313 observed after testing on the continuumtest environment with an article number less than 10000.


gnott:
Re issue https://github.com/elifesciences/issues/issues/4228

PR #1325, opened 2021-10-07
Requirements update to use the latest libraries

gnott:
Re issue https://github.com/elifesciences/issues/issues/5679

Going through the list of libraries installed from the `requirements.txt` file, they are each updated to use the most recent or very recent version of each in almost all cases.

This resulted in some minor code changes, including

- change `newrelic.agent.record_exception()` to `newrelic.agent. notice_error()`,
- in test scenarios, changed `from google.cloud._helpers import _UTC` to `from google.cloud._helpers import UTC`

When using `pytest` to run test scenarios, there were a few warnings listed, most of which are fixed too, including:

- some test data strings specified as raw strings, `r''`,
- specified a `Loader` in `yaml.load()` invocation,
- adjusted the usage of `testfixtures` in `tests/activity/test_activity_send_dashboard_properties.py`,
- changed `self.assertRegexpMatches()` in a test case to `self.assertRegex()`

I'm curious to see if `end2end` tests will produce an error which the unittests cannot, so if this PR is green, I plan to merge it to see, and it can be reverted if things are not looking stable enough to deploy.


gnott:
One more note I forgot to include, these libraries are removed from `requirements.txt` since it seemed like they're not used by anything, and `pipdeptree` confirmed that. I think they may have been part of the older Google cloud libraries (`google-cloud-bigquery==1.21.0`), which are not used in the newer version (`google-cloud-bigquery==2.28.0`):

```
pyasn1==0.4.8
pyOpenSSL==18.0.0
newrelic==2.72.0.52
```

PR #1382, opened 2021-12-09
Cross-platform sed commands.

gnott:
Re issue https://github.com/elifesciences/issues/issues/6980

I tried running the script supplying a specific `elifetools` version, and on Mac which has BSD `sed` the result of the script was not expected. Here's some suggestions for changes which worked for me.

BSD `sed` does not accept any `--` arguments.

It looks like `-E` works on both GNU and BSD `sed` pretty well.

The in-place editing `-i` seems to require a file extension value, otherwise it takes the first character it can find as the file extension. Documentation on this has more details about special considerations for when writing a file in-place which doesn't have a file extension.

### DIFF
--- a/update-dependencies.sh
+++ b/update-dependencies.sh
@@ -24,12 +24,12 @@ if [ -n "$package" ]; then
     pip install -r requirements.txt
 
     # make Pipenv install exactly what we want (==).
-    sed --in-place --regexp-extended "s/$package = \".+\"/$package = \"==$version\"/" Pipfile
+    sed -i .bck -E "s/$package = \".+\"/$package = \"==$version\"/" Pipfile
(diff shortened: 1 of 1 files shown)


lsh-0:
this fails on linux. Let me see what I can do

lsh-0:
could you give my changes a shot and see if they work for you?

gnott:
Super-duper, thanks for that Luke. It makes it more complicated, yes, but I can run it too. The `-i ''` for the file extension has the desired effect, no extra `Pipfile.bak` file is generated.

If you're happy too, please merge if you like, or I can merge if you have no more contributions or comments on this PR.

lsh-0:
> It makes it more complicated, yes, but I can run it too. 

I think the script will end up becoming centralised once I need to port other pipelines and their dependency handling over to it.

PR #1389, opened 2022-01-04
Remove boto.ses functions, not currently used.

gnott:
Re issue https://github.com/elifesciences/issues/issues/916

This project used to use `SES` for sending email and now it uses SMTP, we can remove the old SES code which is no longer used.


gnott:
Re issue https://github.com/elifesciences/issues/issues/916

PR #1462, opened 2022-04-08
update-dependencies, replaces script with a centrally maintained one.

lsh-0:
fyi @gnott 

I need to use the script I created for elife-bot to update single dependencies in other projects now, so I've given it it's own repository (https://github.com/elifesciences/update-python-dependencies) and I'm replacing the instance in elife-bot with a thin wrapper around it.

It should continue behaving the same.

There is also a new option that is supported. The first two args are the same, packagename and packageversion, but the third argument is the locking constraint that can be either 'semver' (`~=`), the default, or 'exact' (`==`).

For example, in bot-lax-adaptor I want to update `elifetools` to version `1.2.3` *exactly* and not have it updated when I update the other dependencies around it.

Let me know if you have any questions or concerns.


gnott:
I like the concept of the script to be centrally located. Initially I had a concern about running scripts downloaded from a GET, though I can see there are few good options to confirm the integrity of the script without making it more complicated. Personally, I would probably be downloading the script first and reading it over before running it, however I imagine you'll be invoking it in automated pipelines. I'm happy to let you decide what is good here, thanks for the update!

lsh-0:
Those concerns are typically well founded when some rando website is telling you to install their program with curl+bash magic as an easy one-liner.

In this case, it's an elife script downloading another elife script. The authors are known and the URL/target is sound but ultimately the decision is yours (as project maintainer). If you want to keep a copy of the script locally in the repo, that's fine.

gnott (approved):

gnott:
Thanks for the additional comments, I approve.

lsh-0:
ok, merging this in. Let me know if you have any problems.

PR #1482, opened 2022-05-25
Dependency `` pinned to ``

elife-alfred-user:
I have run https://alfred.elifesciences.org/job/dependencies/job/dependencies-elife-bot-update-dependency/28/display/redirect which resulted in this pull request.


gnott:
Using the dependency pipeline to update the version of `pyjwt`, the script must be run using no parameters to modify the `Pipfile.lock` file. A side-effect of running it with no parameters is that all package versions are updated. I reviewed them and I think they should not cause any problems, we will see in the `end2end` tests.

PR #1513, opened 2022-07-07
dependencies, bump

lsh-0:


gnott (approved):
Looks like the best way, just update them all, and it looks good, thanks @lsh-0

PR #1514, opened 2022-07-08
python 3.8 upgrade

lsh-0:


lsh-0:
There's no urgency to this if the current version of elife-bot is working without issue. I'll let you merge this one in when it's convenient for you, @gnott .

gnott:
There's no reason to wait I can think of, so I'll merge it now to test end2end.

PR #1553, opened 2022-09-20
Fix logging file_name_map in article_processing.py

gnott:
Fix code merged in PR https://github.com/elifesciences/elife-bot/pull/1552

The `OrderedDict` cannot be `sorted` and retain its structure, instead sort the list of files the `file_name_map` is generated from, which will allow for consistent comparison of log file output in the test scenario.


gnott:
Re issue https://github.com/elifesciences/issues/issues/7677

PR #1576, opened 2022-11-03
Dependency `` pinned to ``

elife-alfred-user:
I have run https://alfred.elifesciences.org/job/dependencies/job/dependencies-elife-bot-update-dependency/72/display/redirect which resulted in this pull request.


gnott:
Used the build pipeline to update `cryptography` and resolve the security alert, it has also updated other libraries to newer versions. It also removed the `docmaptools` library which is fine for now since this project and no other libraries depend on it yet.

PR #1595, opened 2022-12-09
Dependency `` pinned to ``

elife-alfred-user:
I have run https://alfred.elifesciences.org/job/dependencies/job/dependencies-elife-bot-update-dependency/84/display/redirect which resulted in this pull request.


gnott:
In order to also update the `Pipfile.lock`, as opposed to in PR https://github.com/elifesciences/elife-bot/pull/1594, we can update all the library versions used including `certifi`, which is subject to a security vulnerability.

PR #1628, opened 2023-02-20
utils.py, msid_from_doi now handles component/versioned dois.

lsh-0:


tests/provider/test_article.py
```diff
@@ -271,9 +271,9 @@ def test_was_ever_published(self, fake_storage_context):
         fake_storage_context.return_value = FakeStorageContext(
             directory.path, resources
         )
-        self.assertTrue(self.articleprovider.was_ever_published("2444", workflow))
-        self.assertTrue(self.articleprovider.was_ever_published("2419", workflow))
-        self.assertFalse(self.articleprovider.was_ever_published("666", workflow))
+        self.assertTrue(self.articleprovider.was_ever_published("10.7554/eLife.2444", workflow))
+        self.assertTrue(self.articleprovider.was_ever_published("10.7554/eLife.2419", workflow))
+        self.assertFalse(self.articleprovider.was_ever_published("10.7554/eLife.666", workflow))
```
lsh-0, line 276: @gnott , interesting changes here. A string msid was being passed in instead of a doi and 

    int(doi.split(".")[-1])

in `msid_from_doi` was able to handle it.

I'm not sure if that was intentional or a happy accident during development, or if `msid_from_doi` actually needs to handle this case.

gnott (approved):
Thanks for doing the work here @lsh-0! Since it's green I will go ahead and merge it.

PR #1652, opened 2023-03-27
Dependency `` pinned to ``

elife-alfred-user:
I have run https://alfred.elifesciences.org/job/dependencies/job/dependencies-elife-bot-update-dependency/109/display/redirect which resulted in this pull request.


gnott:
Running the pipeline to end up with a modified `PIpfile.lock` file, however this also has used `redis==4.5.3` which is the version we want in regards to security alerts. Green tests so far, let's watch it pass the end2end tests next.

PR #1770, opened 2023-10-10
Dependency `` pinned to ``

elife-alfred-user:
I have run https://alfred.elifesciences.org/job/dependencies/job/dependencies-elife-bot-update-dependency/182/display/redirect which resulted in this pull request.


gnott:
I thought `GitPython` would also need updating to `GitPython==3.1.37` here, but it was already done. These updates can still go through if the tests are green.

PR #1786, opened 2023-10-31
S3 session singleton

lsh-0:


gnott (approved):
Looks like a good way to store the client for future use, and the end2end tests should confirm it works well.

gnott:
For the reason it is not too invasive makes it good, I like it. In future it might make sense to bake it into the `worker.py` because it makes most of the client connections.

PR #1801, opened 2023-11-10
Dependency `` pinned to ``

elife-alfred-user:
I have run https://alfred.elifesciences.org/job/dependencies/job/dependencies-elife-bot-update-dependency/203/display/redirect which resulted in this pull request.


gnott:
Updating the `Pipfile.lock` by running the pipeline with no parameters. It also removes `pyarrow`, which is fine, the newer libraries may not depend on it?

PR #1811, opened 2023-11-30
Check latest preprint DOI in preprint provider build_article()

gnott:
Testing the `ScheduleCrossrefPreprint` today, an early, incomplete, docmap returned a non-eLife DOI when calling `cleaner. version_doi_from_docmap()`. Check that a DOI is found, and that it starts with the right DOI prefix.


gnott:
Also fixed the typo in `PrepringArticleException` to `PreprintArticleException`.

PR #1861, opened 2024-02-12
More connection consolidation

lsh-0:
- [x] ensure test coverage
- [x] review


gnott:
My PR https://github.com/elifesciences/elife-bot/pull/1862 changed `provider/utils.py`, if you will merge in the changes, sorry!

provider/utils.py
```diff
@@ -286,3 +293,25 @@ def content_type_from_file_name(file_name):
...
+
+def set_envvar(key, val):
+    """set a value `val` for the environment variable `key`.
+    raises an `AssertionError` if the requested environment variable is unknown."""
+    assert key in ENVVAR_KNOWN_KEYS, "programming error. unsupported environment key: %s" % key    
+    os.environ[key] = val
+
+def reuse_boto_conn():
+    "returns `True` if the `BOT_REUSE_BOTO_CONN` environment variable has been set to '1'"
+    return envvar('BOT_REUSE_BOTO_CONN', '0') == '1'
```
gnott, line 317: Just curious if from the environment they are always strings?

tests/settings_mock.py
```diff
@@ -107,3 +110,7 @@
 
 # EPP settings
 epp_data_bucket = "epp_bucket"
+
+@mock_aws
+def aws_conn(service, service_creation_kwargs):
```
gnott, line 115: If I'm understanding this part, is it safe to assume this could also be added to the `settings-example.py` file?  Since `worker.py` is invoking this, I assume it will need to be added to the real `settings.py` file for doing deployments.

I hadn't anticipated the settings module to include too much structure, although it is certainly capable of it; it was just an easy way to store Pythonic data.

If there's a way to move this to, say, `utils.py`, and put a flag value in `settings.py` to control it, that's my feedback. Otherwise, I'm happy with another suggestion as you have expressed it here.

starter/objects.py
```diff
@@ -35,12 +35,19 @@ def instantiate_logger(self):
 
     def connect_to_swf(self):
         "connect to SWF"
-        self.client = boto3.client(
-            "swf",
-            aws_access_key_id=self.settings.aws_access_key_id,
-            aws_secret_access_key=self.settings.aws_secret_access_key,
-            region_name=self.settings.swf_region,
-        )
+        if utils.reuse_boto_conn():
```
gnott, line 38: I like the concept around checking - _to reuse, or not to reuse, that is the question..._

gnott:
It's very ambitious to tackle the swf, sqs, sns, all in one go, but you've done it I think. Shaping up to be a good addition!

tests/settings_mock.py
gnott: Thanks, yes, I forgot about parts not used very much, and I see you added the connection logic five months ago to `get_settings()`, it clicks now 👍

starter/objects.py
gnott: For sure we try it out, let's most certainly keep this PR. The results of testing may suggest whether it is better, the same, or worse.

tests/settings_mock.py
lsh-0: It looks like the settings_mock.py file is imported a lot and used directly without any chance to 'inject' the magic `aws_conn` function.

What I'll do is explain what it is and why it is there in each case so it's hopefully not too magic going forwards

PR #1864, opened 2024-02-27
empty commit to trigger tests

lsh-0:
I'm getting a failing test in my centralised connection work and I don't think it's related, but everything also appears to be fine in CI ...

```
pytest tests/activity/test_activity_deposit_decision_letter_ingest_assets.py::TestDepositDecisionLetterIngestAssets::test_do_activity_1
...
E       AssertionError: 'ActivityPermanentFailure' != 'ActivitySuccess'
E       - ActivityPermanentFailure
E       + ActivitySuccess
E        : failed in decision letter zip file example

```

which comes from here:

```python
    def output_bucket_folder_name(settings, manuscript):
>       return settings.decision_letter_bucket_folder_name_pattern.format(
            manuscript=manuscript
        )
E       TypeError: unsupported format string passed to NoneType.__format__
```


gnott:
Ah yes, those remind me of a dependency, and my apologies if it's not documented well. The decision letter parser library uses either a local `pandoc` executable (which I use), or to have Docker running, which is how the tests and the prod instance do it, invoking pandoc that way. You could try one or both of these and try it again locally.

